PR #29, opened 2016-01-21
gschem: Refresh component library cache when reverting page.

### LINKED ISSUE #28: Missing components aren't 'found' when added to project library, and sheet is 'Revert'ed
Briefly discussed a temp fix for this, whilst underlying work on the library system takes place.
To replicate: 
- load a schematic file with components missing from local/project library
- note bunch of red messages about missing components
- copy symbols from 'other' library into project library
- File->Revert
- red messages do not disappear.

Cause of the problem is the symbol cache, which is persistent across anything/everything (it seems!). It should be easy enough to call a 'flush' cache on Revert, so that everything is loaded up from file (as you probably might expect).

peter-b:
Users sometimes use "File→Revert" in the hope of fixing things up when
symbols weren't found.

1. Open page
2. Symbols are missing
3. Copy symbols into expected place
4. Revert page

To make this work as expected, regenerate the symbol cache on page
revert.

Closes #28.

### DIFF
--- a/gschem/src/i_callbacks.c
+++ b/gschem/src/i_callbacks.c
@@ -1539,6 +1539,9 @@ DEFINE_I_CALLBACK(page_revert)
   /* delete the page, then re-open the file as a new page */
   x_window_close_page (w_current, page_current);
 
+  /* Force symbols to be re-loaded from disk */
+  s_clib_refresh();
+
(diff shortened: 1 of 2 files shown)


peter-b:
Still to do:

* [ ] Clean up misleading comment on `s_clib_refresh()`.

PR #40, opened 2017-02-15
ci: Enable test builds using Travis CI

peter-b:
[Travis CI](https://travis-ci.org/) is a continuous integration service that is free to use for open source software hosted on github.

This pull request configures the lepton repository to automatically run `make distcheck` for all pull requests, in order to verify that changes don't break the distributed build. This will easily catch common problems like adding a new file and forgetting to commit it, or forgetting to update a `Makefile.am` after removing a file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,12 +13,17 @@ env:
 
 # Before setting up the source tree, install necessary development
 # headers
-before_install: >
-  sudo apt-get update && 
-  sudo apt-get install -y \
-    libgtk2.0-dev autopoint libgettextpo-dev libstroke0-dev \
-    guile-2.0-dev flex bison groff \
-    texinfo texlive-base texlive-generic-recommended texlive-latex-base
+before_install:
+  - sudo apt-get update
+  - sudo apt-get install -y libgtk2.0-dev autopoint libgettextpo-dev
+    libstroke0-dev guile-2.0-dev flex bison groff texinfo texlive-base
+    texlive-generic-recommended texlive-latex-base
+
+  # FIXME[2017-02-15] This is required because Ubuntu Trusty only has
+  # Guile-2.0.9, which doesn't include SRFI-64
+  - sudo mkdir -p /usr/share/guile/2.0/srfi/srfi-64
+  - sudo curl http://git.savannah.gnu.org/cgit/guile.git/plain/module/srfi/srfi-64.scm -o /usr/share/guile/2.0/srfi/srfi-64.scm 
+  - sudo curl http://git.savannah.gnu.org/cgit/guile.git/plain/module/srfi/srfi-64/testing.scm -o /usr/share/guile/2.0/srfi/srfi-64/testing.scm
 
 # Set up the source tree by running autotools
 install: ./autogen.sh && ./configure --disable-update-xdg-database CFLAGS='-DSCM_DEBUG_TYPING_STRICTNESS=0'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,42 @@
 # Very basic Travis CI (http://travis-ci.org) control file that allows
 # some basic Linux-based continuous integration testing (for free).
 
+sudo: required
+dist: trusty
+
 language: c
+
+env:
+  global:
+    # COVERITY_SCAN_TOKEN
+    - secure: "BIl7HDdWo2EpK63Nta2Sjy/V2rMYt4E5Jl+tWEkWR3be+s61oRg2DoIqdTTntvo8os8gq90Fb8WJJWZQ2806BfsNF+Z+DZqN59iNQOEB2aiA0BhggyOHnF+OoZZfgv7q4Eq0XOS7KLSrnCu18E0qQKxJSjnBrO9k1cLS3QnKDCA="
 
 # Before setting up the source tree, install necessary development
 # headers
 before_install: >
+  sudo apt-get update && 
   sudo apt-get install -y \
     libgtk2.0-dev autopoint libgettextpo-dev libstroke0-dev \
-    guile-2.0-dev flex bison groff texinfo
+    guile-2.0-dev flex bison groff \
+    texinfo texlive-base texlive-generic-recommended texlive-latex-base
 
-# Set up the source tree by fetching Linux-specific prebuilt objects
+# Set up the source tree by running autotools
 install: ./autogen.sh && ./configure --disable-update-xdg-database
 
-# Bootstrap the LCB compiler, build the default target set and run a
-# the default test suite.
-script: make -s all check
+# Compile gEDA and run its tests
+script: >
+  if test "${COVERITY_SCAN_BRANCH}" != "1"; then
+    make -kj4 distcheck
+  fi
+
+addons:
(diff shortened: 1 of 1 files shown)


vzh:
Peter, could you change the first patch so it contains current name of
repository and "master" as a branch name?

I'm not sure about your email address there, though.

vzh:
Second patch:

Why have you set SCM_DEBUG_TYPING_STRICTNESS=0?

Does it somehow affects Travis CI?

Quote from guile sources:

A value of 0 means that there will be no compile time type
checking, since the SCM datatype will be declared as an integral
type.  This setting should only be used on systems, where casting
from integral types to pointers may lead to loss of bit
information.

Does CI work on such systems?

vzh:
Fourth commit:

IMO, if we support Trusty (which IIUC is the only tested system
having guile 2.0.9), we should also somehow inform users of how to
install srfi-64, e.g. in INSTALL, or provide functionality similar
to used in the patch. I think we should not depend on curl or
wget. Wouldn't it be better to just eventually say that we support
guile 2.0.11 and no less?

peter-b:
@vzh Static analysis with the Coverity Scan system is subject to a quota, so it's recommended to only perform scans occasionally (once or twice a week).  Also, it uses a special instrumenting compiler, so you can't do the Coverity Scan analysis in the same build as a "normal" `make check`.

The recommended approach is to push the current `master` to a specially-named `coverity_scan` branch (or some other name) whenever you want to do a scan, and that's how I've set things up.  If the branch is `coverity_scan` then Travis will run the Coverity Scan tools; otherwise, it will do a "normal" build and check.

The [name of the repository in the Coverity Scan system](https://scan.coverity.com/projects/peter-b-geda-gaf) is fixed by whichever name the repository was first registered with the system.  When we renamed our official repository at work from `runrev/livecode` to [`livecode/livecode`](https://github.com/livecode/livecode) it caused some breakage in the system. 😞  I was hoping to get things up and running first, and then tackle the rename later, if that's okay.

The e-mail address here is the e-mail address that the results of each scan get sent to.  I don't think we have a "general" contact e-mail for Lepton yet, but when we have one then we should definitely change the e-mail address specified here.

The macros that Guile uses with non-0 `SCM_DEBUG_TYPING_STRICTNESS` are pretty evil, and invoke undefined behaviour according to the C standard. As a result, they make Coverity Scan detect a large number of arguably-valid errors, but they're not useful ones.  On the other hand, probably the "normal" integration tests should be done with `SCM_DEBUG_TYPING_STRICTNESS=2` so that there is maximum compile-time detection of incorrect use of Scheme values. N.b. the code that is generated by the compiler _should_ be identical, no matter what the setting of this macro is, but the Guile header macros are evil enough that I'm sceptical.

So, to summarize:

1) At the moment I don't think I can change the settings for Coverity Scan integration, until we have a "general" e-mail address for Lepton.  Up until then, actually Coverity Scan is very helpful for detecting programming errors.  Is it okay if we leave these settings the way they are, for now?

2) I will update this PR so that we set `SCM_DEBUG_TYPING_STRICTNESS=2` for non-Coverity builds

3) I think it is best to have a specific `coverity_scan` branch for static analysis builds, instead of always scanning `master`

Is that okay?

peter-b:
@vzh Unfortunately Ubuntu Trusty is the most recent Linux VM provided by Travis.  I looked into installing a newer version of Guile, but unfortunately it is not available in the repositories.  Probably most users won't want to run the test suite, so Lepton should compile and run fine for them even if they use Ubuntu Trusty.  

An alternative approach to using `srfi-64` could be to change the gnetlist tests to use the unit test framework in [libgeda/scheme/unit-test.scm](https://github.com/lepton-eda/lepton-eda/blob/master/libgeda/scheme/unit-test.scm).  I am happy to do this, if you like.  This would keep compatibility with Guile 2.0.9.  Are there some srfi-64 features that you use in gnetlist that are not available with the libgeda Scheme unit test module?

What do you think?

peter-b:
@vzh Another option: the `configure` script could detect whether `srfi-64` support is present, and disable running the tests that need it if necessary.

vzh:
I agree with value 0 for Travis CI.

Though another point, I'll quote further comments in the guile
sources then:

  A value of 2 provides a maximum level of compile time type
  checking since the SCM datatype will be declared as a struct.
  This setting should be used for _compile time_ type checking
  only, since the compiled result is likely to be quite
  inefficient.  The right way to make use of this option is to do
  a 'make clean; make CFLAGS=-DSCM_DEBUG_TYPING_STRICTNESS=2', fix
  your errors, and then do 'make clean; make'.

Would you like our users to complain about `inefficiency' of the
program? ;-)

peter-b:
@vzh 

> Would you like our users to complain about `inefficiency' of the
program? ;-)

Definitely not!  I was suggesting that we make it set to 2 for non-Coverity builds _on Travis_.  I am certain that we should leave at the default setting (1) for all non-Travis builds!

vzh:
@peter-b srfi-64 is currently a quite useful module and I'd prefer
using it. Besides, there is a good testing suite in guix based on
the module, and I wanted first use it, however its license is
GPL-3, and I decided to write something like it myself. I think it
would be very well to use already written code already debugged by
good technicians. These days many people release their code under
GPL-3 like guile itself as I see it on their list. What do you
think about changing a license?

peter-b:
@vzh I'm happy with GPLv3, and I use it for my personal projects (e.g. [lcb-mode](https://github.com/peter-b/lcb-mode)). @ehennes775, what do you think about using GPLv3?

vzh:
@peter-b Your words:

> I will update this PR so that we set SCM_DEBUG_TYPING_STRICTNESS=2 for __non__-Coverity builds

probably I got something wrong?

I'm happy with any decision you wish WRT this variable if it is
just for CI.

BTW, are you using browser or MUA to reply? I tried two browsers
but found no handle for citation :-(

vzh:
To put things other way: what is this non-Coverity thing?

peter-b:
@vzh I am editing this in a browser! You can quote something by putting a `>` at the start of the line, like this:

```markdown
> To put things other way: what is this non-Coverity thing?
```

gives:

> To put things other way: what is this non-Coverity thing?

Firstly: yes, I am suggesting only setting this variable for continuous integration builds.

I am sorry for not being clear enough.  For continuous integration, there are two build types:

- a normal test (we do `make distcheck` using a normal compiler, etc.)  In this case I suggest we set `SCM_DEBUG_TYPING_STRICTNESS=2`
- a Coverity analysis (only on `coverity_scan` branch, using special tools, etc.)  In this case I suggest we set `SCM_DEBUG_TYPING_STRICTNESS=0`

These settings would *only* be set when doing CI builds, and never otherwise.  Is that okay?

vzh:
Got it. I'm OK with your plans for CI and naming.

Let me know if/when all is ready to be pushed.

BTW, IIUC you offer to push only from PR's. Right?  Who is then
responsible to review all them? Not only me I hope? And who will
review my PR's? And after all, is there a simple way to use just
an editor to submit them?

Too many questions and offtopic, I know ;-)

peter-b:
This should be all ready now!

I'll reply to your other comments by email, I think.

PR #41, opened 2017-02-15
gschem: Use escape continuations for keymap processing

peter-b:
Using delimited `call/ec` escape continuations where appropriate is
significantly more efficient that using traditional Scheme
continuations.


vzh:
@peter-b If I should always comment PRs (should I?), I would suggest to make an efficiency test in this case ;-) Though I'm not insisting :-)

PR #42, opened 2017-02-16
Add code of conduct and guide for contributors

peter-b:


vzh:
@peter-b Could you please don't use terms not accepted by some political or religious movements and forbear from using "gender" or "sexual orientation" in our neutral environment. Could it be rephrased more short and general? I don't want to see lepton political and discouraging right from its beginning. I would prefer something like "non-behavioural personal characteristics they have or you think they have". Thanks.

peter-b:
Unfortunately, in other projects they have found problems where people have been harassing contributors over personal characteristics such as gender, and there have been prolonged arguments because "I don't agree that this is a personal characteristic, therefore it's not against the code of conduct so you can't ban me."  So it is a good idea to have an "including but not limited to" list, just because otherwise you end up arguing about the CoC rather than sorting out disruptive people.

I have moved it into a footnote.  Is that okay?

vzh:
@peter-b Not all people depending on ideology or religion accept such new western terms like gender and gender identity. I don't want them to go away due to ideological reasons. Could you please rephrase it as follows: "__sex__, sexual orientation __or__ gender and gender identity ..."?

peter-b:
@vzh I've changed it. Does that wording work better?

vzh:
@peter-b Yeah, this looks better.

vzh:
@peter-b I've found some typos. It seems to be not very convenient to write about anything so small in comments. Is there a better way to deal with it?

peter-b:
@vzh If you go the the "Files" tab of the pull request, and click on the line where the type occurs, you can make line-by-line comments and group them into a review.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,102 @@
...
+  attention-stealing behaviour is not welcome.
+
+## Moderation
+
+These are the policies for upholding our community's standards of
+conduct.
+
+1. Remarks that violate the Lepton standards of conduct, including
+hateful, hurtful, oppressive or exclusionary remarks, are not allowed.
+(Cursing is allowed, but never targetting another person, and neveri n
```
vzh: neveri n -> never in

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,102 @@
...
+
+3. Moderators will first respond to such remarks with a warning.
+
+4. If the warning is unheeded, the user will be "kicked", i.e. kicked
+out of the communication channel to cool off.
+
+5. If the user comes back and continues to make trouble, they will be
+banned, i.e. excluded indefinitely.
+
+6. Moderators may chooses at their discretion to un-ban the user if it
```
vzh: may choose (without -s) ?

vzh:
@peter-b Although you've omitted this point, I still don't like // comment style in C code, especially mixing styles in nearby code blocks.

peter-b:
@vzh Thank you for the thorough review! I've addressed the points you made.

PR #48, opened 2017-02-18
Automatically add newlines to log messages

peter-b:
When we log messages, we do so via `g_log_default_handler()`, which
always adds a newline to the message.  This makes the newlines at the
end of message a bit redundant.  Conceptually, it makes good sense to
leave the newlines off the ends of log messages, because the newline
character doesn't really contain any information.

This patch makes the log paths inside Lepton (both for logging to disk
and for display to gschem's log window) also automatically add
newlines, so they behave the same as the GLib default log handler.

The immediate effect of this is to add lots of extra whitespace to the
log files.  This patch should be followed up by going through our
loggable strings to remove trailing newline characters, possibly using
a regular expression to do most of the heavy lifting.


peter-b:
@gareth8118 This is for you 👍

vzh:
@peter-b Could you please rebase this onto current master?

peter-b:
@vzh I've rebased this!

PR #47, opened 2017-02-18
Rename libgeda to liblepton.

vzh:
Renamed the directory libgeda/ and some files named libgeda.*
The PR commit addresses issue #39.

PR #45, opened 2017-02-17
Get rid of links to the gEDA launchpad bug tracker.

vzh:
All links have been replaced with link to lepton issue tracker.

PR #50, opened 2017-02-18
Change bugtracker link to point to lepton-eda GitHub issue tracker

gareth8118:
Change the bugtracker parameter in the call in AC_INIT() so that the
help messages printed by the tool suite have the leptop-eda
issue tracker URL.

This is a supplementary commit to 92b534e2.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -8,7 +8,7 @@ dnl directory.
 # Set up configuration system
 #####################################################################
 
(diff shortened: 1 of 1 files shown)


peter-b:
Isn't this covered by #47 already?

https://github.com/lepton-eda/lepton-eda/pull/47/files#diff-67e997bcfdac55191033d57a16d1408aR11?

gareth8118:
It is, but it still looked like an omission to #45 since that dealt directly with links to the bug tracker.

PR #58, opened 2017-02-19
gnetlist: Fix a couple of missing format idx specifiers

peter-b:
In commit 8bd7a38f764a, most translatable strings in Lepton were
changed to use explicit parameter indices for their substitutions.
This fixes a couple of places where the strings were left partially
converted, leading to `-Wformat` compile warnings.


gareth8118:
thanks for catching!

PR #64, opened 2017-02-21
gnetlist: Optimise net rename tracking

peter-b:
This PR converts rename tracking from using an alist to using a hashtable.  This improves the time spent on rename creation from O(N^2) in the number of renames to O(1).

The PR is in two parts:

1. Changing the rename alists provided by the `(get-rename-list)` function to be sorted in a canonical ordering, and updating the testsuite
2. Changing the rename implementation to using a hashtable

This proves that the change in rename implementation has no effect on the testsuite results other than the order renames appear in the result of `(get-rename-list)`.

### DIFF
--- a/gnetlist/scheme/gnetlist/rename.scm
+++ b/gnetlist/scheme/gnetlist/rename.scm
@@ -8,7 +8,7 @@
   #:export (search-rename
             add-rename
             get-rename-list
-            set-rename-list!
+            reset-rename!
             rename-all))
 
 ;;; List of renamings
@@ -34,6 +34,9 @@
         (lambda (ls) (set! rename-ls ls) rename-ls))
   (set! %get-rename-list
         (lambda () rename-ls)))
+
+;;; Reset the list of renamings to empty
+(define (reset-rename!) (set-rename-list! '()))
 
 ;;; Getters for rename pairs
 (define source car)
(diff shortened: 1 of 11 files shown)


vzh:
@peter-b A few notes:
- `rename<?` is not used any more (the code is already inside other func) and is broken, so please remove it
- I don't think `log! 'debug` messages are useful for anybody working on the Scheme part as well as for users, so IMO it would be better to throw them out.
I am happy with all other changes.

peter-b:
@vzh Thank you for the feedback!  I've addressed your points by removing `rename<?` and the commented out debugging calls.

vzh:
@peter-b You're welcome!
BTW, I'm curious, why have you replaced `string=?` with `string=` in last function? Do you think it is some faster, or it was just an oversight? No, I don't ask of changing anything, just wondering...

peter-b:
@vzh Um, that wasn't intentional! I meant to write `string=?`, but I didn't realise I had made the mistake because it works!  Maybe it's a synonym in Guile for some reason?

vzh:
@peter-b `string=?` is more general (R5RS) and can take many args while `string=` is from srfi-13 which is built-in in guile and can take only 2 args

peter-b:
@vzh I wonder if `string=` may be _very slightly_ more efficient because it is probably an opcode in Guile? Oh well, in that case it is not incorrect, so I will leave it the way it is.

vzh:
@peter-b I haven't dug so much :-) Let it be as is. I just prefer predicates to have `?` at their end.

Issue #39, opened 2017-02-14
Distinguishing between lepton and geda-gaf


peter-b:
lepton currently installs and uses a lot of documentation, programs and other files that clash with those installed and used by `geda-gaf`.

We want to be backward compatible with `geda-gaf` <= 1.9.2, but also want users to be able to install a newer geda-gaf alongside lepton.  Also, spending a lot of time on renaming things is busy-work that does not really bring much benefit to users.

We should consider gradually migrating towards naming things based on `lepton` in stages.  The current plan:

* [x] Change the tarball name in `configure.ac`, so we can start making releases (#47)
* [x] Load config data from gEDA config files and save back to Lepton config files (The issue is no longer valid since the last release 1.9.10 don't even use *geda.conf* files and transition utility `lepton-upcfg` is available.)
* [x] Add a `--enable-compat-symlinks` flag to `configure`.  This will enable installing a symlink named `gaf` for `lepton-cli`, etc.  It will default to `yes`
* [x] Start renaming tools (etc.) with `lepton`
* [x] Eventually change the default to `--enable-compat-symlinks=no`
* [x] Update documentation (especially [README](https://github.com/lepton-eda/lepton-eda/tree/master/README.md))

Unresolved issues:

* [x] How will we deal with installed resources, e.g. symbol libraries? (We won't do anything special until some new very different strategy is elaborated, just keep things as is, backwards compatible.  It doesn't break anything.)
* [x] How will we deal with user & system Scheme init files, e.g. `$XDG_CONFIG_HOME/gEDA/gafrc`?

Other related tasks:

* [x] Rename the Coverity Scan project (#43)

vzh:
@peter-b I propose to read Scheme rc files as is in their current
paths. There is nothing to do with resource libraries until we'll
implement them other way and add settings for them to new config.

We could add some defines like this:

> /*! Filename for old gEDA system configuration files */
> #define OLD_SYSTEM_CONFIG_NAME "geda-system.conf"
> /*! Filename for gEDA user configuration files */
> #define OLD_USER_CONFIG_NAME "geda-user.conf"
> /*! Filename for gEDA local configuration files */
> #define OLD_LOCAL_CONFIG_NAME "geda.conf"
> /*! Filename for gEDA system configuration files */
> #define SYSTEM_CONFIG_NAME "lepton-system.conf"
> /*! Filename for gEDA user configuration files */
> #define USER_CONFIG_NAME "lepton-user.conf"
[...]

PR #67, opened 2017-02-21
Rename libgedacairo to libleptonrenderer.

### LINKED ISSUE #54: Rename libgedacairo to libleptonrenderer
Our `libgedacairo` actually uses the same _code_ as upstream geda-gaf at the moment, but when compiled it has a dependency on `liblepton` rather than `libgeda`. We should rename it to `libleptonrenderer` or `libleptoncairo` or something.

~~The current split between `libgedacairo` and `liblepton` seems a little arbitrary.  Maybe it would make sense to merge `libgedacairo` into `liblepton` rather than rename the library.  I can't see any immediate downside to this.~~

This blocks v1.9.3 because of the soname conflict with geda-gaf.

vzh:
This branch is to close #54 and to address renaming of core libraries in #39.
The version number of the renamed library has been reset to 1.0.0.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,7 +1,7 @@
 if ENABLE_GATTRIB
 GATTRIB_DIR=gattrib
 endif
-SUBDIRS = liblepton libgedacairo gaf gschem ${GATTRIB_DIR} \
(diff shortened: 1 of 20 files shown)


vzh:
I've rebased and renamed the branch, and the issue is now addressed in #69.

PR #73, opened 2017-02-24
Disable build of contrib directory

gareth8118:
Add autotools control of whether to build the tools in the contrib/
directory using '--enable-contrib' and default it to disabled.


vzh:
@gareth8118 I think, eventually, we'll move contrib away from the repo. It would probably be good to have one repo per utility.

PR #78, opened 2017-02-28
Fix several cast errors found during trying to buld with g++.

vzh:


peter-b:
Note that if we do want to start routinely compiling Lepton with a C++ compiler, we'll want to go through and gradually start replacing all the C-style casts `(T) x` with the more explicit C++ casts (`dynamic_cast<T>(x)`, `const_cast<T>(x)`, `static_cast<T>(x)` and `reinterpret_cast<T>(x)`.  It will provide much more useful compiler messages and will be easier to figure out exactly what's happening each place that a cast is being used.

vzh:
@peter-b Last time I wrote something in C++ was about 20 years ago ;-) I just found it useful to compile with g++. It revealed some issues for me. WDYT of the PR itself?

PR #84, opened 2017-03-03
Add casts for gobject and its class

vzh:
This branch fixes another couple of errors reported by g++.

### DIFF
--- a/gschem/src/gschem_action.c
+++ b/gschem/src/gschem_action.c
@@ -158,7 +158,7 @@ static void gschem_action_class_init (GschemActionClass *klass)
   gobject_class->set_property  = gschem_action_set_property;
   gobject_class->get_property  = gschem_action_get_property;
 
-  gschem_action_parent_class = g_type_class_peek_parent (klass);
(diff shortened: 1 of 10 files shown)


vzh:
@peter-b Good point! I didn't even think of this :-) I'll try to review the code.

vzh:
@peter-b Some parent classes are _Gtk_ classes having no such define macros like `G_OBJECT`. I assume there is no need to make them? And what to do in such cases: add explicit casts (e.g. for `GtkActionClass`) or leave `G_OBJECT_CLASS` as it is now. WDYT?

peter-b:
The `G_DEFINE_TYPE` macro is a bit weird.  Some of the arguments it takes are treated as *strings* rather than as symbols.  So, for example, at the top of `gschem_action.c`, you could write:

```c
G_DEFINE_TYPE (GschemAction, gschem_action, GTK_TYPE_ACTION)
```

vzh:
@peter-b I added a new commit and rebased the branch. All seem to work now. Please look if I did TRT. I have not changed two files (x_*.c) since they need more TLC, I think.
P.S. Thanks for your last comment that enlightened me :-)

PR #86, opened 2017-03-04
Add casts for various Scheme functions in C code

vzh:
This branch fixes the set of errors reported by g++ related to using of Scheme functions in C code.

### DIFF
--- a/gnetlist/src/s_traverse.c
+++ b/gnetlist/src/s_traverse.c
@@ -424,7 +424,7 @@ s_traverse_net (NET *nets, int starting, OBJECT *object, char *hierarchy_tag, in
 
 
 static void
-init_module_gnetlist_core_traverse ()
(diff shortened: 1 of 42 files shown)


gnetlist/src/s_traverse.c
```diff
@@ -440,6 +440,6 @@ s_init_traverse ()
 {
   /* Define the (gsymcheck core check) module */
   scm_c_define_module ("gnetlist core traverse",
-                       init_module_gnetlist_core_traverse,
```
peter-b: Maybe it is "more correct" to change the definition of `init_module_gnetlist_core_traverse()` to:

```C
static void
init_module_gnetlist_core_traverse (void *unused)
{
    /* ... */
}
```

vzh:
@peter-b I've added a commit addressing your suggestion.

vzh:
... and added fix for one module I missed before.

PR #89, opened 2017-03-05
Cast some object using the EDA_RENDERER() macros.

vzh:


vzh:
I've included this commit in #94.

PR #90, opened 2017-03-06
Fix errors in search-rename() function and add inline documentation for it

vzh:
The regression which has been fixed here is present in 1.9.3.


vzh:
@peter-b This is my fault, too, since I reviewed your changes ;-)

PR #91, opened 2017-03-07
Add support for REPL readline history in gnetlist and gschem.

vzh:
New module (geda repl) has been introduced, which is used both by
gnetlist and gschem.  The same readline history file
".lepton_history" residing in the user configuration directory is
used for all tools using this module.

The REPL history is saved if the user quits the REPL explicitly,
or, in gschem, if it closes gschem. The latter is not recommended
though, since it garbles terminal settings and the user needs to
`reset` the terminal to make it functional again.

Loading the module (ice-9 session) in REPL adds support for such
procedures as `apropos' and `help'. Loading (system repl command)
adds support for guile metacommands.


peter-b:
I'm not totally comfortable with this, because it runs the REPL in a separate thread and the Lepton data structures aren't actually thread safe at the moment.  If gschem decided to redraw while you were modifying the page from another thread, for example, you could easily have a data race.

vzh:
@peter-b Some graphics libraries (`libpango` and such) are not thread-safe, anyway. I've learned this hard way. So it is not recommended to run in REPL something affecting graphics. OTOH, I don't add REPL here, it was added over a year ago. Here I just add normal `readline` support to what is already in the repository, and hopefully that should improve user experience.

peter-b:
@vzh I agree, we should put this in, but also I will think about how to make `liblepton` thread-safe!

PR #95, opened 2017-03-11
Fix gattrib errors gotten during compilation with g++

vzh:


vzh:
Fixed GCallback casts (I certainly did it wrong in initial patchset) and updated the branch with numerous fixes of gattrib related g++ warnings.

gattrib/src/x_window.c
```diff
@@ -115,7 +115,7 @@ x_window_init()
   /* This basically sets up the overhead for the sheets, as I understand
    * it.  The memory for the actual sheet cells is allocated later,
    * when gtk_sheet_new is invoked, I think.  */
-  sheets = g_malloc0(NUM_SHEETS * sizeof(GtkWidget *));
+  sheets = (GtkSheet**) g_malloc0 (NUM_SHEETS * sizeof (GtkSheet*));
```
peter-b: I tend to use the [`g_new0()` macro](https://developer.gnome.org/glib/2.50/glib-Memory-Allocation.html#g-try-new0) in situations like this. It does the required cast, so you don't have to type out the typename twice:

```c
sheets = g_new0(GtkSheet*, NUM_SHEETS);
```

gattrib/src/s_attrib.c
```diff
@@ -94,9 +94,11 @@ char *s_attrib_get_refdes(OBJECT *object)
   OBJECT *slot_text_object;
 
   /*------ Try to get the refdes -----*/
-  temp_uref = o_attrib_search_object_attribs_by_name (object, "refdes", 0);
+  temp_uref =
+    o_attrib_search_object_attribs_by_name (object, (char*) "refdes", 0);
```
peter-b: I feel like the correct thing to do here is to make `o_attrib_search_object_attribs_by_name()` take a `const char*` as its second argument.

gattrib/src/s_sheet_data.c
```diff
@@ -134,7 +134,7 @@ void s_sheet_data_add_master_comp_list_items (const GList *obj_list) {
 	printf("      In s_sheet_data_add_master_comp_list_items; found component on page\n");
 	printf(". . . . complex_basename = %s.\n", o_current->complex_basename);
 #endif
-	verbose_print(" C");
+	verbose_print ((char*)" C");
```
peter-b: Same here; there's no reason `verbose_print()` should ever modify the string you pass it, so the parameter should probably be explicitly `const char*`.

vzh:
@peter-b I seem to have fixed all the three issues (and a bit more :-)). Had to reorder some patches so some old ones appeare as new.

PR #96, opened 2017-03-13
Fix g++ build errors for several utility tools

vzh:
Another C++ compatibility patchset fixing g++ reported errors in gsch2pcb, gnetlist, and smash_megafile.

### DIFF
--- a/gnetlist/src/s_traverse.c
+++ b/gnetlist/src/s_traverse.c
@@ -357,7 +357,7 @@ s_traverse_net (NET *nets, int starting, OBJECT *object, char *hierarchy_tag, in
 
   if (object->type == OBJ_PIN) {
 
-    verbose_print (starting ? "p" : "P");
+    starting ? verbose_print ("p") : verbose_print ("P");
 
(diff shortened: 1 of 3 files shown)


gnetlist/src/s_traverse.c
```diff
@@ -357,7 +357,7 @@ s_traverse_net (NET *nets, int starting, OBJECT *object, char *hierarchy_tag, in
 
   if (object->type == OBJ_PIN) {
 
-    verbose_print (starting ? "p" : "P");
+    starting ? verbose_print ("p") : verbose_print ("P");
```
peter-b: What was the warning that this change fixes?  The original code looks okay to me, so I must be missing something important!

vzh: Namely `error: invalid conversion from ‘const char*’ to ‘char*’`. An error, not a warning as in other string conversion cases.

vzh: I'm going to throw it out, anyway, hopefully in near future.

PR #97, opened 2017-03-13
Fix various g++ warnings

vzh:
This branch fixes several warnings reported by g++ in liblepton, libleptonrenderer and gaf.


liblepton/src/g_register.c
```diff
@@ -43,26 +43,26 @@ struct gsubr_t {
 
 /*! \brief */
 static struct gsubr_t libgeda_funcs[] = {
-  { "eval-protected",            1, 1, 0, (SCM (*) ()) g_scm_eval_protected },
-  { "eval-string-protected",     1, 0, 0, (SCM (*) ()) g_scm_eval_string_protected },
+  { (char*) "eval-protected",            1, 1, 0, (SCM (*) ()) g_scm_eval_protected },
```
peter-b: Probably you can fix all of these by changing `gsubr_t::name` from `char* name` to `const char* name`.

liblepton/src/g_rc.c
```diff
@@ -760,8 +760,8 @@ SCM g_rc_reset_component_library(void)
 SCM g_rc_attribute_promotion(SCM mode)
 {
   static const vstbl_entry mode_table[] = {
-    {TRUE , "enabled" },
-    {FALSE, "disabled"},
+    {TRUE , (char*) "enabled" },
```
peter-b: I think it's best to change the definition of the `struct vstbl_entry` so that it uses `const char*` rather than `char*`.

liblepton/src/geda_complex_object.c
```diff
@@ -911,13 +911,17 @@ o_complex_check_symversion(TOPLEVEL* toplevel, OBJECT* object)
   g_return_if_fail (object->complex != NULL);
 
   /* first look on the inside for the symversion= attribute */
-  inside = o_attrib_search_inherited_attribs_by_name (object, "symversion", 0);
+  inside = o_attrib_search_inherited_attribs_by_name (object,
```
peter-b: Same as [this comment](https://github.com/lepton-eda/lepton-eda/pull/95#discussion_r105783257) on #95; I think it would be better to change the prototype for `o_attrib_search_inherited_attribs_by_name()`.

vzh:
@peter-b What do you think about changing the prototype for `export_command_line__utf8_check` to take `const gchar*`, too?

liblepton/src/geda_complex_object.c
vzh: Fixed in #95.

vzh:
I've rebased the branch. The only commit changed is the last one in the series (was the first before).

vzh:
OK, fixed all rest structs using `const char*` in their definitions.

peter-b:
> @peter-b What do you think about changing the prototype for `export_command_line__utf8_check` to take `const gchar*`, too?

@vzh I think that's a great idea, go for it!

PR #98, opened 2017-03-13
Fix gschem g++ compilation errors

vzh:
This patchset fixes lots of cast errors and other errors and warnings related to gschem when compilation is carried out in C++ mode.


gschem/src/gschem_binding_integer.c
```diff
@@ -132,8 +132,8 @@ class_init (GschemBindingIntegerClass *klass)
 {
   GObjectClass *object_class = G_OBJECT_CLASS (klass);
 
-  klass->parent_class.update_model = update_model;
-  klass->parent_class.update_widget = update_widget;
```
peter-b: Aren't these functions already of the correct type?  It looks like they are already correct (taking `GchemBinding*` as only argument and returning `gboolean`).  What warning did you get?

https://github.com/vzh/lepton-eda/blob/2036a7df217c2faf199129b9d17c9cdb472bbc23/gschem/src/gschem_binding_integer.c#L68

gschem/include/prototype.h
```diff
@@ -234,7 +234,7 @@ void i_action_update_status(GschemToplevel *w_current,gboolean inside_action);
 void i_show_state(GschemToplevel *w_current, const char *message);
 void i_set_state(GschemToplevel *w_current, enum x_states newstate);
 void i_set_state_msg(GschemToplevel *w_current, enum x_states newstate, const char *message);
-void i_update_middle_button(GschemToplevel *w_current, void (*func_ptr)(), const char *string);
+void i_update_middle_button (GschemToplevel *w_current, void (*func_ptr) (gpointer, guint, GtkWidget*), const char *string);
```
peter-b: You have written out this function type many times in this patch series!  Maybe it should have a `typedef`?

gschem/src/x_compselect.c
```diff
@@ -79,7 +79,7 @@ compselect_get_view (Compselect *compselect)
     case 1: return VIEW_CLIB;   /* Component library page */
     default:
       g_critical ("compselect_get_view: Unknown tab position\n");
-      return 0;
+      return (enum compselect_view) 0;
```
peter-b: I suggest returning a specific member of `enum compselect_view` here, instead of using a cast.  I suggest `VIEW_CLIB`, since that's the default page in the component selector.  Since this is technically an error pathway, it doesn't matter what you return as long as it's a sensible default.

gschem/src/g_rc.c
```diff
@@ -128,8 +128,8 @@ SCM g_rc_gschem_version(SCM scm_version)
 SCM g_rc_net_direction_mode(SCM mode)
 {
   static const vstbl_entry mode_table[] = {
-    {TRUE , "enabled" },
-    {FALSE, "disabled"}
+    {TRUE , (char*) "enabled" },
+    {FALSE, (char*) "disabled"}
```
peter-b: As I mentioned elsewhere, usually it is best to fix these by changing `struct` definitions or function prototypes to use `const char*`.

gschem/src/x_autonumber.c
```diff
@@ -342,7 +342,7 @@ gint autonumber_match(AUTONUMBER_TEXT *autotext, OBJECT *o_current, gint *number
   if (g_str_has_suffix (str,"?")) {
     isnumbered = 0;
     /* There must not be any character between the "?" and the searchtext */
-    if (strlen(str) != len+1)
+    if ((gint) strlen(str) != len+1)
```
peter-b: `len` comes from a call to `strlen()`, so it should be declared as `size_t len` instead of `int len`.  Then this cast would not be necessary, because the comparison would have `size_t` on both sides.

gschem/src/x_autonumber.c
```diff
@@ -700,7 +700,7 @@ void autonumber_text_autonumber(AUTONUMBER_TEXT *autotext)
 	      /* the beginnig of the current text matches with the searchtext now */
 	      /* strip of the trailing [0-9?] chars and add it too the searchtext */
 	      for (i = strlen (str)-1;
-		   (i >= strlen(searchtext))
+             (i >= (gint) strlen(searchtext))
```
peter-b: Maybe `i` should be `size_t`, since it comes from the return value of `strlen()`.

gschem/src/gschem_binding_integer.c
vzh: Something about casting specific to only two these functions. I cannot reproduce it any more (tried in several branches with various options I used before). Probably this has been fixed some other way. I'll revert the commit.

vzh:
@peter-b Some string conversion issues have been fixed in #95, so they will be done after rebase. Others are addressed in the last five commits.

gschem/src/g_rc.c
vzh: `vstbl_entry` as well as attrib search procedures have been fixed in #95.

gschem/src/x_autonumber.c
```diff
@@ -975,7 +985,7 @@ AUTONUMBER_TEXT *autonumber_init_state()
   if(autotext==NULL) return NULL;
 
   autotext->scope_text = NULL;
-  t=default_text;
+  t = (gchar**) default_text;
```
peter-b: Since `t` is an offset into `default_text`, I suspect it should be `const gchar **t`.

gschem/src/x_image.c
```diff
@@ -129,7 +129,7 @@ x_image_get_type_from_description (const char *description)
   GSList *ptr;
 
   if (strcmp (description, "Portable Document Format") == 0) {
-    return "pdf";
+    return (char*) "pdf";
```
peter-b: I think the buffer returned by `x_image_get_type_from_description()` is not owned by the caller, and because the buffer should not be modified or free'd by the caller.  Maybe it should return a `const char*` rather than a `char*`?

vzh: @peter-b I'll better add another PR to solve these two cast issues. Thank you for your review and for your patience with me!

PR #101, opened 2017-03-15
Enable build for all tools but gattrib in C++ mode

vzh:
Together with two previous unclosed PR's (#97 and #98) these commits ensure `gschem`, `gaf`, `gnetlist` and others (all but `gattrib`) compile without errors (and the first two without warnings on my machine) with *g++* and even work :-)
To ensure the build I replace `-std=gnu99` in *configure.ac* with `std=gnu++03` and do `./autogen.sh && ./configure --prefix=/my/home/dir/lepton CC=g++ && make -k CC=g++ all && make install`. Separate `make install` is to ensure the built programs are installed even if `gattrib` has not been built. Without `CC` on `configure` stage I have warnings from snarfing tools.

### DIFF
--- a/gschem/include/globals.h
+++ b/gschem/include/globals.h
@@ -26,9 +26,6 @@
 /* window list */
 extern GList *global_window_list;
 
-/* Manager for recently used files */
-GtkRecentManager *recent_manager;
-
 /* colors */
 extern GdkColor white;
 extern GdkColor black;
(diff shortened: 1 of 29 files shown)


peter-b:
@vzh When I added the `G_BEGIN_DECLS` and `G_END_DECLS` macros for compiling C++ applications that use `liblepton`, I put them in the toplevel header file (see [this commit](https://github.com/lepton-eda/lepton-eda/commit/2e379818a4b2ed0651b7d24f671bfe7dbac7fd5b)).  I don't know if it works to do the same for compiling `liblepton` itself. I think it is probably best to take your approach and have the `G_BEGIN_DECLS`/`G_END_DECLS` locally to each header file, so I suggest you revert my commit!

PR #108, opened 2017-03-22
Add change action mode hooks in gschem

vzh:
The hooks are run when the gschem action mode changes.
For example, if the user wants to show which mode is switched on, she can use:
> (add-hook! switch-action-mode (lambda (mode) (display mode) (newline)))

### DIFF
--- a/gschem/scheme/gschem/hook.scm
+++ b/gschem/scheme/gschem/hook.scm
@@ -49,3 +49,5 @@
 (define-public action-property-hook %action-property-hook)
 
 (define-public bind-keys-hook %bind-keys-hook)
+
+(define-public switch-action-mode-hook %switch-action-mode-hook)
--- a/gschem/src/g_hook.c
+++ b/gschem/src/g_hook.c
@@ -132,6 +132,32 @@ g_run_hook_page (GschemToplevel *w_current, const char *name, PAGE *page)
   scm_remember_upto_here_1 (expr);
 }
 
+/*! \brief Runs a change action mode hook.
+ * \par Function Description
+ * Runs a hook called \a name, which should expect the single #MODE \a
+ * mode as its argument.
+ *
+ * \param name name of hook to run
+ * \param mode #MODE argument for hook.
+ */
+void
+g_run_hook_action_mode (GschemToplevel *w_current,
+                        const gchar *name,
+                        const gchar *action_mode)
+{
+  scm_dynwind_begin ((scm_t_dynwind_flags) 0);
+  g_dynwind_window (w_current);
+
+  SCM expr = scm_list_3 (run_hook_sym,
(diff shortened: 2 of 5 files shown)


gschem/src/g_hook.c
```diff
@@ -179,6 +205,7 @@ init_module_gschem_core_hook (void *unused)
   DEFINE_HOOK ("%new-page-hook",1);
   DEFINE_HOOK ("%action-property-hook",3);
   DEFINE_HOOK ("%bind-keys-hook",3);
+  DEFINE_HOOK ("%switch-action-mode",1);
```
peter-b: Maybe the hook name should end with `-hook`, for consistency with other hook variables.

gschem/scheme/gschem/hook.scm
```diff
@@ -49,3 +49,5 @@
 (define-public action-property-hook %action-property-hook)
 
 (define-public bind-keys-hook %bind-keys-hook)
+
+(define-public switch-action-mode %switch-action-mode)
```
peter-b: I think it is necessary to add documentation to the [`(gschem hook)` module docs](https://github.com/lepton-eda/lepton-eda/blob/master/docs/scheme-api/geda-scheme.texi#L2385).

gschem/src/i_basic.c
```diff
@@ -207,6 +207,34 @@ void i_action_update_status (GschemToplevel *w_current, gboolean inside_action)
...
+      case MCOPYMODE: mode="multiple-copy-mode"; break;
+      case MOVEMODE: mode="move-mode"; break;
+      case PASTEMODE: mode="paste-mode"; break;
+      case TEXTMODE: mode="text-mode"; break;
+      case SBOX: mode="box-select-mode"; break;
+      case ZOOMBOX: mode="zoom-box-mode"; break;
+      case PAN: mode="pan-mode"; break;
+      case MIRRORMODE: mode="mirror-mode"; break;
+      case ROTATEMODE: mode="rotate-mode"; break;
+      default: mode="select-mode"; break;
```
peter-b: I suggest using a [`g_return_if_reached()`](https://people.gnome.org/~desrt/glib-docs/glib-Warnings-and-Assertions.html#g-return-if-reached) in the default case here, or have no default so that the compiler warns if the `switch` statement does not cover every possible value of `enum x_states`.  If you choose to have no `default:` label, I suggest initialising the `mode` variable to the appropriate default value, e.g.:

```c
const gchar* mode = "select-mode";
switch (newstate) {
    /* ... */
}
```

You don't need a `;` after the closing `}` of a `switch` statement.

vzh:
@peter-b I updated the branch with the changes you requested.

Thank you for review!

PR #111, opened 2017-03-29
Configurable text size in text entry dialog

### LINKED ISSUE #110: 'Text Entry...' dialog should respect 'text-size' option's value
Now it uses a hard-coded value - 12 (in `system-gschemrc` `text-size`
is set to 10). Also, it was found that the font size
checking code in `g_rc_text_size()` also uses hard-coded values
instead of `MINIMUM_TEXT_SIZE` and `DEFAULT_TEXT_SIZE` macros.
That was [reported](https://bugs.launchpad.net/geda/+bug/1565803) almost a year ago, but there's still no feedback.
It seems that the fix is still relevant. What do you think?
If need be, I can submit updated patches or make a pull request.

graahnul-grom:
Closes #110

### DIFF
--- a/gschem/src/x_newtext.c
+++ b/gschem/src/x_newtext.c
@@ -300,9 +300,6 @@ static void newtext_init(NewText *dialog)
 
   dialog->textsizecb = gschem_integer_combo_box_new();
 
-  gschem_integer_combo_box_set_value (dialog->textsizecb,
-                                      g_current_window()->text_size);
-
   gtk_table_attach_defaults(GTK_TABLE(table), dialog->textsizecb, 1,2,1,2);
 
   label = gtk_label_new (_("Alignment:"));
@@ -441,6 +438,9 @@ text_input_dialog (GschemToplevel *w_current)
 
     gschem_integer_combo_box_set_model (NEWTEXT (w_current->tiwindow)->textsizecb,
                            gschem_toplevel_get_text_size_list_store (w_current));
+
+    gschem_integer_combo_box_set_value (NEWTEXT (w_current->tiwindow)->textsizecb,
+                                        w_current->text_size);
 
     gtk_widget_show_all (w_current->tiwindow);
   }
--- a/gschem/src/x_newtext.c
+++ b/gschem/src/x_newtext.c
@@ -75,7 +75,8 @@ dialog_response_apply (NewText *dialog)
   int align = LOWER_LEFT;
   int color = TEXT_COLOR;
   int rotate = 0;
-  int size = 12;
+  GschemToplevel *w_current = GSCHEM_DIALOG (dialog)->w_current;
+  int size = w_current ? w_current->text_size : default_text_size;
   gchar *string = NULL;
   gchar *tmp = NULL;
   GtkTextBuffer *textbuffer;
--- a/gschem/src/x_newtext.c
+++ b/gschem/src/x_newtext.c
@@ -292,7 +292,11 @@ static void newtext_init(NewText *dialog)
                     0);
 
   dialog->textsizecb = gschem_integer_combo_box_new();
-  gschem_integer_combo_box_set_value(dialog->textsizecb, 12);
+
+  GschemToplevel *w_current = g_current_window ();
+  gschem_integer_combo_box_set_value (dialog->textsizecb,
+    w_current ? w_current->text_size : default_text_size);
+
   gtk_table_attach_defaults(GTK_TABLE(table), dialog->textsizecb, 1,2,1,2);
 
   label = gtk_label_new (_("Alignment:"));
--- a/gschem/src/x_newtext.c
+++ b/gschem/src/x_newtext.c
@@ -72,11 +72,17 @@ struct _NewText {
 static void
 dialog_response_apply (NewText *dialog)
 {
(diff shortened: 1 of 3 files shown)


gschem/src/x_newtext.c
```diff
@@ -292,7 +292,11 @@ static void newtext_init(NewText *dialog)
                     0);
 
   dialog->textsizecb = gschem_integer_combo_box_new();
-  gschem_integer_combo_box_set_value(dialog->textsizecb, 12);
+
+  GschemToplevel *w_current = g_current_window ();
+  gschem_integer_combo_box_set_value (dialog->textsizecb,
+    w_current ? w_current->text_size : default_text_size);
```
vzh: `g_current_window()` never returns `NULL` so this check is pointless.

graahnul-grom: If so, it should be explicitly stated in function's documentation.

gschem/src/x_newtext.c
```diff
@@ -75,7 +75,8 @@ dialog_response_apply (NewText *dialog)
   int align = LOWER_LEFT;
   int color = TEXT_COLOR;
   int rotate = 0;
-  int size = 12;
+  GschemToplevel *w_current = GSCHEM_DIALOG (dialog)->w_current;
```
vzh: Please avoid using of direct links to data structures with *gobjects*. They have better means.
Use `GSCHEM_DIALOG`'s property to get `w_current`. E.g.
>  GschemToplevel *w_current;
>  int size;
>   g_return_if_fail (dialog != NULL);
>  g_object_get (GSCHEM_DIALOG (dialog), "gschem-toplevel", &w_current, NULL);
>  g_return_if_fail (w_current != NULL);
>  size = w_current->text_size;

Thus you can check inappropriate usage of the *dialog_response_apply()* on the build stage.
If you're planning to somehow use this dialog without `w_current`, you should check it and exit if it is `NULL` (I don't think it is good idea just now).

In future it would be better to use `gschem_options` to store various settings, I think.

gschem/src/x_newtext.c
```diff
@@ -292,7 +293,10 @@ static void newtext_init(NewText *dialog)
                     0);
 
   dialog->textsizecb = gschem_integer_combo_box_new();
-  gschem_integer_combo_box_set_value(dialog->textsizecb, 12);
+
+  gschem_integer_combo_box_set_value (dialog->textsizecb,
+                                      g_current_window()->text_size);
```
vzh: Please don't use third party objects in the dialog init function. It should add initial values that apply to the dialog only. Other values you can set on new `gobject` instance of the dialog.

Please don't use Scheme extension code inside dialog C code. Here I mean `g_current_window`. I and some devs don't want to mix it.

Since we don't have appropriate properties for text size in `NewText` yet (which has to be fixed in future, I think), you can do the same direct assigning in the function `text_input_dialog()` where new dialog objects are instantiated. For example:
> gschem_integer_combo_box_set_value (NEWTEXT (w_current->tiwindow)->textsizecb,
>   w_current->text_size);

gschem/src/x_newtext.c
vzh: First somebody should go into the code and clearly understand what it does, I think ;-)

graahnul-grom: Yes, I do not 100% understand gschem code yet, especially guile integration stuff. And why bother writing doxygen comments at all? Maybe "go figure yourself how it works, comrade!" would be enough? :)

gschem/src/x_newtext.c
graahnul-grom: Agree, it's better to use properties. Then, `x_multiattrib.c`, `x_pagesel.c` and `x_compselect.c` should be fixed in the same manner?

gschem/src/x_newtext.c
vzh: Current developers usually fix docs if they rewrite some parts of code. However, we have not enough man power to cover all the code base, especially non-modified long unmaintained parts.

gschem/src/x_newtext.c
vzh: Source files with obsolete naming scheme, e.g. which names start with `x_`, are those having many things to fix. @ehennes775 has fixed many dialogs (adding properties, classes and so on) and renamed them using new naming scheme (e.g. *gschem* code file names start with `gschem_`. You can use his code as an example of refactoring. Now that naming scheme is of course obsolete, too, and we have to come up with a new scheme. I think probably something like e.g. `lepton_multiattrib.c` would be suitable.

gschem/src/x_newtext.c
graahnul-grom: Thank you for helping me.

gschem/src/x_newtext.c
graahnul-grom: Speaking of refactoring, would it be possible to [do it](https://github.com/graahnul-grom/lepton-eda/commit/6367613187244a4bed046541a243775bcc3e16a8) for `x_window_create_main()`?
The code there in its current state makes it almost impossible to experiment with
main window layout and widgets.

vzh: Yes, of course. Though I would prefer lesser hunks. Please do a PR.

PR #122, opened 2017-05-13
log! scheme api docs; g_current_window() doxygen docs

graahnul-grom:


docs/scheme-api/geda-scheme.texi
```diff
@@ -2055,6 +2055,8 @@ The @var{level} should be one of the symbols @code{error},
 @code{critical}, @code{warning}, @code{message}, @code{info} or
 @code{debug}.
 
+Note: if @var{level} is set to @code{'error}, main program will terminate.
```
vzh: According to the style of previous sentences, the apostrophe symbol (') is unnecessary here :-)

schematic/src/g_window.c
```diff
@@ -124,7 +124,7 @@ SCM_DEFINE (current_window, "%current-window", 0, 0, 0,
  * \brief Get the value of the #GschemToplevel fluid.
  * \par Function Description
  * Return the value of the #GschemToplevel fluid in the current dynamic
- * context.
+ * context. Never returns NULL.
```
vzh: I could even say: "Signals an error if there is no valid window fluid or the fluid value is NULL. Never returns NULL." WDYT?

docs/scheme-api/geda-scheme.texi
graahnul-grom: yes

schematic/src/g_window.c
graahnul-grom: much better

PR #126, opened 2017-05-16
gschem: Do not warn about missing menu by default.

graahnul-grom:
Log window is cluttered with 'Tried to set the
sensitivity on non-existent menu item' messages
when one of the standard menus in system-gschemrc
is commented out. Do log only when DEBUG is defined.


vzh:
Please use `g_debug` instead. I think users should have a possibility to see debug messages without recompilation.

> G_MESSAGES_DEBUG=all lepton-schematic

PR #130, opened 2017-05-26
symcheck FreeBSD fixes

graahnul-grom:


vzh:
The issue with `guile` is that some users may have several versions installed. On one of my systems I have `guile` pointing to `guile-2.2` and I don't know yet if there may be any compilation/work issues with `lepton-symcheck` in such configurations. Probably this was the reason why I used such a call. I'll check and tell you soon, I hope.

vzh:
OK, just tried what happens on the system with two `guile` versions installed and your changes applied.
> $ guile -c '(display (version)) (newline)'
> 2.2.2
> lepton-symcheck

And it barfs on geda/log.scm compilation:
> ;;; note: auto-compilation is enabled, set GUILE_AUTO_COMPILE=0
> ;;;       or pass the --no-auto-compile argument to disable.
> ;;; compiling /path/to/lepton/share/gEDA/scheme/symcheck/check.scm
> ;;; compiling /path/to/lepton/share/gEDA/scheme/geda/log.scm
> ;;; WARNING: compilation of /path/to/lepton/share/gEDA/scheme/geda/log.scm failed:
> ;;; ERROR: Wrong type to apply: #<unknown-type (0x24f . 0xb6c37288) @ 0xb6cdf9e0>
> ;;; WARNING: compilation of /path/to/lepton/share/gEDA/scheme/symcheck/check.scm failed:
> ;;; ERROR: no binding `user-config-dir' in module (geda os)
> Backtrace:
>            6 (apply-smob/1 #<catch-closure 9d3e950>)
> In ice-9/boot-9.scm:
>     713:2  5 (call-with-prompt _ _ #<procedure default-prompt-handler (k proc)>)
> In ice-9/eval.scm:
>     619:8  4 (_ #(#(#<directory (guile-user) 9d60910>)))
> In ice-9/boot-9.scm:
>    2316:4  3 (save-module-excursion _)
>   3823:12  2 (_)
> In /path/to/lepton/bin/lepton-symcheck:
>     32:15  1 (_)
> In unknown file:
>            0 (_)
> 
> ERROR: ERROR: Wrong type to apply: #<unknown-type (0x24f . 0xb6c37288) @ 0xb6cded20>

The culprit in this case is the built-in function `%log` defined in C code and wrapped in Scheme during compilation with `guile-2.0`, IIUC.

I haven't tried to compile Lepton with guile-2.2 yet and don't know if it is possible. Anyway, AFAIUI, we have two possibilities here now to solve the issue:
- "hardwire" guile full path in `lepton-symcheck` using info from `configure` stage;
- try to make Lepton guile-2.2 compatible

I think the former option is easier (though the latter would be better :-))

WDYT?

vzh:
@graahnul-grom Seems, I have found a solution for my issues with two Guile versions installed. Please check out #131 and try if it works on your system.

vzh:
And one more thought. Could you please add the info about portable Makefile instructions you used into *CONTRIBUTING.md*?

PR #131, opened 2017-05-27
build: Allow build with guile-2.2.

vzh:
This partially reverts commit
bc61db71e1d20d43ab3bf190c99e76521bf6b0ac with new versions used,
which allows compilation with guile-2.2.


graahnul-grom:
Yes, it works on my system (guile 2.0.14).

vzh:
While the tools compile properly with *guile-2.2* and pass all tests, the *lepton-schematic* GUI doesn't work as expected, so I think we'll defer this PR until we make sure the GUI is working with *guile-2.2*.

PR #135, opened 2017-06-04
schematic: Add support for showing widgets as dialogs.

graahnul-grom:
Now that we have the code for widgets management
refactored, it would be nice (and easy) to make a
one more step: at last, let the users decide,
what type of GUI they prefer to use, dialog boxes
(as it was before 1.9.2) or docking widgets.

Sorry to say that, but for me, as a user,
introduction of docking widgets was a complete show-stopper;
since then I have to maintain the local fork
(now extremely outdated). I would be more than happy
(and, I think, not only me) if we could make such a change. :-)
What do you think?

### DIFF
--- a/schematic/src/x_widgets.c
+++ b/schematic/src/x_widgets.c
@@ -1,6 +1,4 @@
 /* Lepton EDA Schematic Capture
- * Copyright (C) 1998-2010 Ales Hvezda
- * Copyright (C) 1998-2016 gEDA Contributors (see ChangeLog for details)
  * Copyright (C) 2017 dmn <graahnul.grom@gmail.com>
  *
  * This program is free software; you can redistribute it and/or modify
@@ -68,17 +66,17 @@ gboolean x_widgets_use_docks()
  * Initialize widgets management.
  * Call this before any other functions from this file.
  * This function reads the value of "use-docks" configuration
- * setting in "gschem.gui" group, which determines
+ * setting in "schematic.gui" group, which determines
  * if widgets will be shown in docks (if true) or as
- * a dialog boxes (if false).
+ * dialog boxes (if false).
  *
  * Configuration setting description:
  * key:   use-docks
- * group: gschem.gui
+ * group: schematic.gui
  * type:  boolean
  * default value: true
  *
- * \return TRUE if use-docks options is set to true, FALSE otherwise.
+ * \return TRUE if use-docks option is set to true, FALSE otherwise.
  */
 void x_widgets_init()
 {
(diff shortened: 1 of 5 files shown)


schematic/src/x_widgets.c
```diff
@@ -39,18 +38,123 @@
...
+
+
+/*! \par Function Description
+ *
+ * Initialize widgets management.
+ * Call this before any other functions from this file.
+ * This function reads the value of "use-docks" configuration
+ * setting in "gschem.gui" group, which determines
+ * if widgets will be shown in docks (if true) or as
+ * a dialog boxes (if false).
```
vzh: Article 'a' is superfluous here.

schematic/src/x_widgets.c
```diff
@@ -39,18 +38,123 @@
...
+ *
+ * Configuration setting description:
+ * key:   use-docks
+ * group: gschem.gui
```
vzh: Please use new name of the program, i.e. "schematic".

schematic/src/x_widgets.c
```diff
@@ -39,18 +38,123 @@
...
+
+  EdaConfig* cfg = eda_config_get_context_for_path (cwd);
+
+  g_free (cwd);
+
+  if (cfg != NULL)
+  {
+    GError* err = NULL;
+    gboolean val = eda_config_get_boolean (cfg,
+                                           "gschem.gui",
```
vzh: Please use new name of the program, i.e. "schematic".

schematic/src/x_widgets.c
```diff
@@ -114,3 +262,57 @@ x_widgets_show_in_dock (GtkWidget* wbook, GtkWidget* widget)
...
+  if (*dialog != NULL)
+  {
+    gtk_window_present (GTK_WINDOW (*dialog));
+    return;
+  }
+
+  GtkWidget* dlg = gschem_dialog_new_with_buttons(
+    title,
+    GTK_WINDOW (w_current->main_window),
+    0,
```
vzh: When compiling using `g++`, it reports a missing cast to *GtkDialogFlags* here. Please fix this.

vzh:
Yeah, many users requested reverting to old dialog style. You've done even better, thanks! I am pretty happy with your changes. :+1:

graahnul-grom:
Maybe something wrong with Travis CI?
`ERROR: Unbound variable: <html>`

vzh:
Yes, I've restarted CI build and now is all OK. Thanks.

PR #138, opened 2017-06-12
schematic: Add filtering support to hotkeys dialog.

graahnul-grom:


vzh:
@graahnul-grom Could you please describe in more details what it is for?

graahnul-grom:
There are 137 default key bindings. Users may have a set of their own (I have 20+). It would be very useful to quickly search by either key or action name instead of grep'ing `system-gschemrc` and local `gschemrc` files.

vzh:
@graahnul-grom I'm pretty happy with your patch. The only question: I see it is necessary to hit space between keys to see correct result. Is it intended? If so, why? For me, it looks inconsistently. I would prefer to hit 'aa' instead of 'a a' to see the result.
Thanks.

graahnul-grom:
It's intentional. Search is performed not only in "keystroke" filed, but also in "action" field, where it (obviously) should be done literally. And I think it's more natural to search for keys as they are displayed in this dialog and menus, passed to keybinding function and referred to in documentation (i.e. "A A", not "AA").
As far as I know, that's the way a similar function works in other program products.

vzh:
OK, thank you. We have to document it somewhere, right?

graahnul-grom:
> We have to document it somewhere

Maybe "Filter:" label next to the edit box will be sufficient?

vzh:
I meant documenting the way how users should search for keystrokes, that is, by typing spaces between keys. One thought is to add a tooltip for the filtering entry. WDYT?
The label `Filter:` would not harm, too, I believe ;-)

PR #165, opened 2017-12-12
schematic: Use new name for config group.

graahnul-grom:
Place "modify-viewport" configuration key (introduced in
commit e050f01) in the "schematic.undo" config group.
Initially, the groups was named "gschem.undo".

### DIFF
--- a/schematic/src/o_undo.c
+++ b/schematic/src/o_undo.c
@@ -65,13 +65,13 @@ void o_undo_init(void)
  * \par Function Description
  *
  * This function reads the value of "modify-viewport" configuration
- * setting in "gschem.undo" group, which determines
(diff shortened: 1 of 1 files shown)


graahnul-grom:
Speaking of configuration, I've compiled
the list of `lepton-schematic` settings
defined in `system-gschemrc` in my [wiki](https://github.com/graahnul-grom/lepton-eda/wiki).

As a first step in migration process I propose
to start defining group names, key names, types
and possible values for current settings.

Maybe we should start a new [discussion](https://github.com/orgs/lepton-eda/teams/lepton-developers).
Then, I can update the wiki page with the results.

vzh:
@graahnul-grom Great! Please go for it. Feel free to use lepton wiki for this.

Issue #172, opened 2018-02-14
Single-Pin-Net in netlist for unconnected pins


mario-giovinazzo:
I have an unconnected pin generating DRC error.
Connecting this pins to NC symbol (sym/misc/nc-right-1.sym) solves but generate a Single-Pin-Net in netlist that I must manually remove.
Is there a way to avoid this?
Thanks

graahnul-grom:
@mario-giovinazzo Could you please provide us with some test schematics and `lepton-netlist` output?

mario-giovinazzo:
This is a simple no-sense circuit,  pins U1-10 and U1-12 are unconnected.
![image](https://user-images.githubusercontent.com/30755972/36280184-c2c2aa58-1299-11e8-8d7d-c189997c4caa.png)
Running “lepton-netlist -g drc2” will produce errors:
```
Checking non-numbered parts...
Checking duplicated references...
Checking NoConnection nets for connections...
Checking nets with only one connection...
Checking pins without the 'pintype' attribute...
Checking type of pins connected to a net...
Checking unconnected pins...
ERROR: Unconnected pin U1:10
ERROR: Unconnected pin U1:12
Checking slots...
Checking duplicated slots...
Checking unused slots...
No warnings found.
Found 2 errors.
```
And running “lepton-netlist -g geda” will produce a correct netlist:

```
START header
gEDA's netlist format
Created specifically for testing of gnetlist
END header

START components
CONN1 device=CONNECTOR_4
D1 device=LED
D2 device=LED
D3 device=LED
D4 device=LED
R1 device=RESISTOR
R2 device=RESISTOR
R3 device=RESISTOR
R4 device=RESISTOR
U1 device=7414
END components

START renamed-nets
END renamed-nets

START nets
GND : U1 7, U1 11, U1 13
NN_net1 : CONN1 3, U1 5
NN_net2 : CONN1 2, U1 3
NN_net3 : CONN1 4, U1 9
NN_net4 : CONN1 1, U1 1
NN_net5 : D2 2, U1 4
NN_net6 : D3 2, U1 6
NN_net7 : D4 2, U1 8
NN_net8 : D1 2, U1 2
NN_net9 : D1 1, R1 1
NN_net10 : D2 1, R2 1
NN_net11 : D3 1, R3 1
NN_net12 : D4 1, R4 1
Vcc : R1 2, R2 2, R3 2, R4 2, U1 14
END nets
```
[...]

PR #176, opened 2018-02-21
gnetlist: Delete nets with "no connect" symbol.

graahnul-grom:
Issue #172.

### DIFF
--- a/gnetlist/scheme/gnetlist.scm
+++ b/gnetlist/scheme/gnetlist.scm
@@ -851,19 +851,127 @@ Lepton EDA homepage: <https://github.com/lepton-eda/lepton-eda>
   (primitive-exit 0))
 
 
-(define (set-toplevel-schematic! files netlist-mode)
+
+; private:
+; function: nets-delete-not-connected()
+;
+; Delete nets if they have "no connect" symbol attached.
+; lepton-eda issue #172
+;
+; [netnames]: list of net names
+; {return}:   updated list of net names
+;
+( define ( nets-delete-not-connected netnames )
+( let
+  (
+  ( ncobjs   #f ) ; list of "no connect" objects
+  ( verbose  #f )
+  ( result  '() ) ; empty list
+  )
+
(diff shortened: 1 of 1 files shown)


graahnul-grom:
@vzh If I understand correctly, this patch is completely unacceptable: should I close the PR? By no means do I consider myself a skilled Scheme programmer; I simply tried to help our user, @mario-giovinazzo, with #172.

graahnul-grom:
My thoughts on how it can be fixed:
1) introduce some global variable (option?) to determine
if *NC* symbols should be deleted
2) set that option to `TRUE` by default
3) check that variable in `gnetlist.scm` and if it's `TRUE`,
delete *NC* symbols
4) backends that need access to all nets can turn this option off
(e.g. `geda`, `drc2` backends)
5) modify misc/nc-*.sym to have only one attribute: `symbol=nc`,
check for its presence in nets-delete-not-connected()

What do you think?

vzh:
@graahnul-grom I would suggest adding of two different accessor functions (probably supplied with two corresponding fields in the `schematic` record) yielding appropriate filtered or non-filtered connection lists. Or one function with an optional argument for the same goal. The backend writers could opt the function/option they need. Most of the lepton backends could use filtered lists while some general ones could use non-filtered connection lists. A command line option key to force filtered or non-filtered output would be of help, too.
As for the *nc-\*.sym* files, on this stage we could just add the new attributes without breaking anything else. Then there is no need to fix special uses defined previously in, say, the `drc2` backend.

graahnul-grom:
@vzh The changes discussed above is implemented in the [tb_065_netlist_noconnect](https://github.com/graahnul-grom/lepton-eda/tree/tb_065_netlist_noconnect)
branch, please take a look.
Should I close this PR and open a new one?

graahnul-grom:
So, I'm closing this PR and will open a new one using the branch
mentioned in the previous comment. We may continue the discussion there.

PR #180, opened 2018-02-27
schematic: Fix: preserve zoom in tabbed GUI.

graahnul-grom:
Zoom level is preserved when the main window is resized.
Also preserve zoom level for already opened pages.


vzh:
Dmitry, it would be easier to get at the salt of your commits if you separate refactoring changes and new insertions by splitting them into distinct patches with appropriate description ;-)

graahnul-grom:
Agree. Will be more careful.

PR #182, opened 2018-03-11
Rename gsch2pcb

graahnul-grom:


vzh:
Eventually managed to get it working, ugh. Having some documentation about `--with-pcb-datadir` in INSTALL would not harm :-/
Thank you for your work!

PR #183, opened 2018-03-28
Do not hardcode file name in file save dialog

graahnul-grom:


vzh:
@graahnul-grom Thanks!

PR #200, opened 2018-04-22
Do not hardcode perl, python, bash paths

graahnul-grom:


graahnul-grom:
`man perl`:
> DIAGNOSTICS
>   The "use warnings" pragma produces some lovely diagnostics. One can
>   also use the -w flag, but its use is normally discouraged, because it
>   gets applied to all executed Perl code, including that not under your
>   control.

And how to pass `-w` flag while using `#!/usr/bin/env perl`?

vzh:
I saw it, as well as the following:
> SYNOPSIS
       env [OPTION]... [-] [NAME=VALUE]... [COMMAND [ARG]...]

Dunno, if the synopsis is valid for non-GNU `env` utilities, but the above description suggests we could use any ARG there.

graahnul-grom:
`#!/usr/bin/env perl -w` always produce this error:
`env: perl -w: No such file or directory`

On `FreeBSD` it can be solved with
`#!/usr/bin/env -S perl -w`

On `Linux` - I have no idea: there is no `-S` switch or something like this.

graahnul-grom:
Found out that there is no standardised way to pass more than one argument to `env`.
Discussed on [stackexchange](https://unix.stackexchange.com/questions/63979/shebang-line-with-usr-bin-env-command-argument-fails-on-linux?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa) and [stackoverflow](https://stackoverflow.com/questions/4303128/how-to-use-multiple-arguments-for-awk-with-a-shebang-i-e).
So I think we'd better do not change the current "shebang" lines in those scripts . If you have no objections, may I close #199?

vzh:
No objections as I've already merged it.

PR #201, opened 2018-04-22
Update README.md

graahnul-grom:


vzh:
I still hope to add anything we did so far to the NEWS file some time, make a PR, and then make an intermediate release. My real life issues (my new job, children, and so on) give me no time to work more productively yet :-/ I don't think the file is fine in the current state, though.

PR #207, opened 2018-05-24
Update lepton-cli man page

graahnul-grom:
Indicate that values passed to --size, --margins
and --align options can be separated either by colon,
semicolon or white space, not just ":".


vzh:
The issue was addressed in the commit 574e506 and the man page contains a paragraph with description of the change. Do we need to repeat the info everywhere?
Wouldn't it be better to just add a link to that paragraph? Or probably move it somewhere else?

graahnul-grom:
my bad.
sorry.

PR #209, opened 2018-05-26
Rename data subdirectory: "gEDA" => "lepton-eda"

graahnul-grom:
I have installed `lepton-eda` and `geda-gaf` with the same `$PREFIX` and so far everything is working fine. Continue testing.

### DIFF
--- a/liblepton/src/edapaths.c
+++ b/liblepton/src/edapaths.c
@@ -36,13 +36,13 @@
 #ifdef __cplusplus
 static const gchar DATA_ENV[] = "GEDADATA";
 static const gchar CONFIG_ENV[] = "GEDADATARC";
-static const gchar DATA_XDG_SUBDIR[] = "gEDA";
+static const gchar DATA_XDG_SUBDIR[] = "lepton-eda";
 static const gchar DATA_GUESS_FILE[] = "scheme/geda.scm";
 static const gchar USER_DOTDIR[] = ".gEDA";
 #else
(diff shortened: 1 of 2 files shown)


graahnul-grom:
@vzh Could we turn the compatibility symbolic links off by default now?
Is it possible to merge the changes we have just discussed on the mailing list (store dialog geometry in a separate file; [wip_cfg_priv_split](https://github.com/graahnul-grom/lepton-eda/tree/wip_cfg_priv_split) branch)?

vzh:
> @vzh Could we turn the compatibility symbolic links off by default now?

I think yes.

> Is it possible to merge the changes we have just discussed on the mailing list (store dialog geometry in a separate file; wip_cfg_priv_split branch)?

[The XDG spec](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) reads that *user-specific non-essential (cached) data* should be written to `$XDG_CACHE_HOME`. I think those data are just *user-specific* and *non-essential* (especially for users like me who use tailing WMs) :-)
Another question is how to name the file containing main window/dialogs data. I think, since the directory name already contains `lepton`, we could just use something like `window.conf` or something like this. WDYT?

P.S. If you open a new issue/PR, it'll be a better place to discuss the issue, and we could point the users who are interested in it to the discussion.

graahnul-grom:
BTW, what window manager are you using?

vzh:
dwm

PR #215, opened 2018-05-30
Rename 2 toplevel directories

vzh:
Two toplevel directories, `gnetlist` and `gattrib` have been renamed to correspond the new naming scheme used in lepton.


vzh:
Since this is just a renaming issue, I merge this in hope anything will be OK and knowing that we can revert any wrong change, anyways ;-)

PR #216, opened 2018-05-31
Update NEWS with info about new changes since 1.9.3

vzh:
For now the file NEWS.md includes the list of any changes up to the current master head except for changes in netlister (which require some more time to sort out). Any suggestions and fixes are appreciated.


vzh:
I believe it's OK to add this as is in order to allow our users/devs to read it and report any mistakes.

PR #219, opened 2018-06-07
Use cache directory to store program-specific settings

### LINKED ISSUE #214: Split user configuration
Currently, user configuration file (`geda-user.conf`) is used to store an important
data that controls behaviour of the tools in the suite. At the same time,
`lepton-schematic` use that file for program-specific settings, like dialog
geometry.  The latter is not important to the user and should be stored separately.
In the draft implementation ([wip_cfg_priv_split branch](https://github.com/graahnul-grom/lepton-eda/tree/wip_cfg_priv_split)) dialog geometry is stored in `geda-priv.conf` file in the user configuration directory. However, there is `$XDG_CACHE_HOME` directory defined in [XDG specification](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) that is designed to be used exactly for that.

graahnul-grom:
Closes #214.
Now dialog geometry data is loaded from/saved to `~/.cache/lepton-eda/gui.conf` and `lepton-schematic` does not write to `geda-user.conf`.

### DIFF
--- a/docs/specifications/config-api.txt
+++ b/docs/specifications/config-api.txt
@@ -21,6 +21,7 @@
         3.1.1  Default context
         3.1.2  System context
         3.1.3  User context
+        3.1.4  Cache context
       3.2  Loading configuration files
       3.3  Saving configuration files
       3.4  Context parents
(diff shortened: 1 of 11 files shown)


vzh:
As this series changes API, please update `config-api.txt` as well.

graahnul-grom:
@vzh Yes, and I'll push the changes to this branch.

PR #222, opened 2018-06-07
Add support for custom schematic font

graahnul-grom:
Add `[schematic.gui]::font-name` configuration key
(string, empty by default) that sets the name of
the font to be used to draw text in schematics.
For example:
```ini
[schematic.gui]
font-name=OpenGost Type A TT Regular
```
This parameter also affects the output created
with `File->Write Image...` and of course should
be distinguished from `[export]::font` configuration
key (font for `lepton-cli export` command).

### DIFF
--- a/schematic/src/gschem_toplevel.c
+++ b/schematic/src/gschem_toplevel.c
@@ -26,6 +26,9 @@ handle_undo (GschemToplevel *w_current);
 static void
 notify_options (GschemToplevel *w_current);
 
+static void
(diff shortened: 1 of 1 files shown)


vzh:
I'm not particularly happy with the fact that we'll have two different key names (`font` and `font-name`) for the same functionality in two different tools. Could we use the same key name here? WDYT?

graahnul-grom:
Good point, I totally agree. I will change the name to `[schematic.gui]::font`.

PR #223, opened 2018-06-30
Font selection widget

graahnul-grom:
Add widget that allow user to interactively select font
for rendering text in schematics. Settings for selected font 
can also be saved in either local or user configuration file.


vzh:
Cool, thanks!

PR #226, opened 2018-07-03
Enable middle mouse button to display popup menu

### LINKED ISSUE #229: No way to show the popup menu if the right mouse button is used for panning
It would be nice if we can assign free (middle) mouse button to display the popup
menu in that case, e.g. by using the following expression in `gschemrc` file:
```scheme
( middle-button "popup" )
```
But it's impossible: the popup menu can be shown using the right mouse
button only. However, it's easy to implement (#226). Testing and suggestions are welcome.

graahnul-grom:
Closes #229


vzh:
OK, let's our users check if/how this works :-)

PR #228, opened 2018-07-14
Enable freestyle colors

### LINKED ISSUE #224: Freestyle colors should not be disabled by default
Disabling that colors effectively prohibit their usage:
if users want to share schematic files, they also have to
share their color schemes, because otherwise some text
will be invisible to those with default setup.

graahnul-grom:
Closes #224


graahnul-grom:
bw

light

dark

vzh:
Please don't forget to update NEWS. Thanks!

PR #237, opened 2018-08-02
netlist: Enable filtering of nets with "no connect" symbol attached

### LINKED ISSUE #172: Single-Pin-Net in netlist for unconnected pins

graahnul-grom:
Add new function `schematic-nets-unfiltered()` which acts on the `<schematic>` record and will return all nets. Now this is used by `drc2` and `geda` backends.
Make `schematic-nets()` function return all nets except those that have `no connect` symbol (`misc/nc-*.sym`) attached to them (should be fine for the most of ordinary backends).
Filtering is done by looking for `symbol=nc` attribute in the symbol attached to a net;
that attribute is added to `misc/nc-*.sym` symbols.
Backward compatibility is maintained by retaining `device=DRC_Directive` and `value=NoConnection` attributes in `no connect` symbols in the library.

An example of how to use these functions in backends:
```scheme
( set! my-nets ( schematic-nets toplevel-schematic ) )
( set! all-my-nets ( schematic-nets-unfiltered toplevel-schematic ) )
```

Closes #172.
Previous PR (now closed) where the issue was discussed: #176.


graahnul-grom:
> If it has no symversion= attribute, I would add symversion=1.0 and then change it to 1.1 in order to reflect its internal minor changes.

Should it be done in two separate commits?

>  adding a new attribute (symbol=) should be visible somewhere in the documentation.

Where should I place the documentation for it?


As a test, I wrote a little `lepton-netlist` backend, [rmnc-dmn](https://graahnul-grom.github.io/gnet-rmnc-dmn.scm.html) ([plain scm](https://graahnul-grom.github.io/gnet-rmnc-dmn.scm)), with `get-connected-nets()` function to be placed in `gnetlist.scm`. Could you please review it? Did you mean *that* programming style?

> Please squash your changes with previous relevant commits.

I'm afraid it's beyound my basic `git` skills. :( Maybe it's better to write all from scratch...

vzh:
> > If it has no symversion= attribute, I would add symversion=1.0 and then change it to 1.1 in order to reflect its internal minor changes.

> Should it be done in two separate commits?

Yes, please.

> > adding a new attribute (symbol=) should be visible somewhere in the documentation.

> Where should I place the documentation for it?

The first place is obviously NEWS. The other one is our wiki, I think. We have to point users somewhere :-) In future, we should make up a plan of how to integrate our wiki info into our repo.

> As a test, I wrote a little lepton-netlist backend, rmnc-dmn (plain scm), with get-connected-nets() function to be placed in gnetlist.scm. Could you please review it? Did you mean that programming style?

The main thing I don't like in it is using of imperative style procedures, i.e. `set!` in this case. Things would look tidier (and work faster) if you use functional style analogs. E.g., you would use `map` instead of `for-each` and then `(apply append '((resulting) (list) (of) (lists)))` to get the result. There are other forms available, both more simple or more complex, to achieve the desired result. Otherwise, your code suits me pretty well. I don't mind include it as is, though hope you'll do it better ;-)

> > Please squash your changes with previous relevant commits.

> I'm afraid it's beyond my basic git skills. :( Maybe it's better to write all from scratch...

Well, there are tools like `magit` and `stgit` that facilitate this work. Though mere `rebase -i ...` should work OK. Please look at the *Interactive mode* section in git-rebase(1), and search for the word `squash` there.

Issue #238, opened 2018-08-03
The log window should be scrolled to the bottom when something is written to the log


graahnul-grom:
For example: I run `DRC`, then open the log window to check the result (the last two lines). But what I see instead? The middle part of the previous `DRC` run:

![log_no_scroll](https://user-images.githubusercontent.com/26083750/43645153-c5974fe8-9739-11e8-98ed-90ff1a757e27.png)

The issue [was reported](https://bugs.launchpad.net/geda/+bug/1493456) (and confirmed) on launchpad back in 2015.

PR #239, opened 2018-08-03
Log window: fix scrolling

graahnul-grom:
Make the log window scroll to the bottom
when something is written to the log.

Issue #238


vzh:
Thanks!

Issue #250, opened 2018-09-07
Connected graphicals in generated netlist


graahnul-grom:
`lepton-netlist` doesn't ignore `graphical` components if they are connected to nets. It happens with some backends (not all), notably, with `PCB` and `drc2`.

Test schematic (attached):
![netlist_graphicals](https://user-images.githubusercontent.com/26083750/45199369-ed7a5280-b273-11e8-9062-e799bd179413.png)

`C1` should not appear in resulting netlist, but it is:

`$ lepton-netlist -o - -g PCB netlist_graphicals.sch`
GND R1-1 R3-2
netx C1-1 R1-2 R2-1
nety C1-2 R2-2 R3-1

`lepton-netlist -o - -g drc2 netlist_graphicals.sch`
Checking non-numbered parts...
Checking duplicated references...
Checking NoConnection nets for connections...
Checking nets with only one connection...
Checking pins without the 'pintype' attribute...
NOTE: Found pins without the 'pintype' attribute: C1:1 C1:2
Checking type of pins connected to a net...
Checking unconnected pins...
Checking slots...
Checking duplicated slots...
Checking unused slots...

If you disconnect `C1` from `netx` and `nety` nets, `lepton-netlist` will generate a correct netlist.

[netlist_graphicals.zip](https://github.com/lepton-eda/lepton-eda/files/2359585/netlist_graphicals.zip)

PR #252, opened 2018-09-09
lepton-netlist: Add a test for graphical= attribute.

vzh:
The lepton issue #250 revealed that there was no test in the
lepton-netlist test suite for filtering components having the
"graphical=" attribute attached. This patch adds the test based on
the schematic by @graahnul-grom provided in the issue.


graahnul-grom:
Alas, I forgot to add `value` attribute to components... Output files would look better with it.

vzh:
@graahnul-grom, just add a PR, if you want to make it better :-)

PR #257, opened 2018-10-06
Fix lepton-schematic -q command line option

graahnul-grom:
-q (--quiet) command line option does not take an argument.


vzh:
Applied, thanks!

PR #268, opened 2018-10-31
Improve diagnostics of input file errors

graahnul-grom:
- make `lepton-schematic` print input file line by line when invoked with `-v` (`--verbose`) flag
- better `"Failed to load file"` error message (include line number where the parser was stopped)

tb_083_parse_verb_screenshot


vzh:
@graahnul-grom, thanks! Please update NEWS as well.

Issue #225, opened 2018-07-03
Reorganize the main menu


graahnul-grom:
A. `File` menu:
  1. delete duplicated `Revert` (already in `Page` menu)

B. `Edit` menu:
  1. combine to one item: `Object Properties...`:
    - `Color...`
    - `Line Width & Type...`
    - `Fill Type...`
    - `Pin Type...`

C. `View` menu:
  1. rename: `Sidebar` => `Right Dock`
  2. rename: `Status` => `Bottom Dock`

D. `Page` menu:
  1. separate `Revert` from the other items
  2. rename: `Revert` => `Revert...`

E. `Hierarchy` menu:
  1. delete duplicated `Documentation...` (already in `Help` menu)

F. `Options` menu:
  1. delete obsolete `Text Size...` item
  2. rename/rearrange, for example:
    - `Snap Grid Spacing...` => `Options...`
    - [separator]
    - `Scale up Grid Spacing` => `Grid +`
    - `Scale down Grid Spacing` => `Grid -`
    - [separator]
	- `Cycle grid styles` => `Grid style: cycle Dots/Mesh/Off`
    - `Toggle Snap On/Off` => `Grid snap: cycle Grid/Resnap/Off`
    - [separator]
    - `Toggle Outline/Box` => `Feedback mode: Outline/Box`
    - `Toggle Net Rubberband` => `Net: Rubberband On/Off`
    - `Toggle Magnetic Net` => `Net: Magnetic On/Off`
    - [separator]
    - `Show Coords Window...` => `Coords Window`
    - `Show Log Window...` => `Log Window`

Comments:
- B.1: all that commands do exactly the same: show `Object` dialog
box (`x_widgets_show_object_properties()`).
- C.3: currently, there is no way to show `find text state` window
unless docking windows GUI is enabled. We need to add a new action,
e.g. `view-find-text-state`. **[done]**
- F.2: `Options` menu:

![2018_07_03 06 21 34](https://user-images.githubusercontent.com/26083750/42197153-25557b3a-7e8a-11e8-939b-71daaa97f8f1.png)

![2018_07_03 06 20 37](https://user-images.githubusercontent.com/26083750/42197161-2bfe1352-7e8a-11e8-8159-f13f9909e9d6.png)

Any suggestions are welcome.

graahnul-grom:
If anyone wants to try the menu modified as proposed, here's the diff attached.

[...]

PR #269, opened 2018-10-31
Add edit-object-properties action

graahnul-grom:
This action replaces  `edit-color`, `edit-linetype`, `edit-filltype` and `edit-pin-type`.
Discussed in #225.


graahnul-grom:
> Should we add backward compatibility funcs for people with custom menus

I think we can add it if there will be demand to do so.

> And please update NEWS with your changes.

OK

graahnul-grom:
But it's simple, so why not? I'll file a PR.

graahnul-grom:
#271

PR #275, opened 2018-11-09
Update vams example schematic

vzh:


graahnul-grom:
Shouldn't [gschemrc](https://github.com/lepton-eda/lepton-eda/blob/master/netlist/examples/vams/gschemrc) be updated?

`(load "../../../schematic/scheme/generate_netlist.scm")`
instead of
`(load "../../../gschem/scheme/generate_netlist.scm")`

graahnul-grom:
In [schematic/scheme/generate_netlist.scm](https://github.com/lepton-eda/lepton-eda/blob/master/schematic/scheme/generate_netlist.scm) netlister is invoked by `gnetlist` command (2 places). Should be `lepton-netlist`.

graahnul-grom:
That `generate_netlist.scm` file is used in `vams` example only.
Should be moved to `netlist/examples/vams/` (?)

PR #280, opened 2018-11-19
Remove obsolete options-text-size action

graahnul-grom:
This obsolete action was used to invoke
"Text Size" dialog, which was removed
in revision 150c850 (back in 2014).


vzh:
`git grep` tells me that there is one more line which should be removed:
> schematic/include/prototype.h:141:SCM g_keys_options_text_size(SCM rest);

PR #282, opened 2018-11-20
Refactor main() function in lepton-netlist

graahnul-grom:
This is done in preparation to converting netlist backends to modules
and placing them in separate sub-directory.

P.S. @vzh Believe it or not, but my hands just fail me to write using patented 
Scheme's `Dog's Breakfast Coding Style (tm)`. :))

### DIFF
--- a/netlist/scheme/netlist.scm
+++ b/netlist/scheme/netlist.scm
@@ -973,7 +973,15 @@ Lepton EDA homepage: <https://github.com/lepton-eda/lepton-eda>
 
   ; Load Scheme FILE before loading backend (-l FILE):
   ;
-  ( for-each primitive-load (gnetlist-option-ref 'pre-load) )
+  ( catch #t
+    ( lambda()
+      ( for-each primitive-load (gnetlist-option-ref 'pre-load) )
+    )
+    ( lambda( tag . args )
+      ( catch-handler tag args )
+      ( netlist-error 1 (_ "Failed to load Scheme file before loading backend.\n") )
+    )
+  )
 
   ; Load backend:
   ;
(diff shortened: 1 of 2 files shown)


vzh:
@graahnul-grom, after about 15 years of working with *vim* and about 3 years working with *emacs*, I should admit that *emacs* (with *paredit*, *scheme-mode* and such) is the better Lisp editor (and, combined with *geiser*, the better guile debugger) in the world. Well, now the style you named *dog's breakfast* one seems to me much tidier than other alternatives (and I don't have to count parens) ;-) (well, again, if I understood your irony properly...)
As for the patch-set, I like some ideas in your changes. However, I myself would avoid imperative procedures like `set!`. Hope to look at the code a bit closer this week.

graahnul-grom:
> @graahnul-grom, after about 15 years of working with _vim_ and about 3 years working with _emacs_, I should admit that _emacs_ (with _paredit_, _scheme-mode_ and such) is the better Lisp editor (and, combined with _geiser_, the better guile debugger) in the world. Well, now the style you named _dog's breakfast_ one seems to me much tidier than other alternatives (and I don't have to count parens) ;-) (well, again, if I understood your irony properly...)

Just kidding; never mind. `DBCS(tm)` brand sounds just great, I'm going to design a logo for it. :-)
Nevertheless, if a style of coding starts to look good only after years of experience with particular editor and particular set of tools, I suspect, there's something wrong with it. Then, it's hard to attract new developers, I think. And it may even provoke someone uninitiated to act somewhat inadequately... (just remember such a natural event as Roland with his stuff :-) ).


> As for the patch-set, I like some ideas in your changes. However, I myself would avoid imperative procedures like `set!`. Hope to look at the code a bit closer this week.

I think, there's nothing wrong with `set!`, it's a basic operation in any programming language. And if it makes the code easier to understand, you should use it, of course.
(BTW, programming is imperative by definition: you told the computer what it have to do :) )

vzh:
You're provoking me to answer a wrong question. Experience is a matter of work, and outlooks change with time. And, while I disagree with you here, I won't start another flame on this topic ;-) Just a couple of notes... Mathematicians in high schools use some language (actually, several languages) which is (are) not convenient for novices, but is (are) convenient for themselves. And even C has many helper tools evolved for several decades, that make the life of a programmer a bit easier. As for special tools, well, some tools, e.g. an oscilloscope, are better when they're used with special helper tools, like probes, though a hammer doesn't need them and is more intelligible ;-) And it's natural, I believe, that when you eventually find the right tool for your work, you use it instead of all other things you used before. And the last but not least, computers are not calculators and programs are not just sets of instructions like great poems are not just sets of letters. So I consider your last *btw* to be wrong, too ;-) SICP's "Foreword" could probably explain all this better than me...

graahnul-grom:
Sorry if I've said something wrong. Actually, I'm talking just about a code style, nothing more. Anyway, I'm convinced that things like these should better be discussed over a beer. :-) Let's continue our work!

PR #294, opened 2018-12-07
Remove config keys through liblepton configuration API

graahnul-grom:
Extend `liblepton` configuration API by adding `eda_config_remove_key()`
function (`config-remove-key!` in Scheme).
I'd also add a function that deletes config groups, too, but it's not clear
what signal should be raised in that case. There's `config-changed` signal
with two parameters (`group`, `key`). If we reuse it for a "group removed"
event, what values those parameters should take? Or do we have to introduce
a new signal?

### DIFF
--- a/docs/scheme-api/lepton-scheme.texi
+++ b/docs/scheme-api/lepton-scheme.texi
@@ -1870,6 +1870,12 @@ in the specified @var{group}.
 Since 1.10.
 @end defun
 
+@defun config-remove-key! cfg group key
+Removes the configuration parameter identified by the given @var{group}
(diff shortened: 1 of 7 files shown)


graahnul-grom:
> what values those parameters should take?

I propose to emit `config-changed` signal on group deletion,
passing group name in `group` parameter, and an empty
string in `key`. Document it in scheme-api manual.
@vzh If there are no objections, I'll add `eda_config_remove_group()`
and `config-remove-group!` functions.

vzh:
@graahnul-grom, please go for it. I think it's a good idea.

PR #301, opened 2018-12-19
Remove config groups through liblepton configuration API

graahnul-grom:
Extend `liblepton` configuration API by adding `eda_config_remove_group()`
function (`config-remove-group!` in Scheme).
The `group` parameter of the `config-changed` signal is set to the group name
deleted and the `key` parameter is set to an empty string.

Related to PR #294 where `eda_config_remove_key()` and `config-remove-key!`
functions were added.

### DIFF
--- a/docs/scheme-api/.gitignore
+++ b/docs/scheme-api/.gitignore
@@ -2,3 +2,4 @@
 lepton-scheme.info
 stamp-vti
 version.texi
+lepton-scheme.html/*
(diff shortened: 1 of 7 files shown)


docs/scheme-api/.gitignore
```diff
@@ -2,3 +2,4 @@
 lepton-scheme.info
 stamp-vti
 version.texi
+lepton-scheme.html/*
```
vzh: This is superfluous here. (BTW, what command should be issued to make the directory, `make html`?). Please put it into a separate commit with appropriate description.
Anything other suits me well, thank you!

vzh: Ah, OK, just checked it.

graahnul-grom: I use these commands in my install script:

`cd docs/scheme-api/`
`make html`
`make install-html`

PR #303, opened 2018-12-20
Add open-page-hook to the lepton-schematic Scheme API

graahnul-grom:
The `open-page-hook` hook is called when an existing schematic page is loaded.


vzh:
:+1:

PR #304, opened 2018-12-20
Add netlist export menu in `lepton-schematic` (for now, for the allegro backend)

vzh:
This is the first step to make all netlister functionality available in `lepton-schematic`.

### DIFF
--- a/netlist/scheme/gnet-allegro.scm
+++ b/netlist/scheme/gnet-allegro.scm
@@ -118,7 +118,7 @@ PINCOUNT ~A
             (connections->string (get-connections netname schematic))))
   (map net->string netnames))
 
-(define (allegro-netlist schematic output-filename)
+(define* (allegro* schematic #:key output-filename)
   (let ((use-stdout? (not output-filename))
         (packages (schematic-packages schematic))
         (nets (schematic-nets schematic)))
@@ -131,5 +131,5 @@ PINCOUNT ~A
     (display "$END\n")
     (allegro:write-device-files packages '() use-stdout?)))
 
-(define (allegro output-filename)
-  (allegro-netlist (toplevel-schematic) output-filename))
+(define (allegro filename)
+  (allegro* (toplevel-schematic) #:output-filename filename))
--- a/netlist/scheme/gnet-allegro.scm
+++ b/netlist/scheme/gnet-allegro.scm
@@ -22,20 +22,22 @@
 ;; Allegro netlist format
 (use-modules (netlist schematic)
              (netlist error)
-             (netlist schematic toplevel))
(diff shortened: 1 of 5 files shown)


vzh:
Are there any interest in such (or the like) functionality at all? I mean netlist export menus?

graahnul-grom:
@vzh Yes, I think it would be very useful to access netlister functionality from the schematic editor. `Export` menu is a good idea (though I'd rather make it optional), but I mean "netlister functionality" in the wider sense: to have access to the results. For example, read `DRC` error data (in some programmatic form, not a text output) in Scheme scripts.

vzh:
@graahnul-grom Thank you for the reply.
As for getting DRC data in GUI, I have an idea to use `blame*` functionality the same way as is used for symbol checking in GUI. That is, a netlister backend adds weak refs to schematic objects using `make-object-property`, and then the info is available until the objects exist. Though my first goal is to adapt all backends to make them available in the export menu.

vzh:
> Speaking of optional menu items: could we split the `system-gschemrc` now?

I don't mind. Please go for it.

noqsi:
> On Jan 9, 2019, at 4:04 AM, Vladimir Zhbanov <notifications@github.com> wrote:
> 
> Though my first goal is to adapt all backends to make them available in the export menu.


How will this work for user-contributed back ends?

John Doty              Noqsi Aerospace, Ltd.
jpd@noqsi.com

vzh:
@noqsi, The users can adopt the approach I used in my last commit in this branch (it is not perfect, I know, any better solutions and suggestions are welcome). Basically, the user has to define what he thinks schematic consists of, the function which should make the netlist for it (I factored out the function `allegro*` in the allegro backend for this), the function that will be invoked from corresponding menu entry (the latter two functions could be combined), and the menu entry itself. The functions can be defined in the `gschemrc` for now. Nothing of this is needed for `lepton-netlist`, of course, it'll work as always.

graahnul-grom:
Now `(schematic netlist)` module (and hence `netlist`) is loaded from
`system-gschemrc`, and every time `lepton-schematic` starts,
`Failed to load config from...` messages are printed to the log.
It would be nice to have an ability to (programmatically) disable this
(i.e. have an access to the `netlist` options, - set by `-w` in this case).

graahnul-grom:
@vzh  I would also suggest that `Netlist` menu actions have some form of feedback, be it a message box or log/stdout report, something like
`netlist: running [allegro] backend...`
`...done. output is written to /path/to/sch/dir/allegro.out`
WDYT?

vzh:
I think, it would be better to request the output file name first time, and thereafter do as you suggested.

PR #310, opened 2018-12-25
liblepton Scheme API functions to lock/unlock objects

graahnul-grom:
`liblepton`:  Add two new functions to the `(geda object)` module
to work with object's `locked` status: `object-locked?` and `set-object-locked!`.
They read and set the `selectable` flag in the `OBJECT` structure.
Now it's possible to programmatically lock/unlock arbitrary objects on a page.

For example, a feature requested in this launchpad [bug report](https://bugs.launchpad.net/geda/+bug/698790) can easily be implemented using these new functions (e.g. in conjunction with the `open-page-hook`).

### DIFF
--- a/liblepton/scheme/geda/object.scm
+++ b/liblepton/scheme/geda/object.scm
@@ -1,6 +1,8 @@
-;; gEDA - GPL Electronic Design Automation
-;; libgeda - gEDA's library - Scheme API
+;; Lepton EDA
+;; liblepton - Lepton's library - Scheme API
 ;; Copyright (C) 2010-2011 Peter Brett <peter@peter-b.co.uk>
+;; Copyright (C) 2012-2016 gEDA Contributors
+;; Copyright (C) 2017-2018 Lepton EDA Contributors
 ;;
 ;; This program is free software; you can redistribute it and/or modify
 ;; it under the terms of the GNU General Public License as published by
@@ -466,3 +468,9 @@
    (lambda (obj) (%mirror-object! obj x))
    objects)
   objects)
+
+
+
+( define-public object-locked?     %object-locked? )
+( define-public set-object-locked! %set-object-locked! )
+
--- a/liblepton/scheme/geda/object.scm
+++ b/liblepton/scheme/geda/object.scm
@@ -2,7 +2,7 @@
 ;; liblepton - Lepton's library - Scheme API
 ;; Copyright (C) 2010-2011 Peter Brett <peter@peter-b.co.uk>
 ;; Copyright (C) 2012-2016 gEDA Contributors
-;; Copyright (C) 2017-2018 Lepton EDA Contributors
+;; Copyright (C) 2017-2019 Lepton EDA Contributors
 ;;
 ;; This program is free software; you can redistribute it and/or modify
 ;; it under the terms of the GNU General Public License as published by
@@ -469,8 +469,5 @@
    objects)
   objects)
 
-
-
 ( define-public object-locked?     %object-locked? )
-( define-public set-object-locked! %set-object-locked! )
-
+( define-public set-object-selectable! %set-object-selectable! )
--- a/liblepton/scheme/geda/object.scm
+++ b/liblepton/scheme/geda/object.scm
@@ -469,5 +469,5 @@
    objects)
   objects)
 
-( define-public object-locked?     %object-locked? )
+( define-public object-selectable?     %object-selectable? )
 ( define-public set-object-selectable! %set-object-selectable! )
--- a/liblepton/src/scheme_object.c
+++ b/liblepton/src/scheme_object.c
@@ -744,7 +744,7 @@ SCM_DEFINE (object_locked_p, "%object-locked?", 1, 0, 0,
 
   OBJECT* obj = edascm_to_object (obj_s);
 
-  return obj->selectable ? SCM_BOOL_F : SCM_BOOL_T;
+  return scm_from_bool (obj->selectable == 0);
 
 } /* object_locked_x() */
 
@@ -773,7 +773,7 @@ SCM_DEFINE (set_object_locked_x, "%set-object-locked!", 2, 0, 0,
 
   OBJECT* obj = edascm_to_object (obj_s);
 
-  int locked = locked_s == SCM_BOOL_T;
+  int locked = scm_is_true (locked_s);
 
   /* do not allow locked objects to be selected:
   */
--- a/liblepton/src/scheme_object.c
+++ b/liblepton/src/scheme_object.c
@@ -1,6 +1,8 @@
-/* gEDA - GPL Electronic Design Automation
- * libgeda - gEDA's library - Scheme API
+/* Lepton EDA
+ * liblepton - Lepton's library - Scheme API
  * Copyright (C) 2010-2012 Peter Brett <peter@peter-b.co.uk>
+ * Copyright (C) 2011-2016 gEDA Contributors
+ * Copyright (C) 2017-2018 Lepton EDA Contributors
  *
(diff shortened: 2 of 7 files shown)


vzh:
Why not just `lock-object!`?

graahnul-grom:
@vzh I thought about giving this function a more simple name, too.
Then, looking through the Scheme API documentation, I decided to
follow the existing pattern (like in almost all object-related function, e.g. 
 `object-color` / `set-object-color!` )
Of course I can change the name, if you think it's worth it.

liblepton/scheme/geda/object.scm
```diff
@@ -466,3 +468,9 @@
    (lambda (obj) (%mirror-object! obj x))
    objects)
   objects)
+
+
+
+( define-public object-locked?     %object-locked? )
+( define-public set-object-locked! %set-object-locked! )
```
vzh: With regard to the `set-object-locked!` function, I would probably
rename it to `set-object-selectable!`, and add
`object-selectable?` instead of `object-locked?` (or together with
it).

liblepton/scheme/geda/object.scm
```diff
@@ -466,3 +468,9 @@
...
+
```
vzh: Please avoid trailing whitespaces/newlines.

liblepton/scheme/geda/object.scm
vzh: ... and change the logic of the C function accordingly.

liblepton/src/scheme_object.c
```diff
@@ -719,6 +721,80 @@ SCM_DEFINE (set_object_color_x, "%set-object-color!", 2, 0, 0,
...
+ * object is considered to be unlocked, otherwise it is locked.
+ *
+ * \note Scheme API: Implements the %object-locked? procedure in the
+ * (geda core object) module.
+ *
+ * \param obj_s  #OBJECT smob to inspect.
+ *
+ * \return       Boolean value indicating whether \a obj_s is locked.
+ */
+SCM_DEFINE (object_locked_x, "%object-locked?", 1, 0, 0,
```
vzh: There is a convention about C function naming if their Scheme
counterparts contain special symbols like `!` or `?` in their
names. `!` transforms into `_x` (eXclamation mark), and `?`
transforms into `_p` (`p` just looks a bit like the question
mark). Please fix this in your code.

liblepton/src/scheme_object.c
```diff
@@ -719,6 +721,80 @@ SCM_DEFINE (set_object_color_x, "%set-object-color!", 2, 0, 0,
...
+            (SCM obj_s), "Check whether an object is locked.")
+{
+  SCM_ASSERT (EDASCM_OBJECTP (obj_s), obj_s,
+              SCM_ARG1, s_object_locked_x);
+
+  OBJECT* obj = edascm_to_object (obj_s);
+
+  return obj->selectable ? SCM_BOOL_F : SCM_BOOL_T;
```
vzh: You could use `scm_from_bool()` instead of `if` shortcut here.

liblepton/src/scheme_object.c
```diff
@@ -719,6 +721,80 @@ SCM_DEFINE (set_object_color_x, "%set-object-color!", 2, 0, 0,
...
+            (SCM obj_s, SCM locked_s), "Lock or unlock an object.")
+{
+  SCM_ASSERT (EDASCM_OBJECTP (obj_s), obj_s,
+              SCM_ARG1, s_set_object_locked_x);
+  SCM_ASSERT (scm_is_bool (locked_s), locked_s,
+              SCM_ARG2, s_set_object_locked_x);
+
+  OBJECT* obj = edascm_to_object (obj_s);
+
+  int locked = locked_s == SCM_BOOL_T;
```
vzh: Just one note. In Scheme boolean world, any non-`#f` value of a variable is `true`. Your expression `int locked = locked_s == SCM_BOOL_T;` makes this... ehm... a bit contrariwise, everything non-`#t` gets `false`, which is wrong, I believe. You would better use the `scm_is_true()` Scheme function in this case.

vzh:
I have other questions as well:
- IIUC, our current file format doesn't support locking for all objects (just for components). Do you plan to extend it? If so, how?
- One related question: how would you use your new functions in the schematic capture program, what plugins/extensions/features could become available for you personally and for the users. What file format changes would they involve (if any).
- There are already stable API functions working with component selectable status. Do you intend to rewrite them?

P.S. (a minor issue) Your non-DBS breaks pretty keyword highlighting in Emacs due to using of spaces just after/before parens. Though it is an issue with Emacs plugin writers who adhere to their DBS, I believe ;-)

graahnul-grom:
@vzh С наступившим Новым годом, товарищ командир! Пожелаем Вам доброго здоровья, и всем нам - творческих успехов и великих свершений!

Thank you for your review.
Frankly, I didn't even think about the file format. The idea was very simple: attach to objects some attribute to indicate that they must be locked, then check it in `open-page-hook` and use `set-object-locked!`. (We can attach attributes to boxes, lines, paths, pictures etc.)

Here's the script to include in `gschemrc` (it checks for `lock=yes` in objects) and a sample schematic:
lock

[tb_105_lock_obj.scm.txt](https://github.com/lepton-eda/lepton-eda/files/2739582/tb_105_lock_obj.scm.txt)

[lock.sch.txt](https://github.com/lepton-eda/lepton-eda/files/2739584/lock.sch.txt)

graahnul-grom:
> There are already stable API functions working with component selectable status. Do you intend to rewrite them?

The only issue with components I see so far is this: when an object is locked in schematic editor, all of its attributes are also locked; however, when schematic is loaded next time, a component remains locked, but attributes are not.

graahnul-grom:
@vzh I fully agree with your review comments. It seems that it's better to rewrite all the stuff in another PR while closing this one. WDYT?

liblepton/src/scheme_object.c
graahnul-grom: 8-\ Didn't know about that. Absolutely counter-intuitive. And just against the common sense, I think. :)

vzh:
@graahnul-grom, Спасибо, Дмитрий, за поздравления, ответно желаю всего доброго! Я своим коллегам по работе и по FOSS-проектам (ровесникам или младшим меня) всегда предлагаю переходить на "ты". Если Вы не против, предлагаю то же и Вам.

As concerning attribs to change object properties, we could use it some time (at least until the next stable release). Though starting the work on a newer, extended file format, in the longer-term perspective, would be a better option. For now, we could use attribs in this case as well as for modifying nets, circles, paths and so on, since there were several requests about changing their view before.

The issue with component attribs seems to be a bug. Please feel free to add an issue about it (is there one on LP?).

You don't have to close the issue and add a new one, if you don't want it (if you want, please do). You have an option of doing `git push --force` to modify your current branch as a whole.

graahnul-grom:
@vzh Идёт! Тогда я - Дима (в смысле, не Дмитрий), хорошо? :-)

I read some horror stories (on stackexchange, IIRC) about using `push --force` on `github`: comments are lost, reviews are turned into a mess, etc.

liblepton/src/scheme_object.c
vzh: As in C, any non-zero value is true. What's wrong? :-)

vzh:
Дим, `git push --force-with-lease ...` is much safer. Please read about the difference with plain `git push --force ...`. And really I use this command in such cases, I had no issues with it yet (more accurately, I use `magit` git gui and shortcuts in *emacs*, which is much more convenient).

vzh:
Nice, thank you!
A minor note: please fix function names in test code (`set-object-locked!`).

PR #315, opened 2019-01-11
Lock/Unlock operations - fixes and improvements

### LINKED ISSUE #314: Locked component's attributes
1. When a component object is locked in `lepton-schematic`, all of its attributes are also locked; however, when the schematic file is loaded next time, a component remains locked, but its attributes are not.

`=>` They should also be locked.

2. If a component and its attributes are locked, to fully unlock it one have to select (by the bounding box) not only a component itself, but also all visible attributes. Otherwise a component gets unlocked, but attributes (that are not selected during unlocking) remains locked.

`=>` When unlocking a component, all its attributes should also be unlocked without having to select them.

graahnul-grom:
Closes #314

### DIFF
--- a/liblepton/src/o_attrib.c
+++ b/liblepton/src/o_attrib.c
@@ -124,6 +124,10 @@ void o_attrib_attach (TOPLEVEL *toplevel, OBJECT *attrib, OBJECT *object,
     return;
   }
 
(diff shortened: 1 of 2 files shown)


vzh:
:+1:

PR #317, opened 2019-01-12
Fix warnings due to uninitialised "Last-Translator" field in po-files.

vzh:
I've carefully checked all the files involved, and found they all were imported from Launchpad by Peter Brett or Peter Clifton (6-9 years ago). New name for the "Last-Translator" field, "Launchpad Translators", has been used in all cases to avoid compilation warnings about the non-initialized field.


vzh:
@graahnul-grom Thanks for the review. Rebasing and merging worked without any problem.

PR #318, opened 2019-01-13
Path object improvements

vzh:
Feature request by Karl Hammar:

> Possible not ASAP, but, for paths (^H), to be able to have
> round join styles instead of default miter style. See e.g.
> http://www.delorie.com/archives/browse.cgi?p=geda-user/2018/07/12/13:25:11

> And in the same vein, for filled paths with line width = 0, to really
> have zero width, not some default width which will enlargen the
> triangle (mostly arrow heads, but filled polygons in general).

I decided to not modify the schematic/symbol file format and do
the latter change in the simple way. Now, when line width of a
path is set to zero, the stroke is not rendered, thus allowing
drawing of filled figures without increasing their size by a half
of the line width, so the object sizes strongly correspond to what
the user wanted.

Attached is a
[test symbol](https://github.com/lepton-eda/lepton-eda/files/2756153/test.sym.txt)
by Karl Hammar.

### DIFF
--- a/libleptonrenderer/edacairo.c
+++ b/libleptonrenderer/edacairo.c
@@ -363,6 +363,7 @@ eda_cairo_stroke (cairo_t *cr, int flags, int line_type, int line_end,
   double dummy = 0;
   cairo_line_cap_t cap;
   cairo_line_cap_t round_cap_if_legible = CAIRO_LINE_CAP_ROUND;
+  cairo_line_join_t join = CAIRO_LINE_JOIN_MITER;
   int num_dashes;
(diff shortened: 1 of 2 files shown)


graahnul-grom:
An interesting (and very good) side effect: with this patch, different `cap styles` can be selected for path objects (with width != 0). It used to be always `square`.

vzh:
It's not a side effect, but the first one of the requested features. OK, since it works for you, and there is no feedback yet from the original reporter, I'll merge the branch.

PR #319, opened 2019-01-14
Let the user set `lepton-netlist` options

vzh:
Request by @graahnul-grom in #304:

> Now (schematic netlist) module (and hence netlist) is loaded from
> system-gschemrc, and every time lepton-schematic starts,
> Failed to load config from... messages are printed to the log.
> It would be nice to have an ability to (programmatically) disable this
> (i.e. have an access to the netlist options, - set by -w in this case).

The code here allows the user setting of arbitrary netlister options in order to change the behaviour of its modules if they're used by other tools (e.g., in `lepton-schematic` GUI). For example, the following lines in `gafrc` prevents warnings on missing xdg system conf files:
```
(use-modules (netlist option))
(set-netlist-option! 'no-warn-cfg #t)
```

### DIFF
--- a/netlist/scheme/lepton-netlist.in
+++ b/netlist/scheme/lepton-netlist.in
@@ -27,12 +27,12 @@ exec @GUILE@ -s "$0" "$@"
 ;;; extension. Let's make it quiet here.
 (define with-toplevel (@@ (geda core toplevel) %with-toplevel))
 (define make-toplevel (@@ (geda core toplevel) %make-toplevel))
-(define gnetlist-option-ref* (@@ (netlist option) gnetlist-option-ref))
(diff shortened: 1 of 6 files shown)


graahnul-grom:
@vzh Very nice, thank you!
Shouldn't we turn those warnings off by default (in `system-gafrc`)?

vzh:
> Shouldn't we turn those warnings off by default (in `system-gafrc`)?

I don't think so. A better option would be to make up a default system config file. Even empty, but better, with default options included. System administrators (and distrib package managers) should decide what suits better for their users. For me, for example, those warnings are good reminders that our config system transition is incomplete, so I don't switch them off :-)

PR #322, opened 2019-01-17
liblepton: Fix two compilation warnings.

vzh:
Fixes warnings by gcc as follows:

  warning: format ‘%lu’ expects argument of type ‘long unsigned
  int’, but argument 3 has type ‘gsize {aka unsigned int}’
  [-Wformat=]

### DIFF
--- a/liblepton/src/a_basic.c
+++ b/liblepton/src/a_basic.c
@@ -364,7 +364,7 @@ GList *o_read_buffer (TOPLEVEL *toplevel, GList *object_list,
 error:
   geda_object_list_delete (toplevel, new_object_list);
 
-  gsize linenum = s_textbuffer_linenum (tb);
+  unsigned long linenum = s_textbuffer_linenum (tb);
   g_prefix_error (err, "Parsing stopped at line %lu:\n", linenum);
 
   return NULL;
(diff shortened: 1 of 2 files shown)


liblepton/src/a_basic.c
```diff
@@ -365,7 +365,7 @@ GList *o_read_buffer (TOPLEVEL *toplevel, GList *object_list,
   geda_object_list_delete (toplevel, new_object_list);
 
   gsize linenum = s_textbuffer_linenum (tb);
-  g_prefix_error (err, "Parsing stopped at line %lu:\n", linenum);
+  g_prefix_error (err, "Parsing stopped at line %u:\n", linenum);
```
graahnul-grom: Anyway, we could avoid any warnings with
```
unsigned long val = linenum;
g_prefix_error (err, "Parsing stopped at line %lu:\n", val);
```

vzh:
I used `g++` on `i686`. Cannot get those warnings on amd64.

graahnul-grom:
Yes, on `amd64` the same warning issued with `%u` format flag:
"warning: format specifies type 'unsigned int' but the argument has type 'gsize' (aka 'unsigned long') [-Wformat]".
I think it's better to avoid using `glib` types like this.
In this particular case the above solution with intermediate variable can be used to prevent warnings.

graahnul-grom:
@vzh Could you please try the following code on `i686`? Does it generate warnings?
```c
size_t val = 1234;
printf( "%zu", val );
```

graahnul-grom:
@vzh Yes, it is.

PR #325, opened 2019-01-25
Amend middle mouse button popup menu code

graahnul-grom:
The code was added in #226 ("Enable middle mouse button to display popup menu").


vzh:
I've compiled the code and it works as expected. Thank you!

PR #326, opened 2019-01-28
Guile 2.2 support

vzh:
This patch-set adds support for build with guile-2.2.


vzh:
I'll apply these patches since they work at least on three systems :-)

PR #328, opened 2019-01-29
schematic: Fix segfault on update menus sensitivity.

vzh:
This segfault was triggered by "C-x" "C-v" on a component, which
caused update of menus and their sensitivity. Since several menu
items have been changed recently, one of them triggered the crash
due to improper menu entry name used.

P.S. I checked other recent menu changes, and other menu items seem to have not been affected by the issue.


vzh:
> Is this documented somewhere?

[Here](https://developer.gnome.org/gobject/stable/gobject-The-Base-Object-Type.html#g-object-get-data). It uses data set by `g_object_set_data()` above in the same file, IIUC.

vzh:
The crash I spoke about appeared sporadically before, not always. However, I triggered the issue when I edited a middle size schematic saved in a special git repo, so I could easily reproduce the crash and investigate it in gdb. I opened the schematic, selected two components, C-x, C-v, and gdb showed segfault after about 4 g_* funcs (g_log and such) calls started in `x_menus_sensitivity`. After the fix, it worked as expected.

PR #329, opened 2019-01-30
Documentation and program information updates

graahnul-grom:
about

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@ Lepton Electronic Design Automation
 ===================================
 
 Copyright (C) 1998-2017 gEDA Developers.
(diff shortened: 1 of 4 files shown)


vzh:
The assertion in the last commit that ChangeLog files are not generated any more, is not quite right. You can find `ChangeLog` targets in several `Makefile.am`s. They're generated when you do `make dist` and lepton tarballs contain them (I've just checked 1.9.7), so that the people who build the programs from tarballs (including building for various distributions) could still access the history.

graahnul-grom:
@vzh Indeed, sorry. I've fixed that.

vzh:
Merged, thank you!

PR #331, opened 2019-02-02
Fix lepton-symcheck startup script

graahnul-grom:
Unless this script is arranged like the
`lepton-netlist.in`, *precompilation* produce
object code that fails at runtime,
complaining about missing modules.


vzh:
Thanks!

PR #333, opened 2019-02-02
Do not guess system documentation directory

### LINKED ISSUE #285: lepton-schematic documentation search path
Choosing one of the documentation items
from the `Help` menu triggers the Scheme code in [gschemdoc.scm](https://github.com/lepton-eda/lepton-eda/blob/master/schematic/scheme/gschem/gschemdoc.scm).

That code tries to guess where lepton-eda docs are installed
and sometimes it fails to do that. For example, on Ubuntu
18.04 I got this error:
```
Backtrace:
In ice-9/boot-9.scm:
 160: 11 [catch #t #<catch-closure 55b428674640> ...]
In unknown file:
   ?: 10 [apply-smob/1 #<catch-closure 55b428674640>]
In ice-9/boot-9.scm:
 160: 9 [catch #t #<catch-closure 55b428c7df20> ...]
In unknown file:
   ?: 8 [apply-smob/1 #<catch-closure 55b428c7df20>]
In gschem/action.scm:
  67: 7 [eval-action! &help-guide]
In gschem/gschemdoc.scm:
  95: 6 [show-wiki "geda:gschem_ug"]
  61: 5 [sys-doc-dir]
In srfi/srfi-1.scm:
 806: 4 [any1 #<procedure 55b428b0d9e0 at gschem/gschemdoc.scm:50:5 (dir)> #]
In gschem/gschemdoc.scm:
  51: 3 [#<procedure 55b428b0d9e0 at gschem/gschemdoc.scm:50:5 (dir)> "/etc/xdg/lubuntu/lepton-eda"]
In unknown file:
   ?: 2 [string-join (#f "share" "doc" "lepton-eda") "/" #<undefined>]
In ice-9/boot-9.scm:
 105: 1 [#<procedure 55b428d99600 at ice-9/boot-9.scm:100:6 (thrown-k . args)> wrong-type-arg ...]
In unknown file:
   ?: 0 [apply-smob/1 #<catch-closure 55b428c7dec0> wrong-type-arg ...]

ERROR: In procedure string-append: Wrong type (expecting string): #f
```

This is due to a bug in `guess-docdir()` function that is simple
to fix (see the discussion on Gitter (29.11.2018) and
the attached patch by @vzh).

Generally, guess-work should be replaced by defining
compile-time variables as it is done in other utilities
(e.g. `lepton-netlist`).

[gschemdoc.diff.txt](https://github.com/lepton-eda/lepton-eda/files/2637016/gschemdoc.diff.txt)

graahnul-grom:
It is defined at compile time, so we can substitute `docdir` in `(gschem gschemdoc)` module. `sys-doc-dir()` function will return that value.

Closes #285

### DIFF
--- a/schematic/po/POTFILES.in
+++ b/schematic/po/POTFILES.in
@@ -110,5 +110,5 @@ schematic/scheme/conf/schematic/menu.scm
 schematic/scheme/gschem/action.scm
 schematic/scheme/gschem/builtins.scm
 schematic/scheme/gschem/deprecated.scm
-schematic/scheme/gschem/gschemdoc.scm
+schematic/scheme/gschem/gschemdoc.scm.in
 schematic/scheme/gschem/keymap.scm
--- a/schematic/scheme/Makefile.am
+++ b/schematic/scheme/Makefile.am
@@ -33,6 +33,11 @@ nobase_dist_scmdata_DATA = \
 	conf/schematic/options.scm \
 	conf/schematic/stroke.scm
 
+gschem/gschemdoc.scm: gschem/gschemdoc.scm.in
+	sed -e 's,[@]docdir[@],$(docdir),g' < $(srcdir)/$@.in > $@
+
+EXTRA_DIST = gschem/gschemdoc.scm.in
+
 MOSTLYCLEANFILES = *.log *~
 CLEANFILES = *.log *~
 DISTCLEANFILES = *.log core FILE *~
--- a/schematic/scheme/gschem/.gitignore
+++ b/schematic/scheme/gschem/.gitignore
@@ -1,0 +1,1 @@
+gschemdoc.scm
--- a/schematic/scheme/gschem/gschemdoc.scm.in
+++ b/schematic/scheme/gschem/gschemdoc.scm.in
@@ -2,7 +2,7 @@
 ;; Scheme API
 ;; Copyright (C) 2011-2014 Peter Brett <peter@peter-b.co.uk>
 ;; Copyright (C) 2011-2015 gEDA Contributors
-;; Copyright (C) 2017-2018 Lepton EDA Contributors
+;; Copyright (C) 2017-2019 Lepton EDA Contributors
 ;;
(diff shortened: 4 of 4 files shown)


vzh:
@graahnul-grom I have a couple of notes about your code. Please change things appropriately. 
- Renaming a file without changing anything in Makefiles is a way to catch false positives when you, e.g. would be going to search for a bug using `git bisect` together with, say, `make distcheck`.
- There is no reason to update po files in the same commit that brings other, non-related, changes. Moreover, I would avoid updating them more often than once per a release. It adds nothing but an appearance of many changes without real results. Over and above, while we have other still unmerged branches, the po-files often get outdated again after merging them.

graahnul-grom:
`po` files are always getting in the way when working with version control, it hinders the development process a lot. Is it possible to fix this, so they do not change every time the  code is modified? I'm not familiar with `gettext` enough, but [this page](https://www.gnu.org/software/gettext/manual/html_node/Translations-under-Version-Control.html) mentions `--no-location` option to `msgmerge` that eliminates comments in `po` files, though I have no idea how to use it.

vzh:
Shot in the dark: you could try `--disable-nls` on configure stage.
Generally, how much trouble you have with it depends on your workflow and tools. I, for example, keep my main repository clear and build in another place using `rsync` for `.git` to copy changes to it in that other place. OTOH, it's no problem to revert changes in po-files in `magit`.

graahnul-grom:
I have no problem checking out `po` files, too.
Even if `--disable-nls` would temporary solve the problem, sometimes we *have to* enable nls for testing. In that case we should be able to see real changes, and not changes that results from some side effects. Current situation is not normal.

graahnul-grom:
@vzh Could you please review the changes I've made according to your suggestions in this PR? That's all in one single commit now.

vzh:
Thanks!

graahnul-grom:
@vzh As it turns out,`--disable-nls` doesn't help.

PR #334, opened 2019-02-06
netlist: Create internally hierarchical refdes as a list.

vzh:
Using of lists instead of strings allows various transformations
of refdeses and fixes the following bugs (as can be seen from the
changed tests):

- Reverse order setting for refdes attribute affected
  netnames (see changes in output of tests
  'hierarchy-config_mangle_refdes_attribute_false_2.out' and
  'hierarchy-config_refdes_attribute_order_true.out' where the
  reverse refdes order is set).

- Now netnames are formed using true reversed hierarchical name,
  before they were formed from hierarchy-tag string and local
  netnames. Since the tag might be not reversed depending of the
  reverse order setting for refdes, the resulting strings did not
  reflect hierarchy (see, e.g., changes in
  'hierarchy-config_net_attribute_order_true.out').

- For the same reason, the separator for "net=" attributes did not
  affect resulting net names (see, e.g.,
  'hierarchy-config_net_attribute_separator_1.out').

- The same was true for "netname=" attributes (see, e.g.,
  'hierarchy-config_netname_attribute_separator_1.out').

- Removing of refdes mangling was possible only if the refdes
  separator was "/". See changes in the procedure
  remove-refdes-mangling().

### DIFF
--- a/netlist/scheme/netlist.scm
+++ b/netlist/scheme/netlist.scm
@@ -342,9 +342,13 @@ REFDES. As a result, slots may be repeated in the returned list."
 
 ;;; Helper function for sorting connections.
 (define (pair<? a b)
-  (or (refdes<? (car a) (car b))
-      (and (string=? (car a) (car b))
-           (refdes<? (cdr a) (cdr b)))))
+  (let ((refdes-a (schematic-component-refdes->string (car a)))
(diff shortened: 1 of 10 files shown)


vzh:
OK, no responses in a month. I'm applying the patch since it fixes several bugs and is easily revertable.

PR #341, opened 2019-02-21
docs/wiki: Allow out-of-source build and install

graahnul-grom:
Lepton EDA can be built out-of-source without
problems, but 'make install' does not install
files from the docs/wiki/ subdir, because it's
assumed that it contains regular files (-type f
switch for the 'find' command). When building
out of the source tree, they most likely will be
symbolic links.

### DIFF
--- a/docs/wiki/Makefile.am
+++ b/docs/wiki/Makefile.am
@@ -4,7 +4,7 @@ wikidir = $(docdir)/wiki
 dist-hook:
 	@for dir in $$(cd $(srcdir) && find . -type d -print | sed -e's:^\./::' ); do \
 	  $(MKDIR_P) $(distdir)/$$dir; \
-	  list=`(cd $(srcdir)/$$dir && find . -maxdepth 1 -type f \! -name 'Makefile*' \! -name '.gitignore' -print | sed -e 's:^\./::')`; \
+	  list=`(cd $(srcdir)/$$dir && find . -maxdepth 1 \( -type f -o -type l \) \! -name 'Makefile*' \! -name '.gitignore' -print | sed -e 's:^\./::')`; \
 	  for file in $$list; do \
 	    cp -p $(srcdir)/$$dir/$$file $(distdir)/$$dir || exit $$?; \
 	  done; \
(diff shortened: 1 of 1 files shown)


vzh:
@graahnul-grom, IIUC, you mean the build AKA VPATH build, right? If it is correct, I'm a bit confused. What symlinks exist and where? Could you please elaborate a bit more about this?

graahnul-grom:
@vzh I'm using `lndir (1)` to create a tree of symbolic links in a new directory, then build in that directory, so that the source tree remains intact:

`$ mkdir build`
`$ cd build`
`$ lndir /path/to/lepton.git/`
`$ ./autogen.sh`
`$ make`

vzh:
Ah, I've read about `lndir` in context of some famous *nix systems. Plain *VPATH* build is a bit other (JFTR):
```
mkdir build
cd build
../autogen.sh # once, it could be done before 'cd build'
../configure
make
```

and doesn't need a parallel symlink tree.

Well, while Debian has `lndir` in one of its packages, I don't see much advantage of using it over the way I described. OTOH, the patch is simple and I cannot express opposition to it.
Anyway, if you like it, please support the commit with better documentation, which could be included to NEWS and wiki.

graahnul-grom:
@vzh I've tried to build the project this way, and the build failed in `symbols/documentation/`:

```
Making all in documentation
Creating ps (using groff)
soelim: can't open 'geda.pic': No such file or directory
```

vzh:
Do you have any errors now, after removing the `geda.pic` dependency?

vzh:
@graahnul-grom And please look at #350.

graahnul-grom:
@vzh Builds successfully after I have applied the patch from #350, thank you.

vzh:
So what will we do with this PR? I think, it would be OK to have it for cases some people like or want to use your approach for some reason. In such a case, a couple of comments about why it's done this way would not harm. WDYT?

graahnul-grom:
I'll write something a bit later (or tomorrow), I have to go now, OK?

vzh:
Sure. Meantime I'll merge the commits since we agreed it's worth to have the code in the repo. Thanks!

graahnul-grom:
Something like this could go to the `NEWS` file. Should I elaborate?

`Lepton EDA` can now be built out-of-source using symbolic links tree (created, for example, with `lndir(1)`).

vzh:
@graahnul-grom 
I imagine changes in three places:
- comments in the source file: they should describe the `-l` option in order to prevent reverting of the changes by new / old  contributors who are not aware / forgot of why the option(s) had  been added
- wiki: a page with extended info about build possibilities  including listing of the approaches above describing their  advantages, as well as references to other docs; well, we  already have a wiki document in the source tree about building  using git sources, it would be nice to merge all the info  together
- NEWS: a short description like you have written (additionally we could say, that plain `VPATH` builds are now working as expected)

graahnul-grom:
@vzh I've added a comment to `docs/wiki/Makefile.am` (#351).

PR #342, opened 2019-02-22
Prevent unintentional updates of PO files

graahnul-grom:
It appears that at last we can get rid of those annoying unintentional changes to `PO` files, that happens during the build process. `PO` files, then, must be manually checked out. When doing hundreds and hundreds of builds, it can drive one really nuts.

1) add POT file to source control, do not ignore it
2) modify `Makevars`:
  - `MSGMERGE_OPTIONS = --no-location` (eliminates comments with line numbers)
  - `PO_DEPENDS_ON_POT = no` (do not modify `PO` files if it's not necessary)
3) `make update-po`

@vzh Please check, if it works for you. If everything fine, I can push here updates for the other tools (`liblepton`, `netlist`, `symcheck`, `cli`, `attrib`).

### DIFF
--- a/schematic/po/.gitignore
+++ b/schematic/po/.gitignore
@@ -9,7 +9,6 @@ POTFILES
 *.sin
 Rules-quot
 Makevars.template
-lepton-schematic.pot
 stamp-po
 stamp-it
 stamp-i18n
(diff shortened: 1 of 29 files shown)


vzh:
Well, the code works, let's give it a try. Thank you!

PR #344, opened 2019-02-25
Fix guile-2.2 deprecation warning in gsch2pcb backend.

vzh:
Port mode variable used by setvbuf() is now defined at the
compilation stage to prevent the deprecation warning due to using
deprecated guile-2.0 port buffering mode name.

Affects #332


vzh:
@graahnul-grom, Could you please test the fix?

PR #345, opened 2019-02-28
Refactor the macro widget code

graahnul-grom:
Beside refactoring, this PR fixes `lepton-schematic` crash after entering `(file-quit)` in the `:` prompt.

### DIFF
--- a/schematic/include/gschem.h
+++ b/schematic/include/gschem.h
@@ -1,3 +1,6 @@
+#ifndef LEPTON_MAIN_HEADER_H_
+#define LEPTON_MAIN_HEADER_H_
+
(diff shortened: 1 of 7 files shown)


vzh:
I'm curious, why are you using `&` in `G_CALLBACK (&func-name)`? Is it your preferred look/style or some of your tools prefers this?

vzh:
Thank you for the refactoring!

graahnul-grom:
It's a matter of style. Yes, in `C` we could write both `func-name` and `&func-name` and get a pointer to function. but in the latter form we state it explicitly (not to compiler, but for a human: "It's a pointer to function"). I like to be precise. :)

PR #346, opened 2019-02-28
Fix File->Save All

graahnul-grom:
On `File->Save All`, all opened pages are silently saved,
even if there are new "untitled" pages among them.
The user should be prompted to save new pages.

### DIFF
--- a/liblepton/include/liblepton/geda_page.h
+++ b/liblepton/include/liblepton/geda_page.h
@@ -92,6 +92,9 @@ PAGE*
 s_page_search (TOPLEVEL *toplevel, const gchar *filename);
 
 PAGE*
+s_page_search_by_basename (TOPLEVEL *toplevel, const gchar *filename);
+
+PAGE*
 s_page_search_by_page_id (GedaPageList *list, int pid);
 
 void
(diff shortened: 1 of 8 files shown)


vzh:
@graahnul-grom 
- Your code correctly skips names of already opened untitled files. Though it may be confusing for the user who saved such a file, and then did `lepton-schematic *.sch`. The next new file will be `untitled_2.sch`. Some log message why it is so would not harm, e.g. **Skipping existing file untitled_1.sch** in the log window.
- You have thrown out `i_set_state_msg` which is wrong, I believe. I for one would like to see some feedback in the case anything went OK without checking the log window and without thinking of possible bugs in the GUI code.
- In your last commit here, the function `x_window_set_current_page()` is used. It displays every page on canvas. On old machines (like I have) this leads to flashing GUI. Since you don't check if the page has been already saved, and if the number of opened pages is large, this slows down the GUI even more. Please avoid this.

graahnul-grom:
@vzh 
- Log messages about skipping files is a good point. I'll add them.
- How did I get `i_set_state_msg()` removed? :-) IIRC, that was not intended. Will be resurrected.
- I agree. Using `x_window_set_current_page()` is just an easy way to do the job. Refactoring of `i_callback_file_save()` is needed.

When I have time, I'll amend the code. Thank you for the review.

graahnul-grom:
@vzh All points you mentioned above are resolved.

vzh:
Third attempt was successful :-)

graahnul-grom:
Using `github`, I became very cautious: sometimes it behaves really strange (if not to say buggy) :-/.

PR #353, opened 2019-03-12
Clean up lepton-attrib

vzh:
- Get rid of useless Scheme rc functions
- Get rid of `gattribrc`

There is no point to support this obsolete rc file any more.


vzh:
Thank you :-)

PR #355, opened 2019-03-17
A couple of fixes in gmk_sym.

vzh:
- Get rid of non-working `-d` (debug) option
- Fix compilation warnings

### DIFF
--- a/contrib/gmk_sym/gmk_sym.1.in
+++ b/contrib/gmk_sym/gmk_sym.1.in
@@ -4,7 +4,7 @@
 gmk_sym \- create rectangular symbols for gschem from a text file
 
 .SH SYNOPSIS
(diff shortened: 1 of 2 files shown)


vzh:
I was always sure that it is written somewhere... for some reason...
And now found it again :-)
Please look at the info about GNU `indent` options that approximate the C indentation style most of lepton/geda devs adhered to. Especially, there is `-nut` ;-)
Those rules are not strong though it is recommended to use them in order the code to look not very odd.
As for new contributors, I would tolerate any style they use if the code works, softly suggesting them to avoid some aspects by pointing to the contributing document.
If you would prefer to notice this case explicitly there, please do. I won't mind.

graahnul-grom:
Ah! There's an `indent` command there, indeed. :-)

PR #356, opened 2019-03-18
Transform gschem.scm into 2 modules

vzh:
This simplifies code a bit and eliminates a superfluous step of loading the file.
OTOH, the procedures defined there are no more global, so the users may have to fix up their extensions by loading necessary modules in them.


graahnul-grom:
@vzh Great! Yet another step in (badly needed) reorganization of Scheme code. +1

PR #364, opened 2019-03-28
Move several functions into newly created lepton modules

vzh:
Several "geda.scm" functions have been refactored and moved into new modules with appropriate names.

### DIFF
--- a/liblepton/scheme/Makefile.am
+++ b/liblepton/scheme/Makefile.am
@@ -14,6 +14,7 @@ nobase_dist_scmdata_DATA = \
 	geda/log.scm \
 	geda/library.scm \
 	geda/repl.scm \
+	lepton/file-system.scm \
 	lepton/library.scm \
 	lepton/version.scm \
 	lepton/legacy-config/keylist.scm \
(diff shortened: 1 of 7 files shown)


graahnul-grom:
@vzh Is it possible to move `component-library-search()` somewhere
( to `(lepton library)` or `liblepton` where `component-library` is defined)
and get rid of `geda.scm` file?

vzh:
@graahnul-grom, in my view, `liblepton` should be replaced with a set of Guile modules (where possible). GUI functions should go to some other library (e.g. `libleptonui`, as we discussed before). I would make two library modules, one for sources/resources, and one for components (say, `(lepton library source)` and `(lepton library component)`). Common functions would go to an additional module (say, `(lepton library common)`), and `(lepton library)` would just reexport the top level functions from the above.  Now I think getting rid of `geda.scm` is wrong. Anyway, we have to initialize our Guile functions somewhere. Today we do it in C and in `*gafrc` though I consider this to be wrong as we're working to remove the *rc* files.  I would rename `geda.scm` to `lepton.scm`, move initialization rc func calls to it, and move the rc funcs to some `*deprecated*.scm` module.

graahnul-grom:
Agree.
geda.scm: maybe`startup.scm` or `init.scm`? though, `lepton.scm` looks quite appropriate.

graahnul-grom:
So, `lepton.scm` will replace `system-gafrc`?

vzh:
> So, `lepton.scm` will replace `system-gafrc`?

Yes, I believe.

PR #366, opened 2019-04-04
Page manager: toggle show full paths + refactoring

graahnul-grom:
This was suggested in launchpad bug report [#721929 "pagemanager is cluttered with paths"](https://bugs.launchpad.net/geda/+bug/721929), and I fully agree with @KaiMartin: when working with hierarchical schematics, full paths just clutters the view most of the time. Now they can be shown at any time or hidden away by checking the "Show full paths" checkbox. By default, they are hidden.
The state of the "Show full paths" checkbox is saved to/loaded from the CACHE configuration context: `show-full-paths` key (boolean), `schematic.page-manager` group.

Additional changes:
- minimum "Filename" column width is reduced
- scrollbars are shown only when needed
- "Right click on the filename for more options..." label is removed

<br />

tb132_pmgr_basenames

tb132_pmgr_paths

### DIFF
--- a/schematic/include/x_pagesel.h
+++ b/schematic/include/x_pagesel.h
@@ -22,7 +22,6 @@
 
 
 typedef enum {
-  PAGESEL_RESPONSE_CLOSE  = 1,
   PAGESEL_RESPONSE_UPDATE = 2
 } PageselResponseType;
 
--- a/schematic/include/x_pagesel.h
+++ b/schematic/include/x_pagesel.h
@@ -21,11 +21,6 @@
 #define LEPTON_PAGESEL_H_
 
 
-typedef enum {
-  PAGESEL_RESPONSE_UPDATE = 2
-} PageselResponseType;
-
-
 #define TYPE_PAGESEL         (pagesel_get_type())
 #define PAGESEL(obj)         (G_TYPE_CHECK_INSTANCE_CAST ((obj), TYPE_PAGESEL, Pagesel))
 #define PAGESEL_CLASS(klass) (G_TYPE_CHECK_CLASS_CAST ((klass), TYPE_PAGESEL, PageselClass))
--- a/schematic/include/x_pagesel.h
+++ b/schematic/include/x_pagesel.h
@@ -39,6 +39,8 @@ struct _Pagesel {
   GschemDialog parent_instance;
 
   GtkTreeView *treeview;
+
+  gboolean show_full_paths;
 };
 
 
--- a/schematic/include/x_pagesel.h
+++ b/schematic/include/x_pagesel.h
@@ -17,6 +17,8 @@
  * MA 02111-1301 USA.
  */
 
+#ifndef LEPTON_PAGESEL_H_
+#define LEPTON_PAGESEL_H_
 
 
 typedef enum {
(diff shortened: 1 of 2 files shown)


graahnul-grom:
@vzh Could you please review this PR? Yesterday it was celebrating its 2nd
"monthaversary". :-)

vzh:
I'm trying to do my best in my current situation :-/ Know, not very successful. Well, my thoughts/review as follows (... snipped annoying notes about useless whitespace changes ;)):

# Several thoughts aload:

## 81cf368d4
I may be wrong, though I think
`gtk_dialog_set_alternative_button_order` is not just for some
simple dialogs. Gtk manual says it is not for dialogs of some
type, but for systems where default value of
`gtk-alternative-button-order` is *TRUE*.  I could imagine a
situation if a dev have used to Windows (and built the code for
it) just having thought the order chosen suits better for his
OS, while the code above meantime determined some other order.
(Didn't check the history of the file, though, sorry.)
OTOH, now the alternative order is the same as the initial order,
so nothing wrong at first glance.

## 2c5f95e39 (and others)
`enum PageselResponseType`--does it really contain an only value?
Could it be replaced by a variable?

## fc9944678
Just an idea: for hierarchical schematics, instead of just
basenames, it would be probably better to show paths relative to
the toplevel schematic. WDYT?

However, for today, as we have no concept of one toplevel schematic,
it would not work well.

# ... and a question:

## 51c12db61
Why do you think the help label about right click menu in page
manager is not helpful? And how does the commit interfere with the
main idea of the patchset?

graahnul-grom:
> `gtk_dialog_set_alternative_button_order`

It would be confusing to swap "Refresh" and "Update" buttons.
Users expects `gtk-alternative-button-order=true` to be
applied to buttons that accepts or rejects dialogs.


> `enum PageselResponseType`--does it really contain an only value?
Could it be replaced by a variable?

Yes, but I would leave that `enum` as is: maybe we'll add some
new buttons later.


> Just an idea: for hierarchical schematics, instead of just
basenames, it would be probably better to show paths relative to
the toplevel schematic. WDYT?

I thought about it, too, while working on this PR.
It's not obvious how to do it consistently, since
a project files hierarchy may be different from what
is shown in the page manager. Anyway, it's for another PR.


> Why do you think the help label about right click menu in page
manager is not helpful?

Context menu is a very well known element of the GUI to
remind the users that it may exist and how to show it.
For example, there is no such label in the "Edit Attributes" dialog.

graahnul-grom:
@vzh Thank you for the review!
One more question: may I start removing deprecated exporting stuff (#419)?

vzh:
> One more question: may I start removing deprecated exporting stuff (#419)?

Sure. Go for it.

graahnul-grom:
@vzh Could you please try to open several files in `lepton-schematic`, then open
the page manager and press PgUp/PgDown keys in pagesel tree view. On my system
`lepton-schematic` crashes. The same with 1.9.7 release.

vzh:
Confirmed. On my system it crashes with the following output:
> Gtk:ERROR:/build/gtk+2.0-ZErHoX/gtk+2.0-2.24.31/gtk/gtkrbtree.c:967:_gtk_rbtree_node_find_offset: assertion failed: (node->left)

graahnul-grom:
`Gtk:ERROR:gtkrbtree.c:967:_gtk_rbtree_node_find_offset: assertion failed: (node->left)`

PR #371, opened 2019-04-08
Use CSS in Scheme API HTML reference pages

graahnul-grom:
Add Cascading Style Sheets (CSS) file, use it
in generated HTML Scheme API reference pages.
Now it sets only background color (to be slightly
darker - to reduce eye fatigue).


vzh:
Hi @graahnul-grom,
the patch doesn't work on `vpath` build for me (`mkdir build; cd build; ../autogen.sh && ../configure && make html`). Probably, some other variable should be used instead of local `srcdir`?

graahnul-grom:
@vzh Fixed; thank you!

PR #372, opened 2019-04-09
Catch exceptions in the (gschem gschemdoc) lepton-schematic module

graahnul-grom:
Catch exceptions that can be thrown by `show-uri()` function
called in the `(gschem gschemdoc)` module.
Show a message box with error information.

tb_136_showuri_catch

### DIFF
--- a/schematic/scheme/gschem/builtins.scm
+++ b/schematic/scheme/gschem/builtins.scm
@@ -395,30 +395,30 @@ found, shows a dialog with an error message."
 
 
 (define-action-public
-    (&help-manual #:label (_ "gEDA Manuals") #:icon "help-browser"
-     #:tooltip (_ "View the front page of the gEDA documentation in a browser."))
+    (&help-manual #:label (_ "Lepton EDA Manuals") #:icon "help-browser"
+     #:tooltip (_ "View the front page of the Lepton EDA documentation in a browser."))
   (show-wiki "geda:documentation"))
 
 
 (define-action-public
-    (&help-guide #:label (_ "gschem User Guide") #:icon "gtk-help"
-                 #:tooltip (_ "View the gschem User Guide in a browser."))
+    (&help-guide #:label (_ "lepton-schematic User Guide") #:icon "gtk-help"
+                 #:tooltip (_ "View the lepton-schematic User Guide in a browser."))
   (show-wiki "geda:gschem_ug"))
 
 
 (define-action-public
-    (&help-faq #:label (_ "gschem FAQ") #:icon "help-faq"
-     #:tooltip (_ "Frequently Asked Questions about using gschem."))
+    (&help-faq #:label (_ "lepton-schematic FAQ") #:icon "help-faq"
+     #:tooltip (_ "Frequently Asked Questions about using lepton-schematic."))
   (show-wiki "geda:faq-gschem"))
 
 
 (define-action-public
-    (&help-wiki #:label (_ "gEDA wiki") #:icon "web-browser"
-     #:tooltip (_ "View the front page of the gEDA wiki in a browser."))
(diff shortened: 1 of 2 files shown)


vzh:
@graahnul-grom While your code as a whole suits me, it still will be
nice if you take my notes below into account.

**Function name**

I would add a dash in the function name: `gschemdoc-show-uri`.

**set! vs let**
I would recommend you reading the following sources regarding
using of `set!` and mutation in general. I added some noteworthy
quotations below wrt code hygiene, coding culture, and overhead in
usage of resources.

[Internal definitions](https://www.gnu.org/software/guile/manual/html_node/Internal-Definitions.html)

... Another noteworthy difference to top level definitions is that within
one group of internal definitions all variable names must be distinct.
That means where on the top level a second define for a given variable
acts like a ‘set!’, an exception is thrown for internal definitions with
duplicate bindings...


[Sandboxed Evaluation](https://www.gnu.org/software/guile/manual/html_node/Sandboxed-Evaluation.html)
... A lot of code culturally doesn’t expect mutation...

... Relatedly, ‘set!’ may allow a sandbox to mutate a primitive,
invalidating many system-wide invariants.  Guile is currently
quite permissive when it comes to imported bindings and
mutability.  Although ‘set!’ to a module-local or lexically bound
variable would be fine, we don’t currently have an easy way to
disallow ‘set!’ to an imported binding, so currently no binding
set includes ‘set!’...

[Variables and the VM](https://www.gnu.org/software/guile/manual/html_node/Variables-and-the-VM.html#Variables-and-the-VM)
... ‘set!’, actually forces an extra memory allocation and
indirection ...

In my experience, `set!` in some cases can even block creating of
tail-calls in loops, making them non-tail-recursive.

What your function could look like:
```
(define (gschemdoc-show-uri url)
  (catch
    #t
    (lambda () (show-uri url))
    (lambda (key subr msg args . rest)
      (let* ((exmsg (apply format #f msg args))
             (docmsg (_ "Could not show documentation:"))
             (ermsg (format #f "~a~%'~a in ~a():~%~%~a" docmsg key subr exmsg)))
        ((@@ (guile-user) gschem-msg) ermsg)))))
```

graahnul-grom:
> I would add a dash in the function name: gschemdoc-show-uri

I would also drop "gschem": we're going to rename modules, anyway.

graahnul-grom:
Thank you for the pointers, I'll study the docs mentioned above
and will keep these features in mind. 
I agree that variables should be declared as locally as possible; however,
in this particular function (I'll amend it as you've suggested) there's nothing
special, so code readability was preferred (as it should, in most cases, I believe).

vzh:
:-) Please look up for "preferred Scheme (or Racket) code style" in the net. I hope you'll find some pointers of how to improve Scheme code readability useful.

vzh:
BTW, is it really necessary for `doc-show-uri` to return `#t` or `#f`?

graahnul-grom:
Good point, the return is not used.

graahnul-grom:
@vzh You mentioned elsewhere that you are compiling updates to the NEWS.md
in some local branch. Even if it's not complete I think it would be nice to publish
something under "Notable changes since Lepton EDA 1.9.7" to inform user
of what's going on.

vzh:
@graahnul-grom The work on NEWS is sporadic and there's nothing to boast of. For now, I just gathered a heap of info that needs much more work to filtrate.

vzh:
I'll merge this branch now, however IMO it's not complete. `doc-show-uri` needs to be somehow merged with `show-uri`, otherwise the *show uri* functionality of those two functions will diverge and the users cannot see and use your improvements.

graahnul-grom:
Yes, error handling could be done in `(gschem util)::show-uri()`.

PR #377, opened 2019-04-16
Rewrite attributes detachment code in lepton-schematic

### LINKED ISSUE #376: Attributes->Detach is useless
`attributes-detach` just strips the selected objects of ALL their attributes (visible and invisible, selected or not). That's all. This nonsense is there since at least `gschem 1.8.0`.

The reasonable behavior would be something like this:
- detach only selected attributes from their respective  parent objects
- do not require object to be selected (attributes can be attached to one object only, and it's easy to determine the object they're attached to)
- print successfully detached attributes to the log (`Attribute detached: [refdes=R1]`)
- if selection is empty, selected attributes are detached or no attributes selected, do nothing
- deselect attributes just detached (to make it clear what was detached - attributes' color will change)

graahnul-grom:
Old behavior was to unconditionally strip all
attributes of the selected objects (visible
and invisible, selected or not).
It has been changed:
-  detach only selected visible attributes
from their respective parent objects
- do not require parent object to be selected
- print successfully detached attributes to
the log
- if selection is empty, selected attributes
are detached or no attributes are selected,
do nothing
- deselect attributes just detached

The code is rewritten in `Scheme` and placed into the `(gschem builtins)` module.
Closes #376


schematic/src/i_callbacks.c
```diff
@@ -2459,52 +2459,7 @@ DEFINE_I_CALLBACK(attributes_attach)
...
-  g_return_if_fail (w_current != NULL);
 
-  /* This is a new addition 3/15 to prevent this from executing
-   * inside an action */
-  if (w_current->inside_action) {
-    return;
-  }
-
-  /* same note as above on i_update_middle_button */
-  i_update_middle_button(w_current, i_callback_attributes_detach,
```
vzh: `(middle-button "repeat")` won't work any more with the
`attributes-detach` action since `i_update_middle_button()` is not
used in your Scheme replacement procedure.  I don't know how to
deal with it in Scheme just now :-/

schematic/src/i_callbacks.c
```diff
@@ -2459,52 +2459,7 @@ DEFINE_I_CALLBACK(attributes_attach)
...
-
-  s_current = geda_list_get_glist( gschem_toplevel_get_toplevel (w_current)->page_current->selection_list );
-  while (s_current != NULL) {
-    o_current = (OBJECT *) s_current->data;
-    if (o_current) {
-      if (o_current->attribs) {
-        detached_attribs = g_list_concat (g_list_copy (o_current->attribs),
-                                          detached_attribs);
-        o_attrib_detach_all (gschem_toplevel_get_toplevel (w_current), o_current);
-        gschem_toplevel_get_toplevel (w_current)->page_current->CHANGED=1;
```
vzh: You've missed setting the page *dirty* flag after detachment in you Scheme code.

schematic/scheme/gschem/builtins.scm
```diff
@@ -310,8 +315,51 @@
...
+  )
+
+  ( let*
+    (
+    ( page ( active-page ) )
+    ( sel  ( if page (page-selection page) '() ) )
+    ( attrs '() )
+    )
+
+    ( define ( is-attr-to-detach obj ) ; predicate
```
vzh: `is-attr-to-detach()` is a predicate, in Scheme world predicates are usually named with question mark appended.  I would recommend you to add it.  And I would suggest another name for this procedure: `detachable-attrib?`. WDYT? (Don't know if it sounds correct in English.)

schematic/scheme/gschem/builtins.scm
```diff
@@ -310,8 +315,51 @@
...
+    )
+
+    ( define ( detach-attr attr )
+      ( detach-attribs! (attrib-attachment attr) attr )
+      ( log! 'message (_ "Attribute detached: [~a]") (text-string attr) )
+      ( deselect-object! attr )
+    )
+
+
+    ( set! attrs (filter is-attr-to-detach sel) )
```
vzh: Please use `let` here instead of `set!` as we discussed before wrt other topic.

schematic/scheme/gschem/builtins.scm
```diff
@@ -310,8 +315,51 @@
...
+
+    ( when ( > (length attrs) 0 )
```
vzh: This can be safely replaced with `(unless (null? attrs) ...)`, which seems to be a bit more readable.

schematic/scheme/gschem/builtins.scm
```diff
@@ -310,8 +315,51 @@
...
+      ( for-each detach-attr attrs )
+      ( run-hook detach-attribs-hook attrs )
+      ( undo-save-state )
+    )
+
+  ) ; let
```
vzh: I would recommend to return `attrs` as the result of `&attributes-detach` to let the users use it further in their scripts.

schematic/src/i_callbacks.c
graahnul-grom: All that "repeat" stuff looks odd to me; probably, it should be somehow
connected with actions (there is already "repeat last action" command).
And mouse buttons indicators (IMHO) brings nothing but confusion and
complicates the code.

schematic/src/i_callbacks.c
graahnul-grom: Thanks! Fixed.
There is a general problem with `set-page-dirty!`: it's a library function, and
we can't determine if it was called in `lepton-schematic`. So, for example, we
cannot update GUI, indicating that the page has been changed.

schematic/scheme/gschem/builtins.scm
graahnul-grom: I think `detachable` is OK.

schematic/src/i_callbacks.c
vzh: We could make up a Scheme callback just for `lepton-schematic` using *glib* functionality to send an appropriate message to the program.  Likewise `log!` and the stuff around it used for message logging in or GUI and non-GUI programs now.

schematic/src/i_callbacks.c
vzh: My concern is mostly not about the GUI (BTW,  I cannot remember when and why the mouse button indication got removed from the status line, maybe you could recall?).  Our action code allows using dot (`.`) to repeat the last action. Can it somehow be used to do the same by the middle mouse button?

graahnul-grom: Indicators weren't removed, but can be hidden by setting `show-mouse-buttons`
to `false` (`schematic.status-bar` group). They should be visible, by default.
`&repeat-last-action` can easily be invoked by the middle button; could you
please merge your fixes to the actions code first (#435)?
And I would remove `i_update_middle_button()` and associated code (it's
simply ugly) and display just "Repeat" in the middle mouse button indicator.
These indicators already takes up too much space in the status bar (fortunately,
they can be turned off now). I seriously doubt in their usefulness. We'd better
expand cryptic `Grid(100,100)` string.

schematic/src/i_callbacks.c
graahnul-grom: Although logging is implemented using *glib* functions, `lepton-schematic`
receives log messages simply via a global function pointer
(`x_log_update_func`, `s_log.c`).
Page change notifications could be done similarly, but it's not an easy task
with the current code state and it's not enough. Scheme code should also be
notified (if possible, via guile hooks).
`CHANGED` field of `st_page` structure should be made private and modification
status should be accessed via functions and in a centralized manner.
Now there is `gschem_toplevel_page_content_changed()` function.
Comment says that *"This function is temporary until library functions
can emit signals"*. Why temporary? What signals?
All this stuff makes an impression of incomplete, disorganized code;
we need to fix it.

schematic/src/i_callbacks.c
vzh: Middle mouse button works as expected.  Again, I didn't managed to prove it (I definitely need some rest :-)), I'm sorry.  OK, I'm going to merge #435 and this PR. Is the last code snippet about to fix the GUI issue we discussed above? If so, please make another PR and we'll finish with it.

schematic/src/i_callbacks.c
vzh: It was temporary because other devs didn't expect the events go the way we watch now. Previously, things were worse, AFAICR. Now it's our task to make them better. It was planned that a C function editing a page should send, say, the "changed" signal to the program, and notify the guile side as well.  The development of `geda-gaf` went parallel in two (or even more) separate ways. Naturally, some people prefer C and gobjects (like Edward), others are admiring for Scheme (like me).  The previous code base had neither of those two approaches. Most people just wanted to improve the C code we had because despite of its quirks it's quite usable, isn't it? ;-)

schematic/src/i_callbacks.c
graahnul-grom: Middle button indicator will not be updated. The code above shows how
to make middle mouse button invoke "repeat last action" (`.`) command.
It could be done after removing `i_update_middle_button()` stuff and
making middle button indicator text static. Would you mind if I start this cleanup?

vzh: Since I haven't understand how strong the first two sentences are related to the latter ones, I'll try to answer them apart:

> The code above shows how
to make middle mouse button invoke "repeat last action" (.) command.

I'm all for using high-level Scheme language instead of low-level C where possible or convenient. OTOH, you know, I'm not for changes for the sake of changes, I'm for convenience and productiveness...

> It could be done after removing i_update_middle_button() stuff and
making middle button indicator text static. Would you mind if I start this cleanup?

...Please go for it if you think things will be better with your changes (in using, coding, etc.) and please stop it, if not.

graahnul-grom: Yes, this will free us from updating almost useless static indicators on each
and every action (cleaner, faster code), will allow to use high-level code
(`&repeat-last-action`) and it will make possible to move more code to
Scheme (I'm going to do so with `Attributes->Attach` next, which is also
not working well).

PR #378, opened 2019-04-18
Fix preview in the lepton-schematic font selection widget

graahnul-grom:
If a custom font is specified in the configuration (`schematic.gui::font`),
there is no preview text visible in the "Preview" box, because the "Size" field
is set to `0` (font size has a separate setting: `text-size` in `gschemrc`).
Fix this by appending font size to the font name string set for
`GtkFontSelection` widget.

### DIFF
--- a/schematic/src/font_select_widget.c
+++ b/schematic/src/font_select_widget.c
@@ -26,6 +26,9 @@
 #include "gschem.h"
 
 
+#define PREVIEW_TEXT_SIZE 18
+
+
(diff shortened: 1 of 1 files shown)


vzh:
Strange a bit size (18) for a document text. Standard sizes in text editors/processors are 11, 12, and 14. OTOH, since we edit schematics, and our preferred font is OpenGOST... I cannot but second this ;-)

vzh:
GH starts to get annoying with its AI :-/

PR #379, opened 2019-04-18
Fix repeat-last-action (the "." key) in lepton-schematic

graahnul-grom:
Make `&repeat-last-action` work.
Display a log message when there is no last action to repeat.


vzh:
Nice work, thanks!

PR #392, opened 2019-05-13
Prevent unwanted updates of PO files: cli, liblepton, netlist

graahnul-grom:
The same changes as in #342, #389:

1) add `pot` file to source control
2) modify `Makevars`:
  - `MSGMERGE_OPTIONS = --no-location`
  (eliminates comments with line numbers in `*.po`)
  - `PO_DEPENDS_ON_POT = no`
  (do not modify `PO` files if it's not necessary)
3) run `make update-po` in `po/` directory

### DIFF
--- a/cli/po/.gitignore
+++ b/cli/po/.gitignore
@@ -7,7 +7,6 @@ POTFILES
 *.sed
 *.header
 *.sin
-lepton-cli.pot
 Rules-quot
 Makevars.template
 stamp-po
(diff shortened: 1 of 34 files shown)


graahnul-grom:
Fr**king GH at it again. As it turned out, it sends PR if you accidentally press 'enter' when the subject field is focused. 
@vzh I wonder, did you ever consider moving the project somewhere else? To gitlab, for example?

vzh:
No, we started here because @peter-b had worked here quite a bit before. I haven't thought about moving somewhere else since then.

PR #393, opened 2019-05-13
Escape status bar message strings in lepton-schematic

graahnul-grom:
`gtk_label_set_markup()` is used to display status
messages in the status bar. Message string needs
to be escaped, so that special characters (e.g.
accelerator keys, such as '<' and '>' ) are
displayed correctly.

Otherwise, you get a ` Gtk-WARNING` like this:
```
Failed to set text from markup due to error parsing markup: Error on line 1 char 13: ' ' is not a valid character following a '<' character; it may not begin an element name
```


vzh:
Indeed, the bug was present at least in 1.9.7. Thank you for the fix!

PR #394, opened 2019-05-14
Get rid of gattrib-menus.xml file

graahnul-grom:
Move menu description `XML` code to `x_window.c`:
define it as static string and pass it to
`gtk_ui_manager_add_ui_from_string()`.


vzh:
What's the benefit of it?

graahnul-grom:
- place two closely related data structures together:
menu description and corresponding actions array
- less files to maintain and install
- no need to search for and load menu file, faster startup

IMHO, it's quite reasonable. And I wouldn't say it's *highly likely (c)*
that `lepton-attrib`'s menu will face an exponential growth in the near future. :)

graahnul-grom:
@vzh An off-topic here, but do you remember 
`lepton-schematic -c '(file-quit)'` crash?
I found out that if a call to [`gtk_main_quit()`](https://github.com/lepton-eda/lepton-eda/blob/3aaf5190bc49d44f24b12af5386c21a0beded0ab/schematic/src/lepton-schematic.c#L149) is replaces with just `exit(0)` in `gschem_quit()`, it stops crashing.
Is it safe to do this substitution?

vzh:
>     * place two closely related data structures together:
>       menu description and corresponding actions array
> 
>     * less files to maintain and install
> 
>     * no need to search for and load menu file, faster startup
> 
> 
> IMHO, it's quite reasonable. And I wouldn't say it's _highly likely (c)_
> that `lepton-attrib`'s menu will face an exponential growth in the near future. :)

That's what should have been written in the PR description :-)

vzh:
> @vzh An off-topic here, but do you remember
> `lepton-schematic -c '(file-quit)'` crash?
> I found out that if a call to [`gtk_main_quit()`](https://github.com/lepton-eda/lepton-eda/blob/3aaf5190bc49d44f24b12af5386c21a0beded0ab/schematic/src/lepton-schematic.c#L149) is replaces with just `exit(0)` in `gschem_quit()`, it stops crashing.
> Is it safe to do this substitution?

Dunno.

vzh:
@graahnul-grom, after a little more thinking... Probably the issue is in that we're trying to stop the `gtk_main` loop while it is not initialized yet, since it should be initialized last in `main_prog()` and we quit before.

graahnul-grom:
@vzh Yes, and we can check for it and act accordingly. So, I think the following solves the problem. WDYT?

```diff
diff --git a/schematic/src/lepton-schematic.c b/schematic/src/lepton-schematic.c
index 17968d106..96089efce 100644
--- a/schematic/src/lepton-schematic.c
+++ b/schematic/src/lepton-schematic.c
@@ -146,7 +146,16 @@ void gschem_quit(void)
   i_vars_freenames();
   i_vars_libgeda_freenames();
 
-  gtk_main_quit();
+  /* Check whether the main loop is running:
+  */
+  if (gtk_main_level() == 0)
+  {
+    exit (0);
+  }
+  else
+  {
+    gtk_main_quit();
+  }
 }
 
 /*! \brief Main Scheme(GUILE) program function.
```

vzh:
I think it's strange there is no corresponding PR yet ;-)

PR #397, opened 2019-05-19
Add system configuration files with default settings

graahnul-grom:
Add `geda-system.conf` and `lepton-system.conf` configuration files.

`geda-system.conf` was generated by `lepton-conf -p`, then it was converted
to `lepton-system.conf ` by `lepton-upcfg` utility.


graahnul-grom:
git conflict resolved.

vzh:
What default values will be used if both `*-system.conf` files are missing for some reason?

graahnul-grom:
Currently, all default values are defined in the code, anyway, so nothing changes.
Perhaps, we should check for presence of system configuration files after
migration to new config is done.

PR #399, opened 2019-05-21
Build and install Scheme API reference HTML pages

graahnul-grom:
Now `make` and `make install` will build and install
`Lepton EDA Scheme Reference Manual` in `HTML` format.


graahnul-grom:
`makeinfo` translates "`-`" into "`_002d`" in the names of html files.
Is there a way to control this behavior?

PR #402, opened 2019-05-21
Make several rc functions return their values.

vzh:
The output of each function has been modified to get its value set by the user instead of just *#t* (or *#f* in the case of an error).  Their results will now be available for using in Scheme scripts.  Affected functions are as follows:
- `bitmap-directory`
- `attribute-promotion`
- `promote-invisible`
- `keep-invisible`


graahnul-grom:
Apropos, about `bitmap-directory`: this is set to `share/lepton-eda/bitmap`
directory (installed by lepton-schematic). In effect, the *library* depends on *application*; that is wrong, I think. `liblepton` uses `gschem-warning.png`
file from that directory (`geda_picture_object.c`).

vzh:
Agreed, that's wrong. `liblepton` is to be made not dependent on images.

graahnul-grom:
At least, we could move that image to the `liblepton` source tree.

vzh:
We definitely need another, GUI, library, `liblepton(g)ui`, as we discussed somewhere else, in order to not overload the code of `liblepton` and of each of our GUI apps (currently, `lepton-schematic` and `lepton-attrib`) and move their common functionality there.

PR #403, opened 2019-05-22
Improve the "Write image..." dialog box

graahnul-grom:
- combo box to select the color mode for export: color or grayscale
- warn that different color map will be used for `PDF` export (label)
- add 1024x768 to the list of sizes
- remember and restore selected parameters:
  - folder path
  - image size
  - output file type
  - color mode

@vzh I would also use the `display` color map when exporting `PDF` - for consistency. What do you think about that?

@vzh What about removing `scripts/image.scm`, `scripts/print.scm`
(and, hence, `--output`, `gschem-pdf()`, `gschem-image()`, `image-color` option)? Deprecated since 2012.

tb_160_write_img_color ss


graahnul-grom:
Need to merge this; it keeps me from working on #419 
(remove `image-color` `gschemrc` option).

vzh:
> @vzh I would also use the display color map when exporting PDF - for consistency. What do you think about that?

Not sure. At least, the user should have a choice.

Issue #369, opened 2019-04-06
Reasonable defaults


graahnul-grom:
After more than two years of collective effort that was put
into making it better, `Lepton EDA` has become quite different
compared to what it was forked from.
However, it's hard to spot the difference at a glance:
almost all new features are disabled by default
(as if we so shy to show them) and, on the other hand,
old, obsolete settings are retained.
Before making a new release I propose to change some defaults.

1. Turn off cursor warping: `(warp-cursor "disabled")`.
In `conf/schematic/options.scm` it is referred to as
"annoying". It's not "annoying", it is outright irritating.
Those who like it knows very well how to turn it on, I'm sure.
But when the infuriated (new) user rushes to `Options` menu,
eager to immediately turn this devil's invention off, what
will he see? There are no configuration settings in `Options`
menu, just some volatile UI tweaks. What then? Try kicad?
Thank you very much. Their schematic editor is not an application,
but an instrument of torture. Could easily take its place in
[torturemuseum.com/instruments-of-torture](http://www.torturemuseum.com/instruments-of-torture/).
It's not worth pushing users to that route.
**[UPDATE]:** fixed in #407

2. Turn off fast mousepan: `(fast-mousepan "disabled")`.
20 years ago it was quite relevant, but with modern (and not
so modern) hardware, drawing boxes instead of text while panning
is not necessary. In exceptional cases (extremely old hardware,
insanely large schematics) it can be enabled, of course.
**[UPDATE]:** fixed in #407 

3. Do not show the log window on startup: `(log-window "later")`.
It may be not so important with docking windows GUI,
but otherwise it's quite annoying.
**[UPDATE]:** fixed in #407 

4. Increase the number of undo steps: `(undo-levels 20)`.
Now it's set to 10. Should be at least 20, I think.
**[UPDATE]:** fixed in #407 

5. Mouse panning with the right mouse button: `(third-button "mousepan")`.
Not only it's inconvenient to move mouse while holding
the middle button when it's actually a wheel (on some mice it may even
tilt), panning with the right mouse button is used in many pcb layout
tools. The advantage of having consistent mouse buttons mapping
becomes obvious when working on pcb project and switching frequently
between schematic and layout editors.
Context menu can be shown on middle-click: `(middle-button "popup")`.
Not a big deal: we have to admit that there is little use for our context
menu at present.

6. Enable tabbed GUI: `schematic.gui::use-tabs = true`.
[...]

PR #404, opened 2019-05-23
Install symbols with conflicting names into separate folders

graahnul-grom:
- install `gnetman` symbols into `share/lepton-eda/sym-gnetman/`
- install `verilog` symbols into `share/lepton-eda/sym-verilog/`
- install `vhdl` symbols into `share/lepton-eda/sym-vhdl/`
- all other symbols are installed, as usual, into `share/lepton-eda/sym/`

After these changes, `component-library-search` can be used with
`share/lepton-eda/sym/` without causing symbol name conflicts.

Affects #369


symbols/gnetman/Makefile.am
```diff
@@ -0,0 +1,18 @@
...
+	cp -p $(srcdir)/*.[sS][yY][mM] $(distdir)
+
+install-data-local:
+	$(MKDIR_P) $(DESTDIR)$(datasymdir)
+	$(INSTALL_DATA) $(srcdir)/*.[sS][yY][mM] $(DESTDIR)$(datasymdir)
+
+uninstall-local:
+	rm -fr $(DESTDIR)$(datasymdir)
+
+MOSTLYCLEANFILES = *.log *~
```
vzh: I would not use those patterns for `*CLEANFILES` from pretty old makefiles, all those log, core and backup files. Please use the names you really found are there in most situations.

graahnul-grom: IIUC, only `MAINTAINERCLEANFILES = Makefile.in` line should be in 
`{gnetman,verilog,vhdl}/Makefile.am`, right?

graahnul-grom: However, according to `automake` [manual](https://www.gnu.org/software/automake/manual/html_node/Clean.html)
```
maintainer-clean should not delete anything that needs
to exist in order to run ‘./configure && make’
```

I see `MAINTAINERCLEANFILES = Makefile.in` in many places,
but `./configure` fails if it cannot find `Makefile.in`:
```
config.status: error: cannot find input file: `symbols/gnetman/Makefile.in'
```
So, is it a bug in `Makefile.am` files?

vzh: AFAIR, I deleted whole `MAINTAINERCLEANFILES` lines before (along with `Makefile.in`) and nothing changed. Maybe it was needed some time ago for autotools, though nowadays I don't see any reason to leave it as is.

graahnul-grom: OK, `*CLEAN*` lines removed.

PR #405, opened 2019-05-23
contrib/scripts: Fix build from the distribution tarball sources

graahnul-grom:
Build fails in `contrib/scripts/` when compiling
from the distribution tarball sources (made by
`make dist`) with `--enable-contrib`, because of
missing `sch2eaglepos.sh` in `EXTRA_DIST`.


vzh:
Hmm, I cannot trigger the issue.
But fair enough...

PR #406, opened 2019-05-23
Fix: lepton-schematic cannot find its gtkrc file

graahnul-grom:
Installed `gschem-gtkrc` file cannot be loaded
because wrong directories were searched for it:
we need to use `eda_get_system_config_dirs()`,
not `g_get_system_config_dirs()`.


vzh:
How to trigger the issue or understand if it exists at all?

graahnul-grom:
Just place some `gtkrc` commands in`$PREFIX/share/lepton-eda/gschem-gtkrc`
(for example, `gtk-theme-name="<theme name>"`).

vzh:
OK, confirmed.

PR #409, opened 2019-05-24
Move netlist backends to a separate folder

graahnul-grom:
Move `netlist` backend files to `backend/` subdirectory, thus making
our `scheme` directory less disorganized.
This folder is added to guile `%load-path` in `gafrc.d/config-netlist.scm`
configuration file (which is loaded by `system-gafrc`).
The next step should be the conversion of backend scm files to guile modules,
so that we could, e.g., `( use-modules ( backend spice ) )`.


graahnul-grom:
@vzh Could we merge it, please?
I'm really tired of this pile of files in our `scheme` directory.

graahnul-grom:
Ура!

PR #412, opened 2019-05-26
Catch Netlist->allegro errors in lepton-schematic

graahnul-grom:
Check errors in code invoked by `Netlist->allegro` menu item:
the `(schematic netlist)` module.

If `allegro*` function fails, backtrace is printed to the log.
If for some reason `allegro` backend cannot be found, `lepton-schematic`
starts in horrible broken state: no menu, no color map, no actions available;
eventually it crashes. All issued errors are absolutely irrelevant, it's very hard to find what actually is wrong.

So, it's better to do some `DBCS` gymnastics now than spend time on debugging later :-).

### DIFF
--- a/schematic/scheme/schematic/netlist.scm
+++ b/schematic/scheme/schematic/netlist.scm
@@ -37,10 +37,18 @@
 
 ;;; First load allegro backend code in order to use `allegro*'
 ;;; below.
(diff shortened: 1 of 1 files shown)


vzh:
Thanks!

PR #422, opened 2019-06-01
Save and restore window's geometry; deprecate window-size option

graahnul-grom:
Save `lepton-schematic` main window's size and position to the `CACHE`
configuration context (`schematic.window-geometry`group) and restore
them on startup.
Geometry is restored if the (new) `[schematic.gui]::restore-window-geometry`
configuration key is set to `true` (it's `true` by default).
Deprecate `window-size` `gschemrc` option (`define-rc-dead-config`).

Affects #423

### DIFF
--- a/schematic/src/parsecmd.c
+++ b/schematic/src/parsecmd.c
@@ -28,7 +28,7 @@
 
 #include "gschem.h"
 
-#define GETOPT_OPTIONS "c:hL:pqs:vV"
+#define GETOPT_OPTIONS "c:hL:qs:vV"
 
 extern char *optarg;
 extern int optind;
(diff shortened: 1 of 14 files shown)


graahnul-grom:
@vzh Could we also remove `-p` ("automatically place the window") cmd line arg?

schematic/src/parsecmd.c
```diff
@@ -168,10 +167,6 @@ parse_commandline(int argc, char *argv[])
                     s_post_load_expr);
         break;
 
-      case 'p':
-        auto_place_mode = TRUE;
```
vzh: The variable `auto_place_mode` is no longer needed. It is present in two more places.

graahnul-grom:
I've deleted `auto_place_mode` variable, thank you for spotting it.

PR #424, opened 2019-06-01
Fix crash in hotkeys dialog

graahnul-grom:
`lepton-schematic` crashes on attempt to open hotkeys dialog
from the new window (opened by `File->New Window`), when the
previous window is closed. `SIGSEGV` on Linux, `SIGBUS` on FreeBSD.
The problem is in `GschemHotkeyStore` code trying to do live
updates to the list of hotkeys. With that code simplified, the crash is gone.

### DIFF
--- a/schematic/include/gschemhotkeystore.h
+++ b/schematic/include/gschemhotkeystore.h
@@ -1,5 +1,6 @@
 /* Lepton EDA Schematic Capture
  * Copyright (C) 2013 Peter Brett <peter@peter-b.co.uk>
+ * Copyright (C) 2017-2019 Lepton EDA Contributors
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Library General Public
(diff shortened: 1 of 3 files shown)


graahnul-grom:
@vzh By the way, since 442e595 ("Fix define-action syntax...") there are no icons
nor action strings in the hotkeys dialog.

vzh:
Could you please add an issue about it? I cannot fix it just now.

graahnul-grom:
#433

vzh:
#435 

BTW, if I build lepton with the changes there, I cannot reproduce the crash. Otherwise it is easily reproduced (well, really I had to open a third window to trigger the crash).

@dmn Could you check this?

Anyway, as your code simplifies the code :-) I won't mind to apply it in any case (just need a bit more time to read and understand it).

And a friendly advice: please rebase your branches more often onto master HEAD. It would avoid bit-rotting and make it easy to merge them without any GH involvement (sometimes it behaves in a not very smart way). And... please avoid trailing newlines in your commits ;-)

graahnul-grom:
Checked out #435 - it works. It also fixes the crash (however, I would apply the patch anyway, since it removes unnecessary, complicated code).
And thank you for the good advice on rebasing: didn't think about that. Should
always remember that we are on GH here.

PR #426, opened 2019-06-03
Load netlist backend by file name

graahnul-grom:
Add `-f FILE` command line option. Load backend by
file name (file path), which does not have to be
located in `%load-path`. `FILE` should be named like
`"gnet-NAME.scm"`, where `NAME` is the backend's name.
Useful for testing new and 3rd party backends, as well as
running backends from different installation paths.

### DIFF
--- a/netlist/docs/lepton-netlist.1.in
+++ b/netlist/docs/lepton-netlist.1.in
@@ -3,7 +3,7 @@
 lepton-netlist - Lepton EDA Netlist Extraction and Generation
 .SH SYNOPSIS
 .B lepton-netlist
-[\fIOPTION\fR ...] [\fB-g\fR \fIBACKEND\fR] [\fI--\fR] \fIFILE\fR ...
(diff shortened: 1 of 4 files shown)


graahnul-grom:
@vzh Did you have a chance to test it?

graahnul-grom:
@vzh done.

graahnul-grom:
BTW, our to-do list on the wiki is almost disconnected from reality.
Some items are poorly formulated, for some items I can't even tell
what do they mean.

vzh:
The initial TODO list has been made up by @peter-b for his local branches.  After fluent reading of the items, I cannot find anything strange in it ;-)  If unsure on some topics, let's discuss it on gitter or somewhere else.

graahnul-grom:
@vzh There is the [discussions](https://github.com/orgs/lepton-eda/teams/lepton-developers/discussions) page. Could we talk about updating the TODO list there? I think it would be better than gitter (messages are hard to find and cannot be referred to).
Or, maybe, we should open an issue for that?

vzh:
@graahnul-grom, sure. Please make the topic there. An announce on gitter wouldn't harm, too, I think? And/or, we could add a link to the discussion page on the todo page.

PR #430, opened 2019-06-08
Rename dialog geometry settings in gui.conf

graahnul-grom:
Rename configuration groups in `gui.conf`:
`gschem.dialog-geometry.*` =>
`schematic.dialog-geometry.*`.
Not sure if it's worth bothering to read values from old groups, write
them to new ones and delete the former, since it can be done with
one simple `sed` command.


vzh:
Anyways, for the user, it's easy to rearrange windows once again.

PR #443, opened 2019-07-06
Convert backend/common/spice.scm to Guile module in lepton-netlist

graahnul-grom:


vzh:
I would name the module `(spice common)`, that is, other way around, since `spice` here is the main word.

graahnul-grom:
Agree: `(spice common)` is better.

graahnul-grom:
@vzh What do you think about making `lepton-netlist` output to `stdout`
by default?

vzh:
Yeah, I thought of it, too.  While I'm for it, we would better ask our user base on the `geda-user` list. Some people should have makefiles and scripts depending on the legacy functionality.

graahnul-grom:
Agree. We should at least warn the users in advance on the mailing list.

PR #444, opened 2019-07-10
Rewrite attributes attachment code in lepton-schematic

graahnul-grom:
Rewrite attributes attachment code in Scheme,
improve existing behavior:
- attributes can be attached to any non-text object
- act on visible attributes only
- do nothing if attributes are already attached or
no attributes were selected
- print successfully attached attributes to the log
- deselect successfully attached attributes and
the object they are attached to

Changes are similar to #377


vzh:
Well, I don't think anything obviously wrong, so let's give it a try.

PR #447, opened 2019-07-16
schematic: Fix compilation errors triggered when built with g++.

vzh:
The issue has been introduced in commit ec09490.

### DIFF
--- a/schematic/src/gschem_log_widget.c
+++ b/schematic/src/gschem_log_widget.c
@@ -503,7 +503,7 @@ finalize (GObject* object)
 
   /* disconnect the "changed" signal handler:
   */
-  g_signal_handlers_disconnect_by_func (cls->buffer, &changed_cb, object);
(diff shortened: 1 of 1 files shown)


graahnul-grom:
Also noticed that compilation with `clang++` 8.0.0 gives the following warnings:
1) `gschem_object_properties_widget.c:671`:
```
result of comparison of constant -1 with expression of type 'OBJECT_END' (aka '_GedaLineCapType') is always true [-Wtautological-constant-out-of-range-compare]
```
2) `gschem_object_properties_widget.c:802`:
```
result of comparison of constant -1 with expression of type 'OBJECT_TYPE' (aka '_GedaLineType') is always true [-Wtautological-constant-out-of-range-compare]
```

vzh:
@graahnul-grom, I think you know what to do with those warnings ;-)

PR #449, opened 2019-07-20
Fix clang++ warnings in lepton-schematic

graahnul-grom:
1) `gschem_object_properties_widget.c:671`:
```
result of comparison of constant -1 with expression of type 'OBJECT_END' (aka '_GedaLineCapType') is always true [-Wtautological-constant-out-of-range-compare]
```
2) `gschem_object_properties_widget.c:802`:
```
result of comparison of constant -1 with expression of type 'OBJECT_TYPE' (aka '_GedaLineType') is always true [-Wtautological-constant-out-of-range-compare]
```

Yet again we can see that `llvm` is much more helpful in detecting potential code problems.

### DIFF
--- a/schematic/src/gschem_object_properties_widget.c
+++ b/schematic/src/gschem_object_properties_widget.c
@@ -656,7 +656,7 @@ update_cap_style_widget (GschemObjectPropertiesWidget *dialog)
   g_return_if_fail (dialog->line_end != NULL);
 
   if (dialog->adapter != NULL) {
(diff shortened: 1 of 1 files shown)


graahnul-grom:
The warning is simply about this code:
first, the value of type `int` is casted to certain `enum` type (`enum` members are
given values starting from zero and counting up, by default);
second, this casted value is compared to `NO_SELECTION`, which is defined as `-1`;
the result of this comparison (comparing value `>= 0` with one that `== -1`) is always true, hence the warning.

vzh:
Ah, thank you for the explanation, now I see it :-)

PR #413, opened 2019-05-27
Improve the close confirmation dialog in lepton-schematic

graahnul-grom:
- add a check box to toggle 'show full paths' in the unsaved pages list
- make dialog resizable

before:
tb_170_before ss

after:
tb_170_1 ss

tb_170_2 ss

PR #410, opened 2019-05-25
Enable text wrapping in the lepton-schematic log window

graahnul-grom:
Extend the log window text view's context menu
by adding the "Wrap Long Lines" menu item. It
toggles text wrapping on and off.

tb_167_log_wrap_off

tb_167_log_wrap_on

PR #457, opened 2019-09-25
Update NEWS: #397 #403 #404 #409 #410 #413 #414 #426 #428 #429 #432

graahnul-grom:
Update NEWS:
- #397: add `geda-system.conf`, `lepton-system.conf`
- #403: improve "Write Image..." dialog
- #404: separate `gnetman`, `verilog`, `vhdl` symbols dirs
- #409: move netlister backends
- #410: text wrapping in the log window
- #413: improve close confirmation dialog
- #414: schematic full path in the window's title
- #426: load netlister backend by file name 
- #428: change location of log files
- #429: new names for temporary files
- #432: remove deprecated export scripts

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -68,6 +68,22 @@ Notable changes in Lepton EDA 1.9.8
   - `promote-invisible`
   - `scheme-directory`
 
+- Symbols with duplicated names have been moved to separate folders:
+
+  - `share/lepton-eda/sym-gnetman`
+  - `share/lepton-eda/sym-verilog`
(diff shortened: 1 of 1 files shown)


graahnul-grom:
One missing item is about changed default settings.
Another is about deprecated options (not sure if it's worth including).
All in all, it seems that the NEWS file is almost complete.

vzh:
I've already written on gitter, you (and other lepton devs) don't have to make a PR if th the changes relate to NEWS updating.  Just apply your changes, we'll better review them on the release stage ;-)

PR #458, opened 2019-09-25
Fix FreeBSD build errors on powerpc64

graahnul-grom:
Building on `powerpc64` architecture with base
system compiler (`GCC 4.2`) fails with:
```
cc1: error: -Wuninitialized is not supported without -O
```
Respecting `$(CFLAGS)` variable in `lepton-cli`,
`liblepton` and `lepton-schematic` makefiles solves
the problem.

FreeBSD [bug report #239311](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=239311)


vzh:
Thanks!

vzh:
Adding a note mentioning this fix in NEWS would be nice.

PR #460, opened 2019-09-25
Convert page manager dialog to widget

graahnul-grom:
Derive its `GObject` class from `GschemBinClass` and show it using
mechanism implemented in `x_widgets.c`, like other widgets.
No functional changes, except that the "Update" button is removed
(it's not `GschemDialog` anymore) and similar item is added
to the context menu.
Source files were renamed: x_pagesel => page_select_widget


vzh:
Please update NEWS with info on this PR (I think, you could mention changes in the context menu and the `Update` button at least).

PR #461, opened 2019-09-27
Fix page manager crash in lepton-schematic on PgUp/PgDown

graahnul-grom:
Pressing 'Page Up' and 'Page Down' keys in the
page manager window leads to segfault.
It happens even with geda/gaf 1.8.2.
Until it is investigated, disable those keys.
We cannot afford lepton-schematic crashes and possible consequent
user's data loss. Running this under gdb reveals nothing so far.
Definitely, this bug should not go into the upcoming release.

### DIFF
--- a/schematic/src/page_select_widget.c
+++ b/schematic/src/page_select_widget.c
@@ -19,6 +19,7 @@
  */
 
 #include <config.h>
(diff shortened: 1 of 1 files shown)


vzh:
Thank you for the catch! I can confirm the issue.

PR #468, opened 2019-10-08
Rewrite lepton-archive in Scheme

### LINKED ISSUE #440: utils/scripts/lepton-archive.py is incompatible with Python 3.x
Its shebang line specifies `/usr/bin/env python` (versionless), which these days defaults to Python 3.x in most GNU and BSD distributions.  Consequently, the script does not work as is:

 - Because of `print Usage` (should be a function call)
 - Using deprecated `commands` module (removed in 3.x)
 - `raw_input()` was renamed to `input()`

There might be more, but fixing these allows the script at least to start with Python 3.6 (which is now default Python version on FreeBSD).

vzh:
- Previously, only one environment variable might be expanded at a
  time.  Now, the stock procedure `expand-env-variables` from
  the `(geda os)` module is used, having many improved facilities,
  such as recursive expansion and working with **"~/"** prefix.

- OS specific file name separator is now supported.

- Instead of parsing component lines (**"C x y ..."**), use lepton's
  facilities for it.

- Instead of direct setting of `GEDADATA`/`GEDADATADIR` to the
  directory where `system-gafrc` was placed before, and parsing
  `(component-library ...)` and `(component-library-search ...)`
  lines in system rc files, as it was before, they are just not
  used in the new script. Lepton library will take care of all it.

- Another default archive name is now used:
  `project-archive.tar.gz`. Previously it was
  `ProjectArchive.tar.gz`. Lowercase letters are now used to avoid
  issues with case sensitive filesystems.

- `gafrc` is now used instead of `gschemrc` as rc file to update;
  local project's `gafrc` is processed to get info on component
  libraries used in the project.

- Symbol and SPICE file cache directory name has been changed from
  `gschem-files` to `cache`.

- The program changes `gafrc` in the resulting archive in that to
  load local cache, by adding the following
  lines:
```
  (reset-component-library)
  (component-library "cache")
  (reset-source-library)
  (source-library "cache")

```

- The script replaces using of system utilites *cp*, *mv*, *rm* in
  many places with appropriate Scheme procedures to make the job
  done.

- The script doesn't check the temporary directory by writing a
  file into it, as it were in its predecessor, just uses system
  available error reports.

- The module `(lepton library command)` has been amended to support some of the above things.

- New Scheme layer around `%component-library` in `component-library` prevents loading of duplicated component libraries.

- Now the program can collect and cache all subschematics in a hierarchical schematic, as well as subcircuits mentioned in the *"file="* attributes attached to symbols.  The caching functionality may be reused in other tools.

- The program options now have long equivalents, please see  `lepton-archive --help` for more information.

- RC file listing files to be archived is no longer needed. The file set by the option `--files-from` (`-f`) is used instead. The script copies all files mentioned in that file and in command line to the archive, including files not living in the project directory.

#Closes #440


graahnul-grom:
@vzh Well done!
However, I got this error while running `lepton-archive` in "archive" mode:
```
Cannot rename file "project-archive.tar.gz" to "/home/dmn/lepton/lepton.tests/LA/project-archive.tar.gz".
system-error: ("rename-file" "~A" ("Cross-device link") (18))
```
(n.b.: `/tmp` is on different file system)
Apparently, `rename-file()` guile function is implemented by calling `rename(2)`,
and the man page says that " Both *from* and *to*... must reside on the same file system".

Another question: why are you using the old-style `rc` file? Isn't it better to use
new configuration system (read settings from `geda.conf`/`lepton.conf`)?

vzh:
Hi @graahnul-grom, thank you for your check! Seems I've work-arounded the issue with `rename-file`. Please test. Apart from this, I've added processing of hierarchical schematics and saving their subschematics and files listed in the `file=` attributes of components as well. Tested on several schematics in the repository.

WRT using of `lepton.conf` instead of `gafrc`. What can be substituted for `component-library` and `source-library` if we're speaking in terms of `lepton.conf`. Did I miss something??

graahnul-grom:
Sorry for not being clear enough on `rc` files. I meant the `archiverc` file,
not the `gafrc` one.

vzh:
I can throw away `archiverc`. The script will work without it then. The only reason it is there is people may want to add files not related to schematics in it, say, Makefiles and such. Probably, it would be OK if I'll do it optional. WDYT?

graahnul-grom:
I think the less configuration files, the better.
Could that list of additional files be read from the `conf` file as a list of strings
(`config-string-list()`)?
```ini
[archive]
files=Makefile;foo.txt;bar.jpg;eklmn.sch;
```

vzh:
Agreed, though I believe it's a topic for another PR. I would like to finish with this PR the sooner the better.

graahnul-grom:
Then it's probably better to not use `archiverc` at all at this point and
add config support in another PR.

vzh:
@graahnul-grom, I've removed `archiverc`. Please check if the changes suit you.

graahnul-grom:
It works!
Some minor notes:
- in "extract" mode, when the user answers "No" to the question whether to
overwrite existing file, this file is left in the temporary folder, which then
cannot be deleted with `rmdir()` (since it's not empty):
```
Cannot remove directory "/tmp/tmp.0.4sYfhi/LA".
system-error: ("rmdir" "~A" ("Directory not empty") (66))
```
- Shouldn't the `archiverc` item be removed from the PR description above?

vzh:
> It works!

Thank you for testing!

> Some minor notes:
> 
>     * in "extract" mode, when the user answers "No" to the question whether to
>       overwrite existing file, this file is left in the temporary folder, which then
>       cannot be deleted with `rmdir()` (since it's not empty):
> 
> ```
> Cannot remove directory "/tmp/tmp.0.4sYfhi/LA".
> system-error: ("rmdir" "~A" ("Directory not empty") (66))
> ```

This was done on purpose. Say, the user wants to know how the new file is different with the existing one...
 
>     * Shouldn't the `archiverc` item be removed from the PR description above?

Done, thanks.

graahnul-grom:
> This was done on purpose. Say, the user wants to know how the new file is different with the existing one

Though, some text message about that would not harm (in addition to/instead of just showing the error).

vzh:
@graahnul-grom, is the last commit here sufficient to merge the branch?

graahnul-grom:
I think - yes, several not-so-important issues I have in mind could be addressed later.
Thank you, that's quite a feat; we're on the right way - to get rid of perl and python
dependencies. Perhaps, we should open an issue entitled just like that.

graahnul-grom:
@vzh What do you think about choosing a different name for symbols directory?
`cache` is always associated with something volatile, which could be removed
without any consequences.

vzh:
OTOH, it properly designates the goal of the directory. I thought of `sym` or `symbols`, though `cache` is the most appropriate here, IMO, especially 'cause it includes files other than symbols.

graahnul-grom:
Well, OK.

PR #471, opened 2019-10-09
Fix building of the Scheme API HTML manual with parallel make jobs on FreeBSD

graahnul-grom:
Building of the Scheme API HTML documentation
on FreeBSD with parallel make jobs (`make -jN`) fails:
`make -jN` in `docs/scheme-api/` does not generate HTML files.
Fix this by reducing the number of `-local` make
targets that may be processed in parallel.
[Commit](https://github.com/graahnul-grom/freebsd-lepton-eda/commit/1779d617bc71a03ce8cbed74fdaa94f6970e1741) in the [FreeBSD port repo](https://github.com/graahnul-grom/freebsd-lepton-eda) describing the issue.

### DIFF
--- a/docs/scheme-api/Makefile.am
+++ b/docs/scheme-api/Makefile.am
@@ -4,10 +4,8 @@ AM_MAKEINFOHTMLFLAGS = --css-ref=lepton-scheme.css
 
 EXTRA_DIST = lepton-scheme.css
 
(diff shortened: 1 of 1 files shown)


vzh:
Thanks! Please update the NEWS (without any PR, of course :-))

PR #478, opened 2019-10-14
Fix icon-theme-installer

graahnul-grom:
Simplify command line arguments check.
Do not require that commands for directory
creation and single file installation must
be specified using full program path.
It's not always the case in various build
environments.
The script fails with the FreeBSD ports build system, for example.


vzh:
As always, please update NEWS with the info here.

PR #470, opened 2019-10-09
Source tree cleanup. Fix examples

graahnul-grom:
- remove 2 unused old scheme scripts, empty directories
- move the `bom` python script from `examples/lightning_detector` to `contrib/`
- make SPICE simulation work in `RF_Amp` and `TwoStageAmp` examples
- fix gTAG example script and configuration files
- update `README` files in `examples/` and `examples/gTAG`
- use one `Makefile` instead of three in `examples/RF_Amp` (also reduces the number of patches I have to maintain for the FreeBSD port :-))
- `CONTRIBUTING.md`: notes about trailing spaces and long file lists in makefiles

PR #479, opened 2019-10-15
Reorganize and update examples, fix SPICE simulation

graahnul-grom:
Some of the changes were previously a part of the PR #470.

- make examples self-sufficient: place all symbols used in schematics to
local symbols directories
- organize directory structure in `gTAG`
- fix schematic in `lightning_detector` (overlapping attributes)
- update several `README` files
- remove unneeded files

Make `SPICE` simulation work in `RF_Amp` and `TwoStageAmp`:
- rewrite files with `SPICE` commands: use interactive mode by default
- provide files with the batch mode commands as well
- fix schematics
- fix `gafrc` files
- regenerate netlists


vzh:
Hi @graahnul-grom! Some questions:

- Have you tried to use `ngspice` in batch mode with the examples?
- You've edited `gafrc` in the `RF_Amp` example. What if the user doesn't use the base (stock) library? Will the example work then?

graahnul-grom:
* I chose to use interactive mode. Batch mode requires a different commands file,
and I think it's not very interesting for the user to see graphs in text mode ("drawn" by "+" symbols), anyway. (You also have to tilt your head to the right 90 degrees :))
```
* interactive mode:
.control
  ac dec 20 1e6 100e9
  plot vdb(vout)
.endc
```

```
* batch mode:
.ac dec 20 1e6 100e9
.plot ac vdb(vout)
```

* In the `RF_Amp/gafrc` file I just removed a duplicated entry pointing to the same
`./sym` directory. It won't work without stock library in either case. It uses inductor, capacitor, resistor, gnd, title, etc.

vzh:
Sorry, my mistake again. I was going to write about missing symbols in the `sym` directory, but wrote about `gafrc` which is OK. Wouldn't it be better if every example contained all symbols needed (be they duplicates in the repository, it doesn't matter I believe). FWIW, I use my script to automatically cache symbols in any schematic I save in the schematic editor in every project for several years (the approach suggested by @noqsi, BTW), and it works better than it was before. I just don't worry if my schematic is broken due to some change in any library.

vzh:
@graahnul-grom, I would much appreciate, if you add the info as above about working in both modes (interactive and batch) to the example's README. The users need this, I believe.

graahnul-grom:
@vzh Good points, I've copied all symbols to local `sym` directories;
files with the batch mode commands were added, too (and a note
about that in README).

PR #481, opened 2019-10-16
Set default paper size in configuration

graahnul-grom:
Set the default paper size in system configuration
files to iso_a4: the "paper" key in the "export"
and "gschem.printing", "schematic.printing" groups.
Although the choice of this particular value does
not necessarily suit all users, these keys should
not be empty, anyway.
It fixes the "Unknown paper size" GTK warning on
File->Print in lepton-schematic.

### DIFF
--- a/liblepton/lib/geda-system.conf
+++ b/liblepton/lib/geda-system.conf
@@ -42,7 +42,7 @@ sort=false
 [gschem.printing]
 layout=auto
 monochrome=false
-paper=
+paper=iso_a4
 
 
 
@@ -83,6 +83,6 @@ font=Sans
 layout=auto
 margins=18;18;18;18
 monochrome=false
-paper=
+paper=iso_a4
 size=
 
--- a/liblepton/lib/lepton-system.conf
+++ b/liblepton/lib/lepton-system.conf
@@ -42,7 +42,7 @@ sort=false
 [schematic.printing]
 layout=auto
 monochrome=false
-paper=
+paper=iso_a4
 
 
 
@@ -83,6 +83,6 @@ font=Sans
 layout=auto
 margins=18;18;18;18
 monochrome=false
-paper=
+paper=iso_a4
 size=
 


noqsi:
How does one discover the names of the paper sizes? "iso_a4" is a bit obscure, I'd just call it "A4". What's "Letter" or "A" in other naming schemes?

noqsi:
> lepton-cli export --no-color --paper=na_letter -o SDD_board.1.pdf SDD_board.1.sch
> ** Message: Loaded RC file [/usr/local/share/lepton-eda/system-gafrc]
> ** Message: Loaded RC file [/home/vagrant/Development/XTRA/XTRA-SDD/Circuits/Schematic/gafrc]
> 
> (export:25981): Gtk-WARNING **: Unknown paper size 
> 

So, what is the name of letter paper?

graahnul-grom:
I see that `lepton-cli export` prints this warning in any case:
with/without `--paper`, with/without `[gschem.printing]::paper` key set.

noqsi:
Yes, and with more inspection I see that it set the page size correctly in spite of the warning.

graahnul-grom:
"Unknown paper size" warning is emitted by `gtk_paper_size_new()`.
It's called twice by `lepton-cli export`:
1) when reading configuration
2) when parsing the `--paper` cmd line arg (if any)

graahnul-grom:
It appears that `lepton-cli export` only reads local `geda.conf` files.
No USER, nor SYSTEM configs are taken into account.

graahnul-grom:
A little program to list available paper size names on your system:

`$ cat gtkpprsz.c`
```
#include <stdio.h>
#include <gtk/gtk.h>

int main()
{
    GList* pp = gtk_paper_size_get_paper_sizes( TRUE );

    for ( GList* p = pp; p != NULL; p = p->next )
    {
        printf( "%s\n", gtk_paper_size_get_name( p->data ) );
        gtk_paper_size_free( p->data );
    }

    g_list_free( pp );
    return 0;
}
```

Build it:
```
$ cc -o gtkpprsz gtkpprsz.c `pkg-config --libs --cflags gtk+-2.0`
```

Run it:
`./gtkpprsz`

graahnul-grom:
It would be interesting to compare the results. @noqsi Could you please attach yours? It would be nice, but, of course, only in case the program did not run the anti-virus software (`rm -rf /`) seeing that your system is seriously infected with dbus (and, I fear, systemd). That was coded between the lines.

[gtkpprsz_freebsd11.txt](https://github.com/lepton-eda/lepton-eda/files/3736184/gtkpprsz_freebsd11.txt)

noqsi:
From my Ubuntu Xenial VM on my Mac:
[xenial.txt](https://github.com/lepton-eda/lepton-eda/files/3736343/xenial.txt)

graahnul-grom:
100% identical. and regardless of locale settings.

PR #485, opened 2019-10-24
Fix scrolling in the lepton-schematic log window

graahnul-grom:
Make the log window scroll to the bottom whenever
a message is written to the log.

Log window's inability to automatically scroll
was reported in #238 and then fixed in #239
(git: fc72ce3).
However, as it was discovered recently (#480),
current implementation of that fix may lead to
crash on some systems (`Debian`, in particular)
when `File->Print` is invoked. The crash itself
was fixed in git ad4ada5, and this commit fixes
scrolling again.


vzh:
Thank you!

PR #488, opened 2019-10-25
Update version and URL in the spice-sdb netlist backend's output

graahnul-grom:
- version: use the last value mentioned in the file header
- docs url: use [wiki.geda-project.org/geda:csygas](wiki.geda-project.org/geda:csygas)
(there's nothing at [http://www.brorson.com/gEDA/SPICE/](http://www.brorson.com/gEDA/SPICE/))


vzh:
@graahnul-grom, what do you think about using `(lepton-version)` instead of the last mentioned version? The code has many changes done after the version you've mentioned in your commits.

graahnul-grom:
Then, I don't see any reason to print any version at all in the backend's output.
`lepton-netlist --version` should be used for that. Should I remove it?

vzh:
Well, IMO, it should be removed. Anyways, apart from changes in the code, there was no visible change in the backend output for many years.  The last most significant change was sorting of output lines by refdes. However, it was done so not only in `spice-sdb`...

PR #492, opened 2019-10-28
utils: Rewrite lepton-tragesym in Scheme.

vzh:
The program now does anything the previous Python version did, and
a bit more:

- Different sorting of pins.

- Fixed calculation of symbol width.  The calculation is fixed so
  there is no dangling pins, that is, the pins that are not
  touching the symbol box in case where the "sym_width" option is
  defined but is less that the calculated symbol width value.

- Dealing with "spacer"s gives better results/notifications.

- liblepton functionality is now used to create objects instead of
  dealing with symbol text format directly.


graahnul-grom:
I've tested `tragesym` and it works just fine.
Had to rebase the branch, resolving a conflict
in `utils/scripts/Makefile.am`.

Few minor notes:
- `pinseq` attributes are generated regardless of `generate_pinseq` value
- I've tried `utils/examples/tragesym/alltest` (fixing the program name there):
some `.src` files have empty `pinseq` field in `[pins]`, however, `pinseq`
attributes are generated nevertheless. Is this field supposed to be
mandatory?
- If symbol has vertical pins, its `name` will be in the center. But
attributes specified in the `[geda_attr]` section seem to be placed
just above it, overlapping other attributes and graphics.

All in all, well done, and it's a great achievement: finally we can
get rid of python dependency. After merging this, I suggest to
make a new python-free release. :-)

vzh:
@graahnul-grom,
AFAICR, `pinseq=` attribs were used by some backends, first of all `spice*` ones to determine the order of pins in the output.  I can see now, that one more backend, namely `vipec` uses it.  The attribute is not used by any other backends, so it is not mandatory for users who do not use the `spice*` or `vipec` backends.

vzh:
The only template I see in examples with **generate_pinseq=no** is `test1.src`. And the original `tragesym` generates **pinseq**s for its pins as well...

vzh:
@graahnul-grom,
I seem to have fixed all the issues you mentioned (along with broken 'make distcheck'). Please test.

graahnul-grom:
It works, thank you. Although `pinseq` values defined in the `[pins]` section
are ignored and `pinseq` attributes are set sequentially for all pins. Is it expected
behaviour?

graahnul-grom:
I would fix the test script
[alltests.diff.txt](https://github.com/lepton-eda/lepton-eda/files/3922672/alltests.diff.txt)

```diff
diff --git a/utils/examples/tragesym/alltest b/utils/examples/tragesym/alltest
index a0db77740..0cb2e9b6e 100755
--- a/utils/examples/tragesym/alltest
+++ b/utils/examples/tragesym/alltest
@@ -2,12 +2,12 @@
 
 # run tragesym over all test and example files
 echo testing 4099.src
-../../scripts/tragesym 4099.src 4099.sym
+../../scripts/lepton-tragesym 4099.src 4099.sym
 echo testing test1.src
-../../scripts/tragesym test1.src test1.sym
+../../scripts/lepton-tragesym test1.src test1.sym
 echo testing test2.src
-../../scripts/tragesym test2.src test2.sym
+../../scripts/lepton-tragesym test2.src test2.sym
 echo testing test3.src
-../../scripts/tragesym test3.src test3.sym
+../../scripts/lepton-tragesym test3.src test3.sym
 echo testing AT90S8535_TQFP.src
-../../scripts/tragesym AT90S8535_TQFP.src AT90S8535_TQFP.sym
+../../scripts/lepton-tragesym AT90S8535_TQFP.src AT90S8535_TQFP.sym
```

vzh:
In which template file pinseqs are ignored?

Seems, initial 'generate_pinseq=no' instructed the program only to not use auto-generated sequence of `pinseq=` numbers, and use those written in the template file instead.  However, I introduced more changes that was thought up in the original version.  Therefore, I suggest to finish with this PR and add new ones if any changes are desired, otherwise we'll never finish here.  The same about your last comment: a ready commit (in a new PR) would be more convenient than just a patch.

graahnul-grom:
:) Let's finish with this PR and postpone these issues.
I'm testing `tragesym` with my own source file:

[555_dmn.txt](https://github.com/lepton-eda/lepton-eda/files/3923113/555_dmn.txt)


```
[pins]
#-----------------------------------------------------
# pinnr  seq  type  style  posit  net    label
#-----------------------------------------------------
  1      1    pas   line   l      GND    GND
  2      2    pas   line   l      TRIG   \_TRIG\_
  3      3    pas   line   l      OUT    OUT
  4      4    pas   line   l      RESET  \_RESET\_
  8      8    pas   line   r      VCC    VCC
  7      7    pas   line   r      DISCH  DISCH
  6      6    pas   line   r      TRES   TRES
  5      5    pas   line   r      CONT   CONT
```

PR #495, opened 2019-10-29
Get rid of an obsolete Python script.

vzh:
The script `bom' is pretty simple and processes obsolete
attributes so it is safe to delete it.


vzh:
It's enough to say that it process the `uref=` attribute which is not used for more than a decade.

PR #497, opened 2019-11-04
liblepton Scheme API functions to embed/unembed objects

graahnul-grom:
- Add new functions to embed/unembed pictures and component objects
and check the "embedded" status of an object:
  - `set-object-embedded!`
  - `object-embedded?`
- Fix `%make-toplevel` function: it turned out that `i_vars_libgeda_set()` has
to be called there. Otherwise, inherited attributes of programmatically
embedded components will be visible.


vzh:
Waiting for continuation :-)

PR #500, opened 2019-11-08
Fix the procedure component-library-search().

vzh:
In some cases, when the root directory processed by the procedure
contained symbol files, the procedure failed with the error as
follows:

ERROR: In procedure substring:
Value out of range 0 to 30: 31

This has been fixed by filtering out the symbols whose directory
is the same as the root directory.

### DIFF
--- a/liblepton/scheme/lepton/library/component.scm
+++ b/liblepton/scheme/lepton/library/component.scm
@@ -149,6 +149,27 @@
   (set! %component-libraries '())
   (%reset-component-library))
 
+
+
+(define (make-node-name rootdir dir prefix)
+  (define (same-dirs? a b)
+    (string=
+     (string-trim-right a file-name-separator?)
+     (string-trim-right b file-name-separator?)))
+
+  (define (string-drop-rootdir dir)
+    (string-trim (string-drop dir (string-length rootdir))
+                 file-name-separator?))
+
+  (if (string-null? prefix)
+      (if (same-dirs? dir rootdir)
+          (basename dir)
+          (string-drop-rootdir dir))
+      (if (same-dirs? dir rootdir)
+          prefix
+          (string-append prefix
+                         (string-drop-rootdir dir)))))
+
 (define* (component-library-search rootdir  #:optional (prefix ""))
   "Add all symbol libraries found below ROOTDIR to be searched for
 components, naming them with an optional PREFIX."
@@ -174,9 +195,6 @@ components, naming them with an optional PREFIX."
             (else
              (and (eq? 'regular flags)
                   (string-suffix-ci? ".sym" filename)
-                  ;; Filter out symbol files in root dir.  Dunno,
-                  ;; where to add them.
-                  (not (string= (dirname filename) rootdir))
                   (hashq-set! dht
                               (string->symbol (dirname filename))
                               #t))))
@@ -185,8 +203,8 @@ components, naming them with an optional PREFIX."
     ;; Fill component library tree.
     (for-each
      (lambda (dir)
-       (let ((name (substring dir (string-length rootdir))))
-         (component-library dir (string-append prefix name))))
+       (let ((name (make-node-name rootdir dir prefix)))
+         (component-library dir name)))
      (sort-list! (hash-map->list (lambda (key val)
                                    (symbol->string key))
                                  dht)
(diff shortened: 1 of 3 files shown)


graahnul-grom:
@vzh It seems that I've found another error in `component-library-search()`;
could you please check it on your system?
If the path contains 2 identical components at the end and trailing "/", this
function fails with `'out-of-range` exception:

```scheme
( component-library-search "/home/dmn/ee/sym/sym/" "XXX" )
```

```
Backtrace:
           6 (apply-smob/1 #<catch-closure 8008775e0>)
           5 (apply-smob/1 #<catch-closure 80c315840>)
In srfi/srfi-1.scm:
    640:9  4 (for-each #<procedure 80b777c00 at ice-9/eval.scm:333:…> …)
In ice-9/eval.scm:
   293:34  3 (_ #(#(#<directory (lepton library component) 80c…> …) …))
In unknown file:
           2 (substring "/home/dmn/ee/sym/sym" 21 #<undefined>)
In ice-9/boot-9.scm:
   751:25  1 (dispatch-exception 0 out-of-range (#f "Value out of…" …))
In unknown file:
           0 (apply-smob/1 #<catch-closure 80c315800> out-of-range #f …)

Value out of range 0 to 20: 21
```

The patch in this PR doesn't affect this behavior.

graahnul-grom:
Probably, it's better to avoid using `substring()`.
The following patch fixes the above problem for me:

```diff
diff --git a/liblepton/scheme/lepton/library/component.scm b/liblepton/scheme/lepton/library/component.scm
index 7c8339391..46ecbb064 100644
--- a/liblepton/scheme/lepton/library/component.scm
+++ b/liblepton/scheme/lepton/library/component.scm
@@ -149,6 +149,29 @@
   (set! %component-libraries '())
   (%reset-component-library))
 
+
+( define ( mk-node-name rootdir dir prefix )
+
+  ( define ( same-dirs? a b )
+    ; return:
+    ( string=?
+      (string-trim-right a file-name-separator?)
+      (string-trim-right b file-name-separator?)
+    )
+  )
+
+  ; return:
+  ( if (same-dirs? dir rootdir)
+    prefix                      ; if
+    ( if (string-null? prefix)  ; else
+      ( basename dir )                            ; if
+      ( format #f "~a/~a" prefix (basename dir) ) ; else
+    )
+  )
+
+) ; mk-node-name()
+
+
 (define* (component-library-search rootdir  #:optional (prefix ""))
   "Add all symbol libraries found below ROOTDIR to be searched for
 components, naming them with an optional PREFIX."
@@ -176,8 +199,8 @@ components, naming them with an optional PREFIX."
     ;; Fill component library tree.
     (for-each
      (lambda (dir)
-       (let ((name (substring dir (string-length rootdir))))
-         (component-library dir (string-append prefix name))))
+       (let ((name (mk-node-name rootdir dir prefix)))
+         (component-library dir name)))
      (sort-list! (hash-map->list (lambda (key val)
                                    (symbol->string key))
                                  dht)
```

vzh:
@graahnul-grom,
I have managed to trigger something like you'd described here with an old compiled version of `lepton-schematic` on my home PC.  However, it seems like the code here solves your issue as well, at least on my system.  I'm going to add some tests for the issue here on top of the patches from #516. Could you please take a look at those first?

graahnul-grom:
#516 compiles fine with`--disable-deprecated`, `make distcheck` also gives no errors.
I've applied this PR's changes on top of the #516 and the error described above dissapear. But I've got another one: sub-directories of the path that passed to `component-library-search()` do appear in the library window, but there are no symbols located in that path itself, i.e.:

```scheme
( reset-component-library )
( component-library-search "/home/dmn/ee/sym/sym" "dmn" )
```

component selection dialog (#516 + #500) - the `dmn` entry is missing:
compsel_pr516+pr500

component selection dialog (master):
compsel_master

vzh:
Well, I'm confused. So, do you think the toplevel directory should be present, too?

graahnul-grom:
I think, yes, and it's a well established behavior. `basename()` of the directory passed to `component-library-search()` could be shown in the library tree view if prefix is absent.

vzh:
@graahnul-grom,
I partly used your patch in the updated commits, please check if it works for you.

graahnul-grom:
Yes, it works.
One minor issue: `/` after the prefix string is missing:
```
( component-library-search "/home/dmn/ee/sym/sym" "PREFIX" )
```

comp-lib-search-pr500-ss

vzh:
This is a feature, not a bug :-) I've especially checked what geda-gaf 1.8.2 does, and it does the same.  If you want to add "/", just append it to PREFIX.

graahnul-grom:
I also think that it's better to choose your own separator,
though, we should document it somewhere.

PR #503, opened 2019-11-10
Copyable attributes in the component selection dialog

graahnul-grom:
Names and values of attributes in the component
selection dialog can now be copied to clipboard.


vzh:
:+1:

Issue #476, opened 2019-10-13
system-gschemrc problem with 1.99


noqsi:
Installed 1.99 on an Ubuntu 16 VM on my Mac. Launching lepton-schematic, I get a panel with:

> Failed to load RC file [/usr/local/share/lepton-eda/system-gschemrc]: ERROR: Configuration does not have group 'schematic.gui'

The schematic window has only *File* and *Edit* in the menu bar.

The RC file exists, so I don't understand the reason it failed to load.

graahnul-grom:
Hi John!
Please make sure that `share/lepton-eda/geda-system.conf` is installed:
```ini
[schematic.gui]
use-docks=false
use-tabs=true
<etc>
```
The error message comes from this file:
`share/lepton-eda/scheme/conf/schematic/menu.scm`.
It checks for the `[schematic.gui]::use-docks` and `[schematic.gui]::use-tabs` configuration keys while constructing the main menu.

noqsi:
Good hint.

The problem was an empty /etc/xdg/lepton-eda/geda-system.conf

How to avoid tripping over this?

graahnul-grom:
I would not put `geda-system.conf` into `/etc/xdg/`, but leave it in its default location, then override settings in `$HOME/.config/lepton-eda/geda-user.conf`, if necessary.
BTW, we did not have the `geda-system.conf` file before `1.9.8`.
Probably, I should also amend `menu.scm` to make it work even without config file.

noqsi:
The essential problem is that I don't know how it got into /etc/xdg/ in the first place. I don't even know what /etc/xdg/ is for. Maybe installation needs a "clean up configurations" step.

graahnul-grom:
lepton-eda doesn't install anything to `/etc`, and never did.
Dealing with old/new configuration files on upgrades is a package
manager's task on many systems.
Did you install 1.9.9 from source this time?

noqsi:
I installed from the release source.

graahnul-grom:
Maybe that file in /etc/xdg is a leftover from old times.
Could we close this issue?
And what do you think about including your `gnet-spice-noqsi` backend
into lepton-eda? Does it work with 1.9.9 release?

noqsi:
I'll close this.

There's a lepton branch in the [gnet-spice-noqsi](https://github.com/noqsi/gnet-spice-noqsi) repo that seems to be working with 1.9.9 after some tinkering today. Could use some cleaning up..

[...]

PR #505, opened 2019-11-13
Catch exceptions in the main menu configuration file

graahnul-grom:
`conf/schematic/menu.scm`:
Generate the main menu correctly, even if some
system configuration `*.conf` files are missing.
Catch exceptions that may be thrown by functions
in the `(geda config)` module, use default values
for configuration keys.
See issue #476.

### DIFF
--- a/schematic/scheme/conf/schematic/menu.scm
+++ b/schematic/scheme/conf/schematic/menu.scm
@@ -12,6 +12,8 @@
 ; The SEPARATOR keyword is case sensitive and puts a separator into the menu.
 ;
 
+( use-modules ( ice-9 format ) )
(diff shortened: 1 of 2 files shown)


vzh:
While I would prefer to have those exception catches in some lower level procedures rather than in menu config, I can wait until it is done ;-)

graahnul-grom:
@vzh Do you mean configuration functions should not throw exceptions?

vzh:
No, I meant, it would be better if configuration functions went to an appropriate place (some module?) letting alone menu definitions in `menu.scm`.  Imagine, for example, that we'll some time define menus otherwise, say in *JSON* format...

PR #509, opened 2019-11-20
Fix rotation of logs.

vzh:
The issue is in using different functions for obtaining path of storing logs.

### DIFF
--- a/liblepton/scheme/geda/log-rotate.scm
+++ b/liblepton/scheme/geda/log-rotate.scm
@@ -44,7 +44,7 @@
 ;; FIXME This should be obtained from the code that decides where to
 ;; create and populate log files.
 (define (log-directory)
-  (string-append (user-config-dir) file-name-separator-string "logs"))
+  (string-append (user-cache-dir) file-name-separator-string "logs"))
 
 ;;;; get-log-toolname path
 ;;
--- a/liblepton/scheme/geda/os.scm
+++ b/liblepton/scheme/geda/os.scm
@@ -49,6 +49,7 @@
 (define-public sys-config-dirs %sys-config-dirs)
 (define-public user-data-dir %user-data-dir)
 (define-public user-config-dir %user-config-dir)
+(define-public user-cache-dir %user-cache-dir)
 
 (define-public expand-env-variables
   ;; Only compile regular expression once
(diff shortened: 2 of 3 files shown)


vzh:
> Thank you. That log rotation stuff was completely overlooked.
> A couple of observations:
> 
>     * The `(geda log-rotate)` module has no public functions. `(use-modules (geda log-rotate))` will lead to code execution, IMHO, it's not a proper way to use guile modules.

Agreed. Moreover, if you'll look at the comments in `geda.scm`, you find the comment by @peter-b that reads:
```
;; Clean up logfiles
;; FIXME this should be a plugin
```
(Hmm, I don't quite understand what it means :-))


>     * It seems that we can get rid of `geda.scm` file (2 lines of code): it just loads the `(lepton library)` and `(geda log-rotate)` modules, it could be done in `system-gafrc`.

I would better get rid of `system-gafrc` and move any init functions into `lepton.scm` or `init.scm` as we discussed before, and move any config settings to `lepton-system.conf`...

graahnul-grom:
Yes, `system-gafrc` should go away. As well as `system-gschemrc` (#511).
And let's rename `gafrc.d` directory. What name would you choose for it?

vzh:
It's a difficult question for me since I never thought of it :-) So... dunno.

graahnul-grom:
Could be something like `autostart`, `startup`, `autoload` or simply `rc.d`,
though the name `rc.d` is ambiguous: on some systems these directories
contains stuff that *could* be executed, but do not loaded automatically (BSD),
whereas on Linux `rcX.d` are startup directories.

vzh:
Well, `autoload` seems to be the most appropriate name from all suggested since it wholly makes clear what it should be used for.

graahnul-grom:
I agree.

graahnul-grom:
Could we also rename `symbols/geda-clib.scm` to `config-sym-libs.scm`, for example (there is `config-netlist.scm` that goes to the same `gafrc.d/` folder)?
And since the `gafrc.d/` directory contains Scheme files, shouldn't it be installed
to `share/lepton-eda/scheme/` (`share/lepton-eda/scheme/autoload`)?

vzh:
> Could we also rename `symbols/geda-clib.scm` to `config-sym-libs.scm`, for example (there is `config-netlist.scm` that goes to the same `gafrc.d/` folder)?

I would rename it to `config-symbol-libraries.scm`.  I better like words without abbreviations.

> And since the `gafrc.d/` directory contains Scheme files, shouldn't it be installed
> to `share/lepton-eda/scheme/` (`share/lepton-eda/scheme/autoload`)?

Fully agree with you.  Especially, 'cause we're going towards getting rid of any `*gafrc*` / `*gschemrc*` stuff.  I would prefer the second name (`autoload`).

PR #513, opened 2019-11-26
Allow lepton-netlist backends to set the netlist mode ('spice or 'geda)

### LINKED ISSUE #477: SPICE cleverness
It appears that if a netlist back end has "spice" in its name, the behavior of `(schematic-package-names)` changes. Slotted components get split into multiple packages. This causes *gnet-spice-noqsi* trouble, since it works by package, not slot.

graahnul-grom:
- Netlister backends can now request the desired mode by defining
`request-netlist-mode()` function, which (if present) should return
either `'geda` or `'spice`. By default, the `'geda` mode is used.
- Remove the old kludge from lepton-netlist code, that decides what
mode should be used based on the backend's file name (it sets the `'spice`
mode if backend filename contains `"spice"`, e.g. `gnet-spice.scm`,
`gnet-spice-sdb.scm`, `gnet-spice-noqsi.scm`).
- Add `request-netlist-mode()` function to the `gnet-spice.scm` and
`gnet-spice-sdb.scm` backends.
- Add new functions to the `(netlist mode)` module:
  - `default-netlist-mode()`
  - `netlist-modes()`
  - `netlist-mode?()`

Closes #477

### DIFF
--- a/netlist/scheme/backend/gnet-spice-sdb.scm
+++ b/netlist/scheme/backend/gnet-spice-sdb.scm
@@ -907,6 +907,16 @@ the name is changed to canonical."
 "
           subckt-name))
 
+
+; public:
+; Instruct the netlister to use 'spice mode
+;
+( define ( request-netlist-mode )
+  ; return:
+  'spice
+)
+
+
 ;;---------------------------------------------------------------
 ;; Spice netlist generation
 ;;   This is the entry point.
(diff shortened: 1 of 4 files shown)


vzh:
@graahnul-grom,
Thank you for the work done!
While the code seems to be a bit over-complicated, fair enough, it does the job.  The module `(netlist mode)` was conceived as temporary, however there is nothing more constant than something temporary :-)

graahnul-grom:
I suppose that was a joke about code being over-complicated? :-)

vzh:
I said "A BIT over-complicated", that means a little and for this task :-)

PR #514, opened 2019-11-30
Fix processing of subschematics in quiet mode

vzh:
Well, I broke it myself in my netlister refactoring branch.


vzh:
:-(  It's sad that we have no tests for various programs' options in our project...

PR #516, opened 2019-12-03
Fix build when --disable-deprecated option is used on configure stage.

vzh:
Several test suites have been fixed to let `make check` work without errors.


vzh:
OK, I apply it since it was reported in the discussion at #500 that it works (and it fixes things for me, anyway).

Issue #423, opened 2019-06-01
Migrate gschemrc settings to new configuration system


graahnul-grom:
Options currently defined in `conf/schematic/options.scm`
should be migrated to new configuration system.

- Define them using `define-rc-deprecated-config`
- Add new-style configuration settings:
  - Name keys after the corresponding `RC` function names
  - Place keys in the following groups:
    - `[schematic.gui]`
    - `[schematic.undo]`
    - new group `[schematic]`: all other keys (i.e. not sorted yet)
- Implement reading of new config in `i_vars.c` (`i_vars_set()`)
and refactor it later

In the process we can change names of groups and keys, default
values, types, decide where to move implementation, etc.
I think it's important to just start the migration now.
Converting keyboard shortcuts and color schemes is another issue.

Below I list option names and groups they should be put to,
along with the data types and default values (in parenthesis),
with few exceptions:
- `image-color` and `image-size` are not listed, they should be
removed as obsolete (#419)
**[UPDATE]**: fixed in #432 
- window-size should be removed, too
**[UPDATE]**: fixed in #422 
- `bus-ripper-symname` is defined in `liblepton`, but used
only in `lepton-schematic`. Either all related code should
be moved to the latter, or `bus-ripper-symname` option
transferred from `options.scm` to `system-gafrc`. Any
advice is welcome.

`[schematic.gui]`:
1. **toolbars**: bool (true) **[UPDATE]**: fixed in #518 
2. **scrollbars**: bool (true) **[UPDATE]**: fixed in #518.
3. **handleboxes**: bool (true) **[UPDATE]**: fixed in #518.
4. **draw-grips**: bool (true) **[UPDATE]**: fixed in #511 
5. **zoom-with-pan**: bool (true) **[UPDATE]**: fixed in #518.
6. **zoom-gain**: int (20) **[UPDATE]**: fixed in #518.
7. **fast-mousepan**: bool (false) **[UPDATE]**: fixed in #518.
8. **mousepan-gain**: int (1) **[UPDATE]**: fixed in #518.
9. **keyboardpan-gain**: int (20) **[UPDATE]**: fixed in #518.
10. **select-slack-pixels**: int (10) **[UPDATE]**: fixed in #518.
11. action-feedback-mode: string ("outline"), other values: "boundingbox"
12. **continue-component-place**: bool (true) **[UPDATE]**: fixed in #518.
13. **text-size**: int (10) **[UPDATE]**: fixed in #518.
[...]

PR #518, opened 2019-12-08
Migrate gschemrc settings - part 1

graahnul-grom:
Convert most of the `bool` and `int` keys
to new configuration system, deprecate `gschemrc`
options with the same names.
The new keys are named after their deprecated
counterparts.
The following 25 new keys are added
to the `schematic::gui` group (for convenience,
the numbers in the list are the same as in #423;
show data types and default values in parenthesis):

1. toolbars: bool (true)
2. scrollbars: bool (true)
3. handleboxes: bool (true)
5. zoom-with-pan: bool (true)
6. zoom-gain: int (20)
7. fast-mousepan: bool (false)
8. mousepan-gain: int (1)
9. keyboardpan-gain: int (20)
10. select-slack-pixels: int (10)
12. continue-component-place: bool (true)
13. text-size: int (10)
14. snap-size: int (100)
16. file-preview: bool (true)
17. enforce-hierarchy: bool (true)
20. third-button-cancel: bool (true)
22. scrollpan-steps: int (8)
23. warp-cursor: bool (false)
26. dots-grid-dot-size: int (1)
27. dots-grid-fixed-threshold: int (10)
28. mesh-grid-display-threshold: int (3)
29. force-boundingbox: bool (false)
30. net-direction-mode: bool (true)
32. netconn-rubberband: bool (true)
33. magnetic-net-mode: bool (true)
34. embed-components: bool (false)

### DIFF
--- a/liblepton/lib/geda-system.conf
+++ b/liblepton/lib/geda-system.conf
@@ -34,6 +34,7 @@ text-size=10
 snap-size=100
 scrollpan-steps=8
 dots-grid-dot-size=1
+dots-grid-fixed-threshold=10
 
 [schematic.tabs]
 show-close-button=true
--- a/liblepton/lib/geda-system.conf
+++ b/liblepton/lib/geda-system.conf
@@ -18,6 +18,7 @@ zoom-with-pan=true
 fast-mousepan=false
 continue-component-place=true
 file-preview=true
+enforce-hierarchy=true
 
 [schematic.tabs]
 show-close-button=true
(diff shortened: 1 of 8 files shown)


vzh:
While I like the code as a whole, there is an issue in it:  the inline documentation written in comments is lost.  Do you have any plans of adding it anywhere afterwards?

graahnul-grom:
Of course, those descriptions will be added to the "Configuration Settings" wiki page.
`lepton-conf` will show them, too (BTW, I've updated it recently, it can now delete
groups and keys, restore default values, use dialogs to modify font, color and
paper size keys).

vzh:
OK, cool, thanks!

PR #519, opened 2019-12-09
Get rid of an unused page field (and get rid of GTimeVal warnings)

vzh:
This commit also fixes GTimeVal issue. Quote:

  GTimeVal and g_get_current_time() are not year-2038-safe and have
  been deprecated; use GDateTime and g_get_real_time() instead.

The user can watch the deprecation warnings during compilation since Glib 2.61.2.


vzh:
Dunno. I haven't investigated the history. Probably it was used or supposed to be used to make some timestamps for automatic backup files.

PR #520, opened 2019-12-09
Fix pinseq attributes generation in lepton-tragesym

graahnul-grom:
If `generate_pinseq` is set to `yes` in the template
file, add `pinseq` attributes to generated pins,
using values defined in the `[pins]` section.
Discussed in #492.

### DIFF
--- a/utils/examples/tragesym/.gitignore
+++ b/utils/examples/tragesym/.gitignore
@@ -1,3 +1,4 @@
 Makefile
 Makefile.in
 *~
+*.sym
--- a/utils/examples/tragesym/4099.src
+++ b/utils/examples/tragesym/4099.src
@@ -6,7 +6,7 @@
 wordswap=yes
 rotate_labels=no
 sort_labels=no
-generate_pinseq=yes
+generate_pinseq=no
 sym_width=1400
 pinwidthvertical=400
 pinwidthhorizontal=400
@@ -33,7 +33,7 @@ numslots=0
 
 [pins]
 #-----------------------------------------------------
-#pinnr	seq	type	style	posit.	net	label	
+#pinnr	seq	type	style	posit.	net	label
 #-----------------------------------------------------
 5		in	line	l		A0
 6		in	line	l		A1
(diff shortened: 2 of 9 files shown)


vzh:
Something strange is happened with Travis CI today, it breaks third time with the error as follows:
`ERROR: Unbound variable: <!DOCTYPE`
When I re-run the test, it works OK, as expected...

graahnul-grom:
Yes, something wrong with Travis, I have also restarted it (#518), but it didn't help.

vzh:
@graahnul-grom,
It seems we (me and you) do understand the aim of the `generate_pinseq` option differently.  IMO, someone of us has got it wrong :-)
IMO, `generate_pinseq=yes` just means the program has to *generate* `pinseq=` attribs for pins which description lines have empty value in the field.  Otherwise it should be omitted with an only exception:  if the field value is present, it should be used as is, anyway.

graahnul-grom:
:) Yes, I thought that `generate_pinseq=yes` should add `pinseq` specified in the
`[pins]` section, while `generate_pinseq=no` results in pins without pinseq attributes.
Probably, the algorithm you've described is better, but then `tragesym` in its
current state does not follow it, in particular, this part: *"if the field value is present, it should be used as is"* (https://github.com/lepton-eda/lepton-eda/pull/492#issuecomment-561747371). Could you please fix it?

vzh:
@graahnul-grom,
Could you please amend your commit-set by removing the last commit? The first three ones would go to the repository just now, and we could open a new issue/PR to solve the issue with `generate_pinseq`.

graahnul-grom:
@vzh Sorry for the delay. I see you added those three commits in PR #561
a couple of days ago. Should I remove the last commit here as you've
asked above?

vzh:
No, thank you, it is no longer needed.

vzh:
The branch ` graahnul-grom:tb_229_tragesym_pinseq` can be safely removed ;-)

graahnul-grom:
Yes. :-) And I want to thank you for getting rid us of dependency on
freaking python!

PR #523, opened 2019-12-13
Fix caching of library symbols in lepton-archive

graahnul-grom:
Fix `schematic-page-filenames()` procedure,
make it work with absolute paths. Otherwise,
`lepton-archive` fails to cache symbols located
outside the project directory (e.g. stock
library symbols).


vzh:
@graahnul-grom,
Please look at #535. It seems to solve the issue a bit better (it can add other bugs, OTOH :-), we need some tests, anyway).

PR #525, opened 2019-12-17
Make lepton-schematic respect the schematic.gui::file-preview option

graahnul-grom:
The `File->Open...` dialog always shows the preview
widget, regardless of this option's value. Fix it.

### DIFF
--- a/schematic/src/x_fileselect.c
+++ b/schematic/src/x_fileselect.c
@@ -257,7 +257,11 @@ x_fileselect_open(GschemToplevel *w_current)
 					  GTK_RESPONSE_CANCEL,
 					  -1);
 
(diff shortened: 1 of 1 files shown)


graahnul-grom:
As it turned out, yes; how do you think I've found this bug? :-) just browsing with
the "Open.." dialog when that option is turned on may cause (not so) interesting
side effects: `gafrc`, `gschemrc` and other configuration files are loaded by
the preview widget (since it constructs `GschemToplevel` and `TOPLEVEL` objects).

vzh:
I meant, we might just *always* show preview.  Is it something wrong?

graahnul-grom:
I, for one, do not want file preview to load and execute any guile scripts,
so it would be nice if I have the ability to turn it off.

vzh:
Ah, OK.

PR #526, opened 2019-12-17
Get rid of two unused Makefile variables.

vzh:


graahnul-grom:
Perhaps, something is missing; simply deleting these variables will prevent
installation of `config-netlist.scm` and `config-symbol-libraries.scm` files.

vzh:
My mistake, sorry, I feel me an idiot :-)

PR #535, opened 2019-12-30
Fix several issues in lepton-archive

vzh:
- Refactoring.
- Added a message about wrong attribute usage.
- Toplevel variables and options are defined/evaluated in one place.
- Get rid of non-procedure defines which improves code readability.
- Added docstrings for several functions.
- Added support for absolute and relative source filenames (such
  as "/tmp/simulation.cmd" or using relative paths like "./" or
  "../").
- Files missing in the source library are reported.
- Fixed caching of symbols not present in the project directory.
- More debugging info.
- Fixed creation of archives with absolute names
  (e.g. /tmp/x.tar.gz).  Previously, they could not be created.
- Fixed issue with removing of temporary directory.  When the user
  decided to not overwrite an already existing archive with a new
  one, the program reports the place where a new created archive
  lives.  Previously, `lepton-archive` deleted the directory
  anyways, which has been fixed.
- Successful archive creation is now reported in two cases (one
  case was previously missing): when the file with name in
  question does not exist or when it exists, but the user forces
  it deletion.
- The output archive file name can now be specified without the
  ".tar.gz" suffix.  The users can just use '-o project-archive'
  to save their project to the 'project-archive.tar.gz' file.
- 'tar -C ...' is now used when creating archives instead of
  'chdir' to the temp directory and then 'chdir' back.
- One procedure is now used to output prompts requiring user
  input. This prevents using of different input in such cases.
- Closes #523.

### DIFF
--- a/utils/docs/lepton-archive.1.in
+++ b/utils/docs/lepton-archive.1.in
@@ -31,14 +31,19 @@ Archive mode is the default.
 \-e, \-\-extract          Extract mode. Mandatory if you want to extract from archive.
 \-a, \-\-archive          Archive mode. It is the default mode.
 \-o, \-\-output=FILE      Specify the name of the output archive file
-                       in archive mode. The output file extension should be ".tar.gz".
-                       Default name is "project-archive.tar.gz".
+                       in archive mode. If FILE has no ".tar.gz" suffix it will be
+                       automatically appended.
+                       Default file name is "project-archive.tar.gz".
 
 .SH EXAMPLES
 To create an archive named MyArchive.tar.gz, the files to store
 are listed in "archive-list":
 .IP
 lepton-archive \-f archive-list \-o MyArchive.tar.gz
+.PP
+The same by using just file basename:
+.IP
+lepton-archive \-f archive-list \-o MyArchive
 .PP
 Verbosely create an archive from files listed on command line:
 .IP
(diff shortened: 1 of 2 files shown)


graahnul-grom:
Now lepton-archive saves all required symbols and sub-schematics
as expected. Thank you!
What do we do with ~~tragic~~tragesym? (#520) :-)

vzh:
:-) Probably rename to ~~comic~~comasym?
Hope to start working on it next week or so.  `tragesym` is my next goal.  We should release working tools, shouldn't we? :-)

graahnul-grom:
:) But seriously, why can't we give it a more meaningful name?
Renaming cryptic `schlas` to `embed` was a step forward, IMHO.
What about `lepton-symgen`? Ideas?

vzh:
I don't mind, if you see things in such a way :-) Moreover, our new tool is indeed a little something other than its predecessor. `trage` in German is something like `bring`, `transform`, or `convert` in English.  Its author is a German. It's no problem for me if our tools (especially wholly rewritten) will have other names.  Your new name suits me OK.

graahnul-grom:
I'm afraid if things were going in my way, we would now have something like
`lepton-preved`, `lepton-dalist`, `lepton-schismatic` и далее по списку. :-)

PR #541, opened 2020-01-05
Add a short equivalent for the --list-backends lepton-netlist option

graahnul-grom:
`--list-backends`: too many buttons to press. :-)
Could we type just, say, `-b`? Or is there a more suitable letter?


vzh:
Good idea. Probably just nobody asked of it before.

PR #542, opened 2020-01-09
Deprecate other gafrc options

graahnul-grom:
- `attribute-promotion` => `library::attribute-promotion` (`bool`, `true`)
- `promote-invisible` => `library::promote-invisible` (`bool`, `false`)
- `keep-invisible` => `library::keep-invisible` (`bool`, `true`)
- `make-backup-files` => `schematic::make-backup-files` (`bool`, `true`)
- `always-promote-attributes` => `library::always-promote-attributes` (string list,
`footprint;device;value;model-name`)
- to make it work, `config-load!()` guile function has been modified to
accept a new optional boolean parameter: `force-load` (false by default).
So, a configuraion is not **re**loaded by default. Otherwise, any call
to `config-load!` made after the deprecated options are set will erase them
(a couple of refreshing days with debugger revealed that :)).

### DIFF
--- a/liblepton/lib/geda-system.conf
+++ b/liblepton/lib/geda-system.conf
@@ -96,6 +96,7 @@ log-window=later
 
 [schematic.attrib]
 promote=true
+promote-invisible=false
 
 
 
--- a/liblepton/lib/geda-system.conf
+++ b/liblepton/lib/geda-system.conf
@@ -94,6 +94,9 @@ logging=true
 auto-save-interval=120
 log-window=later
 
+[schematic.backup]
+create-files=true
+
 [schematic.attrib]
 promote=true
 promote-invisible=false
--- a/liblepton/lib/geda-system.conf
+++ b/liblepton/lib/geda-system.conf
@@ -101,6 +101,7 @@ create-files=true
 promote=true
 promote-invisible=false
 keep-invisible=true
+always-promote=footprint;device;value;model-name
 
 
 
--- a/liblepton/lib/geda-system.conf
+++ b/liblepton/lib/geda-system.conf
@@ -94,6 +94,9 @@ logging=true
 auto-save-interval=120
 log-window=later
 
+[schematic.attrib]
+promote=true
+
 
 
 #
--- a/liblepton/lib/geda-system.conf
+++ b/liblepton/lib/geda-system.conf
@@ -97,6 +97,7 @@ log-window=later
 [schematic.attrib]
 promote=true
 promote-invisible=false
+keep-invisible=true
 
 
 
--- a/liblepton/scheme/unit-tests/t0402-config.scm
+++ b/liblepton/scheme/unit-tests/t0402-config.scm
@@ -84,9 +84,9 @@
     (assert-equal a (config-load! a))
     (assert-true (config-loaded? a))
     (chmod *testdirAconf* #o000) ;; Make conf unreadable
-    (assert-thrown 'system-error (config-load! a)))
+    (assert-thrown 'system-error (config-load! a #:force-load #t)))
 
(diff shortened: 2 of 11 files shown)


liblepton/lib/geda-system.conf
```diff
@@ -135,3 +136,15 @@ monochrome=false
...
+
+#
+# liblepton:
+#
+
+[library]
+attribute-promotion=true
+promote-invisible=false
+keep-invisible=true
+always-promote-attributes=footprint;device;value;model-name
```
vzh: @graahnul-grom,
Since the new config keys are related to `lepton-schematic` only, I would suggest to add two new groups: `[schematic.attribute]` and `[schematic.backup]` and move the keys there. Then we could shorten the key names, say, `attribute-promotion` -> `promotion`, `always-promote-attributes` -> `always-promote`, `make-backup-files` -> something like `save-files` or `enable-save`, or even reverse the logic and use just `disable` to disable backups (seems even better IMO).

liblepton/scheme/unit-tests/t0402-config.scm
```diff
@@ -83,9 +83,9 @@
     (assert-equal a (config-load! a))
     (assert-true (config-loaded? a))
     (chmod *testdirAconf* #o000) ;; Make conf unreadable
-    (assert-thrown 'system-error (config-load! a)))
+    (assert-thrown 'system-error (config-load! a #t)))
```
vzh: FWIW, here and in other places I would use something like `'force` or `'force-load` instead of just `#t` to improve readability.

liblepton/src/i_vars.c
```diff
@@ -40,43 +37,111 @@ int   default_make_backup_files = TRUE;
...
+  gboolean val = eda_config_get_boolean (cfg, group, key, &err);
+
+  gboolean success = err == NULL;
+  g_clear_error (&err);
+
+  *result = success ? val : defval;
+  return success;
+}
+
+
```
vzh: A thought aloud: we could move config processing, say, to `liblepton/scheme/geda/config.scm` the way like it is done, e.g., for `lepton-netlist`.

liblepton/scheme/unit-tests/t0402-config.scm
graahnul-grom: It's a good idea. Changed to `force-load` key parameter.

liblepton/lib/geda-system.conf
graahnul-grom: Maybe [schematic.attribute**s**]?
```ini
[schematic.attributes]
promote=true
always-promote=footprint;device;value;model-name
promote-invisible=false
keep-invisible=true
```

Logic inversion in settings could be quite confusing, IMHO.
```ini
[schematic.backup]
create-files=true
```

vzh: ... or just `[schematic.attrib]`

liblepton/scheme/unit-tests/t0402-config.scm
vzh: He-he, I've meant just using `(config-load! a 'force-load)`. The symbol `'force-load` is always true. But fair enough, your new code is OK with me.

liblepton/src/i_vars.c
graahnul-grom: What is supposed to be moved to `geda/config.scm`?
And I don't see any code related to lepton-netlist there.
Could you please elaborate on this?

vzh: I meant something for `liblepton` like we have now for netlister in `netlist/scheme/netlist/config.scm`, that is, processing of config settings in Scheme.

graahnul-grom: I thought about that, too.

PR #543, opened 2020-01-11
Deprecate the default-titleblock RC option in lepton-schematic

graahnul-grom:
Replace it with the `schematic.gui::titleblock-symname` configuration
key (string, "title-B.sym" by default).

@vzh A few questions:
- I've renamed the option by analogy with the `bus-ripper-symname`,
but we can follow the general pattern and retain the old name. WDYT?
- Where does the `B` paper size used? Maybe we should change the
default value (`A3` or `A4`, for example. BTW, `iso_a4` is the default
value for the `paper` keys)?
- In the old code the titleblock symbol is placed at position `(40000 40000)`.
Why exactly these coordinates are used? Could it be placed, say, at `(0 0)`?

### DIFF
--- a/liblepton/lib/geda-system.conf
+++ b/liblepton/lib/geda-system.conf
@@ -44,6 +44,7 @@ scroll-wheel=classic
 grid-mode=mesh
 dots-grid-mode=variable
 net-selection-mode=enabled_net
+default-titleblock=title-B.sym
 
 [schematic.tabs]
 show-close-button=true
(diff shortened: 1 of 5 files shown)


vzh:
- I would prefer just `titleblock` (or leave the old name) instead of this abbreviation with `symname` at the end.
- The `B` size is imperial (British) one, and the gEDA project has been founded in US.  Moreover, for most people in the Lepton team English is the first language :-)
- Such coords could be used in the times when `gschem` was not able to use negative X,Y numbers.  IIUC, nothing prevents us from using other coords.  However in this case some coords may get negative and I don't know if it is good or not for reading/parsing.

BTW, why didn't you change `geda*.conf` to `lepton*.conf` in the warning messages?

graahnul-grom:
Option's name has been changed to `default-titleblock`.
It's value and symbol position remains the same.
File names in the warning message are set using `config-filename()`.

PR #544, opened 2020-01-13
Fix Netlist->allegro error in lepton-schematic

graahnul-grom:
`lepton-schematic` logs an error when the `Netlist->allegro`
menu item is clicked.

`make-toplevel-schematic()` function (which is an
alias for `file-name-list->schematic()`) no longer
accepts 2 arguments (since git rev. 23e8685).


vzh:
Thanks :-)

PR #546, opened 2020-01-17
Fix hierarchical netlist processing

### LINKED ISSUE #528: gsch2pcb work <--> lepton-sch2pcb does not, different netlists
Then working with a schematic with hierarchy I have discovered gsch2pcb work but then using leptong-sch2pcb net name is not propagated down in hierarchy and there is no warning.

vzh:
The patch-set includes changes as follows:

- Added docstrings and comments for several functions.

- The module `(netlist rename)` has been removed.  All previous renaming
  functionality has been wholly thrown away.

- The module `(netlist pin-net)` has been removed.  `<pin-net>` records
  are no longer used in netlisting.

- Introduced *hierarchical* (*over-port*, or *vertical*)
  connections which are used to produce net names for schematics
  with subcircuits.  This resulted in the following changes in
  output netlists:

  - The `geda` backend no longer displays intermediate renamings
    the same way it is used for unnamed nets shown in netlists.
    Renamed unnamed nets of a subschematic are shown by the geda
    backend differently, e.g.,
    "hierarchical/subciruit/name/unnamed_net_at_XxY", where X and
    Y are coords of one of their members.

  - If a net has more than one preferred attribute attached
    ("net=" or "netname="), the lowest one by means of
    the predicate refdes<?() is chosen then.

  - The sequence numbers of unconnected pins and unnamed nets have
    been reduced due to removing of internal ones from the
    sequence.

- The <package-pin> record now contains a new field,
  `port-connection`, which holds a hierarchical
  `<schematic-connection>` a pin belongs to.

- Added new tests for a hierarchical schematic with toplevel
  unnamed nets (thanks to Karl Hammar who provided initial test
  schematics the schematic was compiled from).

- Added new checks for the `<schematic-port>` records:

  - If inner port component has no pins.

  - If there are several components that could represent a port
    for a pin.
          
- The `<subschematic>` record now contains a new field, `parent`.
  It is intended to contain the component the subschematic belongs
  to, if any.

- Avoid crashes when source file cannot be found.

- `error()` has been replaced with `netlist-error` in several places.

Closes #528


vzh:
I'll merge the commits since Nicklas confirmed it works as expected (please see our today's conversation on the geda-user list).

PR #547, opened 2020-01-18
Enable paper selection in the lepton-schematic's print dialog

graahnul-grom:
Enable the "Paper size" and "Orientation" combo boxes:

disabled

=>

enabled


vzh:
:+1:

PR #552, opened 2020-01-29
Rename gschem modules

vzh:
Rename modules named `(gschem something)` to `(schematic something)`.

Affects #39.


graahnul-grom:
Awesome, you did it!

It's time for me to start using some kind of deprecation
warnings suppressor :)
```scheme
; Usage: ( use-lepton-modules '(lepton log) '(geda log) )
;
( define ( use-lepton-modules new old )
( let
  (
  ( mnew ( false-if-exception (resolve-interface new) ) )
  ( mold #f )
  )

  ( if mnew
    ( module-use! (current-module) mnew )
  )

  ( unless mnew
    ( set! mold ( false-if-exception (resolve-interface old) ) )
    ( if mold
        ( module-use! (current-module) mold )
    )
  )

) ; let
) ; use-lepton-modules()
```

vzh:
The better way would be to fix old code, I believe, wouldn't it? :-)
Anyways, we won't support backwards compatibility with `geda-gaf` for ever. The next "stable" release (be it 1.10 or 2.0) should be the last one supporting it.
(BTW, now I'm preparing a next chunk of renamings, now it is about `(geda core *)` modules...)

Should I consider your previous comment as your approval for this PR? ;-)

graahnul-grom:
> Should I consider your previous comment as your approval for this PR?

Definitely! I was playing around with new code yesterday and was
astonished with the amount of `use-modules` I have to change in
my local scripts (blpp). Then I thought that from time to time it's
necessary to run different versions of lepton-eda (e.g. for supporting
users), so I came up with that somewhat clumsy solution shown above.

> The next "stable" release (be it 1.10 or 2.0) should be the last one supporting it.

:man_facepalming:
Please, please
**Major** release!

I'm for `2.0`.

vzh:
[Please, be patient](https://github.com/lepton-eda/lepton-eda/wiki/Roadmap) ;-)

PR #554, opened 2020-01-30
Do not use deprecated (geda config) module

graahnul-grom:


vzh:
Thank you!
As you can see, superfluous spaces make git-grepping not such successful :-)

graahnul-grom:
Aha, `DBCS` strikes back! :-)
`git grep 'geda config'`? No!
`git grep -E 'geda +config'`!

vzh:
... strikes back via YOUR code ;-) so its your style. And don't tell me about other languages, the same happened with many people before more than once during refactoring.
Using `git grep -E` is annoying.

graahnul-grom:
Please do not worry, I'm just kidding. :)
`grep -E` is awkward, I agree.

PR #556, opened 2020-01-31
Rename core modules

vzh:
- Renamed `(geda *)` and `(gschem *)` core modules.
- Updated NEWS.
- Added missing `(geda object)` tests.

Affects #39.

PR #560, opened 2020-02-03
Get rid of annoying word "complex" in the repository

vzh:
The word **complex** is ambiguous since it is used in Scheme to designate complex numbers.

The patch set is aimed to avoid using the word **complex** in file names and in the code of libraries and core tools where **component** is meant:

- Rename the module `(lepton core complex)` to `(lepton core component)`.
- Rename several C files containing the name in their names.
- Get rid of the unused type `GedaComplex'.
- Rename the type `COMPLEX` to `COMPONENT`.
- Rename the type `OBJ_COMPLEX` to `OBJ_COMPONENT`.
- Rename lots of local variables and function arguments.
- Rename local and global C functions and core Scheme procedures.
- Replace or remove the word in comments.
- Rename appropriate fields of the `st_object` structure.
- Avoid using of the word in several messages of the tools (in `printf` and `fprintf`).
- Fix doxygen descriptions and comments in the code to avoid using of the word.
- Get rid of the unused field `inside_redraw` in the `st_gschem_toplevel` structure.
- Get rid of the unused prototype `o_complex_set_filename`.
- Fix Lepton Scheme API documentation.
- The Scheme object type `'complex` is left as is for backwards compatibility.


vzh:
Argh, I hurried and pushed the (rebased) branch with commits from #556.
Well, this should be applied after #556 is in the repo. Let's wait :-)

graahnul-grom:
Isn't the commit "rename: (gschem core hook) to (schematic core hook)."
already in master?

vzh:
It is there, and this is just an artifact of rebasing.  This commit (the part left) only updates NEWS.

graahnul-grom:
`object-type()` function from the `(lepton core object)` still returns
"complex" for component objects. Is this intended?
```c
SCM_SYMBOL (complex_sym , "complex");
```
`liblepton/src/scheme_object.c:38`

vzh:
Yes, for backwards compatibility with user scripts.  Maybe we make a change in some *stable* release.

graahnul-grom:
Please do not use this *stable/unstable* terminology. Maybe it makes
sense for Debian operating system distribution (I don't know, it's none
of my business) but here it is utterly inappropriate and harmful.

I've found also `complex_place_list_changed_hook`, which is not
exported and not used anywhere. What is it for?

vzh:
It was supposed to be used in `schematic/scheme/conf/schematic/deprecated.scm`.  Please see the commented out call for `complex-place-list-changed-hook`. The `place_list` is the list of selected objects being moved.  It can change when you start moving, insert those objects somewhere, or escape the operation.  The user might want to do some action on the objects being manipulated.

I wasn't sure what to do with this code, so it was left as is.

graahnul-grom:
Thank you. I've tried to install that hook and found
out that it is invoked each time the tree selection is changed in the
"Select Component..." dialog. The parameter passed to the handler
function is the object to be placed. The hook is invoked several times:
for the component itself and for each of the promotable attributes.
It also called when the behavior combo box selection is changed.
So, I think it could be quite useful.
Why isn't it exported the same way as other hooks?

vzh:
Probably because no one ever noticed it? ;-)

graahnul-grom:
:) I stumbled upon it just yesterday while grepping
the repo for word "complex".

PR #561, opened 2020-02-04
lepton-tragesym: fix the option 'generate_pinseq'

vzh:
Closes #520
Affects #492

### DIFF
--- a/utils/examples/tragesym/.gitignore
+++ b/utils/examples/tragesym/.gitignore
@@ -1,3 +1,4 @@
 Makefile
 Makefile.in
 *~
+*.sym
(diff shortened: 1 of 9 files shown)


vzh:
Right.

graahnul-grom:
@vzh I remember we have discussed renaming of `tragesym`
to `symgen`. Do you think it's worth it, or should we better
leave it alone?

vzh:
We could rename it. However, I think, it would be reasonable in case we add some new functionality to it, e.g. support for other formats, other option names etc.  In the state it is, the program is just tragesym's sibling.

graahnul-grom:
OK, agree, though it has been 100% rewritten. :-)

PR #566, opened 2020-02-07
Init commit for new examples

drforbin:
Ok.. maybe this will work.
Sorry for all the problems.
I am currently reading two books on git and will get myself up to speed.
In the past I only used git in rather trivial ways.


vzh:
OK, this is much better. There are some issues with makefiles, but we'll fix them later.
Thank you!

vzh:
@drforbin,
You can now safely delete this branch to keep your git history clean, and `git merge --ff` your master branch with upstream master.

drforbin:
Vlad, What is best way to merge master with up stream?

vzh:
If your local `master` has only commits from upstream, you can proceed as follows:
```
git checkout master
git merge --ff origin/master
```
Where `origin/master` is the name of the upstream master branch on your system.  It may have another name, e.g., on my system it is named `lepton/master`.

If you have local commits in your master that do not exist in the upstream master, you could do:
```
git checkout -b new-branch master
git checkout master
git reset --hard some-common-history-commit-hash
git merge --ff origin/master
```
Here `some-common-history-commit-hash` is the hash of the commit which is shared both by upstream `master` and your local `master`.  Use `git log` or `git gui` to find the common history point in such a case.

PR #568, opened 2020-02-10
Fix build of lepton-schematic with g++.

vzh:


vzh:
@graahnul-grom, thank you for correction!

PR #574, opened 2020-02-15
Fix 'make check --disable-deprecated'

vzh:
On my system the command fails in some cases, e.g. if a previous version of lepton is installed.  There are two patches here:
- The first one fixes this by using especially crafted variable for this goal `LEPTON_INHIBIT_RC_FILES`.
- The second one renames alike variable in `lepton-cli` for uniformity.


vzh:
You'll probably need to have some old lepton installed (say, before renaming of  core Scheme modules) to reproduce those errors, not sure. In my case, it bit me for quite a long time.

PR #580, opened 2020-02-21
Misc small fixes in various places

vzh:
BTW, do we still need 'autogen.sh' in the repository? On my system, `autoreconf -ivf` does the job as expected without issues.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -25,7 +25,7 @@ ac_script=configure.ac
 am_version=1.11.0
 aclocal_flags="$ACLOCAL_FLAGS -I m4"
 tooldir=build-tools
-podirs="liblepton/po cli/po schematic/po gattrib/po gnetlist/po symcheck/po"
+podirs="liblepton/po cli/po schematic/po attrib/po netlist/po symcheck/po"
 
 srcdir=`dirname $0`
 if test "x$srcdir" = x ; then srcdir=.; fi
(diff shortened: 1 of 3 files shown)


noqsi:
autogen.sh is a familiar mechanism. I see -ivf and think "in vitro fertilization" ツ

vzh:
:-) it can be just -if. Well, we could put `autoreconf` inside `autogen.sh`, it does the very job we have several lines of code for. And it has been a standard thing in autotools for quite a while.

graahnul-grom:
`autoreconf -if` works.

graahnul-grom:
Indeed, could we reduce `autogen.sh` to
```sh
#!/bin/sh

autoreconf -if
```
?

vzh:
Let's try ;-)  If something will go wrong, we'll revert it.  Then we would need to update README to mention only `autoreconf` while leaving `autogen.sh` in place for some time.

graahnul-grom:
I don't know about other package building systems, but the
FreeBSD ports infrastructure doesn't use `autogen.sh` at all,
it just invokes `autoreconf` if `USES=autoreconf` is specified
in the [Makefile](https://svnweb.freebsd.org/ports/head/cad/lepton-eda/Makefile?revision=521770&view=co).

graahnul-grom:
`README`: the copyright line at the top of the files is slightly outdated
(`2017-2019 Lepton Developers`). And why do we need it at all, if the
same copyright information is present in the `License` section below?

vzh:
Yup, one of copyright chunks could be removed.

PR #582, opened 2020-02-22
Remove obsolete stuff - part 3

graahnul-grom:
- remove `--disable-deprecated` `configure` option
- remove `--enable-compat-symlinks` `configure` option
- use new name for custom `gtkrc` files (`lepton-gtkrc`)
- don't mention obsolete environment variables in man pages

It's possible to move `gtkrc` loading code to `liblepton` so that
other tools in the suite (e.g. `attrib`) could use it, but it makes
`liblepton` dependent on `GTK`.


vzh:
It would be better to create a separate library for GUI functions, we discussed this before.

BTW, after building with the previous part of changes and installing lepton, I was at a loss after launching `lepton-schematic`: another look, another behaviour, and what to do next?  We have to provide some almost seamless way of porting user's config to new directories.  I would suggest a dialog box in `lepton-schematic` offering the user to copy old settings (`gafrc`, `lepton-user.conf` or, if it is missing, `geda-user.conf`) from `.gEDA` to the new config residence, or suggest to make a new config from scratch.  Non-GUI tools could output some descriptive text about this.
WDYT?

graahnul-grom:
It's a good idea.
- should we offer to copy `gschemrc`, too?
- if we're going to copy configuration files anyway, maybe it's time to switch to `lepton*.conf`?
- what do you mean by *making a new config from scratch*?

vzh:
>     * should we offer to copy `gschemrc`, too?
Not sure. Are there useful settings there these days? Probably some scripts are loaded from the file? The latter case would involve manual intervention, though.

>     * if we're going to copy configuration files anyway,
>      maybe it's time to switch to `lepton*.conf`?
Of course ;-)

>     * what do you mean by _making a new config from scratch_?
That is, if there is no `lepton-user.conf` in `.local/share/lepton-eda` (is this a correct place for config? :-)), and the user does not want to copy the old stuff or there is no such file yet, the program prompts the user to create it with default settings, and, if he agrees, actually creates it.

graahnul-grom:
i.e. you propose to make a copy of `lepton-system.conf` in the
user's configuration directory (`~/.config/lepton-eda/`)? Why not just
`touch(1)` `~/.config/lepton-eda/lepton-user.conf` if there's no old
file (geda-user.conf) to convert?

graahnul-grom:
Anyway, I've started working on it, need to simplify
`(lepton legacy-config)`, `(lepton legacy-config keylist)`...
There's a better conversion method than currently implemented.

graahnul-grom:
`gschemrc`: I mean user's code there, of course, all the option
are already deprecated.

drforbin:
I think we should have ONE config file with sections (i.e.) gitconfig.
Also this should be a text file.

drforbin:
Dmn,

Just to clarify. All config options which are not specified as NEW are handled in the old way in the old files just moved to the new locations?

drforbin:
Is the project specific gafrc going to be depreciated?

vzh:
@graahnul-grom, there is no need to copy any config file if the user has no one :-) Using of `touch` as you suggested would work as expected, I think.

vzh:
@drforbin, we're going towards eliminating any `*rc` files. So, yes, some time all `gafrc` and `gschemrc` will be useless.  Lepton's config files should be declarative ones. We have to specify, though, how to load user specific extension scripts...

graahnul-grom:
@drforbin All the options have been deprecated, except those that
intended to work with component and source libraries in `gafrc`.
Deprecated options are still read from the `gschemrc` and `gafrc`
files, but, if encountered, the warning message is printed to `stderr`.

graahnul-grom:
@drforbin BTW, `git` doesn't use only one configuration file.
There is `.git/config` file in every repository (local settings).
So, this is similar to our approach, where we use `lepton-user.conf`
for user configuration (like `~/.gitconfig`) and local configuration file
`lepton.conf` for project-specific settings (like `.git/config`).

PR #588, opened 2020-03-07
symcheck: Fix localization.

vzh:
The localization was not working at least since rewritting the
tool in Scheme.

PR #589, opened 2020-03-07
netlist: Add localization support.

vzh:
Fixed (really, added) localization support for netlister's new Scheme code.


vzh:
As in #588, how this works will be apparent when I add a PO file full of translations.

PR #590, opened 2020-03-09
Update the lepton-upcfg utility

graahnul-grom:
- source configuration file is copied before the conversion, so that
custom settings and comments are preserved.
- remove the `--system` command line option: `geda-gaf` has no system
configuration file, lepton-eda already provides one with new settings,
so there is no point in converting system config.
- print upgraded configuration file contents to standard output by
default; to copy resulting configuration file to appropriate location,
use (new) `--copy` (`-c`) command line option.
- to convert `geda-user.conf` in `~/.gEDA/` directory, the `--old`
(`-o`) command line option has been added.

### DIFF
--- a/liblepton/scheme/lepton/legacy-config.scm
+++ b/liblepton/scheme/lepton/legacy-config.scm
@@ -11,11 +11,20 @@
   #:use-module  ( lepton config )
   #:use-module  ( lepton legacy-config keylist )
 
+  #:export      ( upcfg-log )
   #:export      ( config-upgrade )
   #:export      ( warning-option-deprecated )
   #:export      ( warning-option-obsolete )
 )
 
+
+
+; public:
+;
+( define ( upcfg-log fmt . args )
+  ( apply format (current-error-port) fmt args )
+  ; ( apply log! 'message fmt args )
+)
 
 
 ; get configuration context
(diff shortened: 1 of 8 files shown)


liblepton/scheme/lepton/legacy-config.scm
```diff
@@ -8,27 +8,47 @@
 
   #:use-module  ( ice-9 format )
   #:use-module  ( ice-9 match )
+  #:use-module  ( ice-9 optargs ) ; define*
```
vzh: `optargs` is not required for `define*`, just for `define*-public`.

vzh:
... and
```
$ lepton-cli config --user
/home/vzh/.config/lepton-eda/geda-user.conf
```
Hmm...

vzh:
Shouldn't be `config_legacy_mode` set to `FALSE` by default?

liblepton/scheme/lepton/legacy-config.scm
graahnul-grom: yes, thank you.

graahnul-grom:
So, we have to deal with multiple user config files. Not sure if this can be reasonably automated.

graahnul-grom:
Maybe it's better to add an option to pass the source config file's location.

vzh:
> I didn't turn legacy config mode off yet in `liblepton`. Perhaps, it's time to do so.

Yes, we've already removed the option `--disable-deprecated`, so it's time to go forward.

vzh:
> So, we have to deal with multiple user config files. Not sure if this can be reasonably automated.

The logic could be simple:
- search for all possible places where legacy config files may live;
- report any of them found;
- use the most recently updated one for upgrade and report this to the user prompting about manual intervention in case the option is not suitable.

graahnul-grom:
There are 3 user config files, or did I forget something?

1. `$HOME/.gEDA/geda-user.conf`
2. `$XDG_CONFIG_HOME/gEDA/geda-user.conf`
3. `$XDG_CONFIG_HOME/lepton-eda/geda-user.conf`

vzh:
Yes, I cannot remember any other locations.

graahnul-grom:
@vzh Though I didn't want to make this utility interactive, it seems
that otherwise it could become more complicated...
I've added the code that searches for available user configuration
files (those 3 files above) and offers the user to choose which one
to upgrade. Maybe some messages, interactive prompt behavior,
as well as the code itself should be improved.

vzh:
I'm fine with the change. And now, I think, it's time for a new release :-) Please update NEWS with changes in this commit set.

PR #595, opened 2020-03-11
Get rid of some fields of the TOPLEVEL structure

vzh:
Reduce code dependency on `TOPLEVEL` by removing some of its fields. Instead, use new config settings to get same info.


graahnul-grom:
Excellent refactoring, bravo! :-)

PR #596, opened 2020-03-13
Fix the lepton-archive log file name

graahnul-grom:
Do not use the "lepton-" prefix for log files generated by `lepton-archive`
for consistency with the other tools in the suite.
BTW, `lepton-archive`doesn't seem to write anything to the log...

### DIFF
--- a/utils/scripts/lepton-archive.in
+++ b/utils/scripts/lepton-archive.in
@@ -861,5 +861,5 @@ the file manually.\n"))
     ((@@ (guile-user) parse-rc) "lepton-archive" "gafrc" )
     ;; Init log domain and create log file right away even if
     ;; logging is enabled.
-    (init-log "lepton-archive")
+    (init-log "archive")
     (main)))


vzh:
> Do not use the "lepton-" prefix for log files generated by `lepton-archive`
> for consistency with the other tools in the suite.

I see only `lepton-netlist` initialises log in such a way. Hmm... What do other tools do? Isn't `init-log` useless?
Well, yes, it's better to use shorter name, since it goes to `lepton-eda` folder, anyways.

> BTW, `lepton-archive`doesn't seem to write anything to the log...
It may do it in future, though I'm in doubt whether the chunk of code does anything useful.
Anyway, thank you!

graahnul-grom:
Other tools use `s_log_init()` C function (schematic, attrib), passing
the short program name to it.
`init-log` calls `s_log_init()`, too.

vzh:
Ah, thanks :-)

PR #598, opened 2020-03-14
Turn off legacy configuration mode

graahnul-grom:
Use new `lepton*.conf` configuration files by default.
Use new names for configuration parameters in `liblepton`,
`lepton-schematic` and `lepton-netlist`.


vzh:
Great, thank you!

PR #606, opened 2020-03-20
lepton-netlist: Update Russian translation

vzh:
Localisation support has been reinstated in #589. Now, here is an updated translation into Russian.

### DIFF
--- a/netlist/po/ru.po
+++ b/netlist/po/ru.po
@@ -1,18 +1,19 @@
-# Russian translation for gEDA gnetlist
+# Russian translation for lepton-netlist
 # Copyright (c) 2014 gEDA Developers
-# Copyright (c) Vladimir Zhbanov <vzhbanov@gmail.com>, 2014.
+# Copyright (c) 2014, 2020 Vladimir Zhbanov <vzhbanov@gmail.com>
 #
-# This file is distributed under the same license as the gEDA gnetlist package.
+# This file is distributed under the same license as the lepton-eda
+# package.
 #
 # Переводчикам: см. примечания в liblepton/po/ru.po
 msgid ""
 msgstr ""
-"Project-Id-Version: geda gnetlist\n"
+"Project-Id-Version: lepton-eda\n"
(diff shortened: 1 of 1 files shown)


graahnul-grom:
@vzh Unfortunately, I can't get translated messages on my system.
Haven't tried on Linux yet, though.
I set both `LC_ALL` and `LANG` environment variables
to `ru_RU.UTF-8`.
`LANG` works for other tools (e.g. lepton-schematic), but not
for `lepton-netlist`. The same with `symcheck` (#605). :-(

vzh:
Please check that `@localedir@` has been replaced with a correct value in
the expression `(bindtextdomain %textdomain "@localedir@")` in `lepton-netlist`. Then, at least `lepton-netlist --help` should work :-) (Some messages have not been prepared for translation in the code.)

graahnul-grom:
Yes, I have checked it. `@localedir@` was correctly replaced.

vzh:
In output of `make` in `netlist/po` I see:
> test ! -f ./lepton-netlist.pot || \
  test -z "nl.gmo ru.gmo" || make nl.gmo ru.gmo

and then
> rm -f ru.gmo && /usr/bin/msgfmt -c --statistics --verbose -o ru.gmo ru.po

May it be that `ru.gmo` has not been re-created on your system?

graahnul-grom:
*.gmo files are created. Strangely, if I delete them, then `make`
does nothing to re-generate these files.
I tried this also on Linux (antix), and still unable to get localized
messages from `lepton-netlist` (other tools are OK, running
`lepton-cli` with the same command gives translated output).

vzh:
Hm, I see `antix` is based on `Debian stable`. I run `Debian` and anything is translated properly.  I would suggest to concentrate on the translation text and apply the branch in order to not get stuck here.  Besides, the issue will anyway be solved by moving netlister's Scheme modules to liblepton's directory along with all translation files, which I hope will happen in near future.

PR #607, opened 2020-03-22
build: Get rid of warnings about possibly uninitialized variables.

vzh:
The report by @bdale:

I just found a patch in the Debian BTS from our Ubuntu friends.  It
turns out that if you compile lepton-eda wiht -O3, which they do on at
least one platform, you get warnings about possibly uninitialized
variables.  The diff they offered up is pretty trivial, I've added it to
my packaging of 1.9.10, you might want to merge it into your repo too.


vzh:
I'll merge this since the changes are trivial and cannot break anything.

PR #614, opened 2020-03-29
lepton-symcheck: Fix 'make check'

vzh:
'make check' failed on systems, where localised messages were
output (found that after adding  l10n support and Russian translation).


vzh:
I'll merge this, since it apparently fixes things on my system.

PR #615, opened 2020-03-29
Fix placeholder rendering on export

vzh:
The commit set includes changes and fixes as follows:
- Fixed placeholder rendering in the output of `lepton-cli export`.
- Moved text bounds calculation code from `lepton-cli` and `lepton-schematic` to `liblepton` and simplified it.
- Fixed assertion failures when calculating text object bounds, but no renderer is available, e.g. when a Scheme script was evaluated by `lepton-cli`.
- Removed `TOPLEVEL` field `rendered_text_bounds_func`.
- Changed build order: `libleptonrenderer` is now compiled before liblepton.
- Fixed library names in some docs and comments.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,7 +4,7 @@ endif
 if ENABLE_CONTRIB
 CONTRIB_DIR=contrib
 endif
-SUBDIRS = liblepton libleptonrenderer cli schematic ${ATTRIB_DIR} \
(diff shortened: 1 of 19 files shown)


vzh:
@graahnul-grom, thank you! Yes, we should communicate more often. Any suggestions are welcome.  The GH format suggests some means for communicating, though we could use more these days :-)

PR #619, opened 2020-04-02
Use Ubuntu Xenial for continuous integration builds.

vzh:
Xenial is used as default build environment on travis-ci.com since
about a year:
https://changelog.travis-ci.com/xenial-as-the-default-build-environment-99476

Switching to it allows to get rid of additional installation of SRFI-64 Scheme modules that are missing in Ubuntu Trusty due to old Guile version 2.0.9 being used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,7 @@
 # some basic Linux-based continuous integration testing (for free).
 
 sudo: required
-dist: trusty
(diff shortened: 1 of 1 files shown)


bdale:
Looks good to me

vzh:
Thank you. Then I'll merge it just now.

Issue #622, opened 2020-04-06
Printing Issue : Text Translates In Y-Axis


suboptimallife:
I am having a problem when printing schematics from Lepton and Google searches didn't bring up anything relevant. I was using a Lepton build downloaded from source back in January but downloaded 1.9.10.20200319 and the problem carries over between versions. Text which is positioned correctly in the schematic is translated in the y-axis in both print preview and print-to-file output (see attachment [print_test.pdf](https://github.com/lepton-eda/lepton-eda/files/4436818/print_test.pdf)). The issue is repeatable with different schematics, I have created an example schematic (file extension changed for GitHub compatibility [print_test.txt](https://github.com/lepton-eda/lepton-eda/files/4436837/print_test.txt)).

When using lepton-cli export the resulting output file is entirely blank (see attachment [print_test01.pdf](https://github.com/lepton-eda/lepton-eda/files/4436842/print_test01.pdf))

vzh:
Could you please embed your title symbol into the schematic file or use `lepton-archive` to get all files necessary to correctly render it?
And  could you please show us the exact `lepton-cli` command you used for export?

suboptimallife:
lepton-archive output attached. Commands used were:

lepton-cli export -f pdf -o print_test01.pdf -p iso_a3 -l landscape --no-color print_test.sch
lepton-cli export -f pdf -o print_test01.pdf -p iso_a3 -l landscape print_test.sch
lepton-cli export -o print_test01.pdf print_test.sch

suboptimallife:
[project-archive.tar.gz](https://github.com/lepton-eda/lepton-eda/files/4438218/project-archive.tar.gz)

vzh:
OK, thanks.  Could you try the following commands and report their output here:
> lepton-cli --version
> lepton-schematic --version

Please try also to `rm ~/.cache/guile -rf` and re-check if `lepton-cli` works or not.

The issue with printing from `lepton-schematic` may be solved in #623.  I would very appreciate if you rebuild Lepton using the branch there and try if it works better.

suboptimallife:
lepton-cli --version Lepton EDA 1.9.10.20200319 (git: )
lepton-schematic --version Lepton EDA 1.9.10.20200319 (git: )

rm ~/.cache/guile -rf still results in a blank pdf output from lepton-cli export

suboptimallife:
[...]

PR #623, opened 2020-04-06
schematic: Use the same font for printing and rendering on the screen.

vzh:
Affects #622.
Incorrect font rendering issue on printing may be caused by different fonts used in GUI and for printing.


vzh:
@graahnul-grom, thank you for review!

vzh:
JFTR, the patch did not solve the issue it was written for.  OTOH, it makes printed output in `lepton-schematic` more predictable.

PR #624, opened 2020-04-06
Get rid of TOPLEVEL arguments in many functions.

vzh:


vzh:
Since here are mostly mechanical changes, and I've reviewed them myself after a while, I'll push them to check if anything is OK.  Please feel free to blame and curse me if anything will hit you :-)

PR #625, opened 2020-04-09
schematic: Fix dereference before NULL check (coverity scan issue 209544).

vzh:


schematic/src/o_move.c
```diff
@@ -539,9 +539,6 @@ void o_move_check_endpoint(GschemToplevel *w_current, OBJECT * object)
   PAGE *page = gschem_page_view_get_page (page_view);
   g_return_if_fail (page != NULL);
 
-  if (!object)
-  return;
```
graahnul-grom, line 543: Wouldn't it be better to move this check to the beginning of the function?

vzh: Yes, you're right.  It's better to amend the check.

vzh: Please check the recent version.

graahnul-grom: OK

PR #628, opened 2020-04-11
Fix several potential memory leaks in liblepton and attrib

graahnul-grom:
Interestingly, these were detected by the `scan-build` static
code analysis tool (part of `clang`), but I can't find them
 among the coverity scan issues.

@vzh `scan-build` generates html output very similar to what we see
on scan.coverity.com. If need be, I can send it by e-mail (my first
intention was to upload it to my lepton page, until I `du -hsx` it:
almost 80 Mb :-), though in compressed form it takes about 400 Kb).

### DIFF
--- a/attrib/src/x_gtksheet.c
+++ b/attrib/src/x_gtksheet.c
@@ -71,7 +71,6 @@ x_gtksheet_init()
   const gchar *folder[]= {_("Components"),
                           _("Nets"),
                           _("Pins")};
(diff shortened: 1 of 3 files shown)


vzh:
> I can send it by e-mail

Please, no :-) Let's fix what coverity site shows us first. (I suspect, it may use the same utility, maybe an older or newer version.)

PR #632, opened 2020-05-05
lepton-schematic widgets as top-level windows

graahnul-grom:
Allow widgets to behave like top-level windows in non-docking GUI.
Add a new configuration key to turn it on and off:
`schematic.gui::use-toplevel-windows` (`false` by default).
This feature was requested by Girvin Herr on the geda-user
mailing list [1]. Quote:

> BTW: Another annoying thing I noticed about lepton-schematic is that the
> additional windows (i.e. status and page manager) do not honor window
> focus - they stay on top, which makes it difficult to work in the
> schematic editing main window without moving them aside. I do believe
> that gEDA 1.8 did honor window focus and the same windows went to the
> background when focus was changed to the main window. Fixing that could
> be an enhancement for a subsequent lepton release.

@vzh Could you please verify will this work with a tiling window manager?
Perhaps, you could suggest a more appropriate name than `use-toplevel-windows`.

[1] http://www.delorie.com/archives/browse.cgi?p=geda-user/2020/01/21/15:31:24

### DIFF
--- a/schematic/include/gschem_find_text_state.h
+++ b/schematic/include/gschem_find_text_state.h
@@ -22,11 +22,15 @@
  * \brief
  */
 
-#define GSCHEM_TYPE_FIND_TEXT_STATE           (gschem_find_text_state_get_type())
-#define GSCHEM_FIND_TEXT_STATE(obj)           (G_TYPE_CHECK_INSTANCE_CAST ((obj), GSCHEM_TYPE_FIND_TEXT_STATE, GschemFindTextState))
-#define GSCHEM_FIND_TEXT_STATE_CLASS(klass)   (G_TYPE_CHECK_CLASS_CAST ((klass),  GSCHEM_TYPE_FIND_TEXT_STATE, GschemFindTextStateClass))
-#define GSCHEM_IS_FIND_TEXT_STATE(obj)        (G_TYPE_CHECK_INSTANCE_TYPE ((obj), GSCHEM_TYPE_FIND_TEXT_STATE))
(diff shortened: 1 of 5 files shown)


vzh:
@graahnul-grom, I've compiled `lepton-schematic` with your changes. AFAICS, the following dialogs/windows are no longer modal:
- Page manager
- Log window
- Edit text
- Color scheme editor
- Object properties
- Options
- Font

The following ones are still modal:
- Add attribute
- Add component
- Add text
- Print dialog (*)
- Open file (*)
- Save as (*)
- Execute script (*)
- Edit slot
- Page revert (*)
- Add picture (*)
- Autonumber dialog
- Coord window
- Hotkeys (*)
- About (*)

Not sure if any one marked with (*) should be non-modal at all.  In doubt of others. I'd like `Add component` to be non-modal in the first place.

I would suggest the config key name `modal-dialogs` or `modal-windows` (reverse its meaning), and set it initially to `true` to maintain previous behavior.

graahnul-grom:
This PR introduces the config option that may change the behavior
of eight widgets managed in `x_widgets.c` code:

1) Page manager
2) Log window
3) Edit text
4) Color scheme editor
5) Object properties
6) Options
7) Font selector
8) Find text results

At present, those 8 widgets behave like *non-modal dialogs*; setting
the option make them behave like *top-level windows*.
No other windows/dialogs are affected.
To be able to manage other dialogs this way, we must convert them
to "widgets" (derive them from `GschemBinClass`, add `w_current`
property, etc. etc.).

Maybe the terminology is confusing:
- *modal dialog*: dialog box that blocks any input to other windows.
- *non-modal dialog*: dialog box that does not block input to other windows,
  but stays on top of its parent window.
- *top-level window*: window/dialog that can be raised/lowered independently,
  i.e. can be placed on top of the main window or below it.

So, this setting is not about dialogs being "modal" or not.

The "Select Component..." dialog is already non-modal.

PR #633, opened 2020-05-07
Use srfi-64 to test Scheme functions in liblepton

vzh:
- Merged two Scheme test suites into one based on SRFI-64
- Fixed pollution of **tmp/** with temporary schematic files
- Added tests for promotable attributes


vzh:
@graahnul-grom, thank you for review :-)

PR #634, opened 2020-05-07
Alternative placeholders for missing symbols

graahnul-grom:
Draw either new (smaller) or classic placeholders (giant
red triangles with an exclamation mark and two lines of
text :) for missing schematic symbols, depending on value
of the `schematic.gui::draw-smaller-placeholders`
configuration key (`false` by default).
It's an open question how placeholders should look like to
not turn a schematic with missing symbols into incomprehensible
bloody mess... :-) It's just an attempt. Any ideas are welcome.

Two types of placeholders (to scale):

3-placeholder-cmp-dmn
4-placeholder-cmp-dmn

A small (!) example schematic (real schematics are hard
to look at without `Volidol`, in either case):

0-placeholder-test-dmn

1-placeholder-test-dmn

2-placeholder-test-dmn

[placeholder-test-dmn.tar.gz](https://github.com/lepton-eda/lepton-eda/files/4595351/placeholder-test-dmn.tar.gz)

### DIFF
--- a/liblepton/lib/geda-system.conf
+++ b/liblepton/lib/geda-system.conf
@@ -46,6 +46,7 @@ dots-grid-mode=variable
 net-selection-mode=enabled_net
 default-titleblock=title-B.sym
 use-toplevel-windows=false
+small-placeholders=true
 
 [schematic.tabs]
 show-close-button=true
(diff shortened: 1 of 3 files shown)


vzh:
I like it. Though the config key name could be shorter (say, `small-placeholders=true`).

graahnul-grom:
I've renamed the key to `small-placeholders`.
IIUC, you think it should be set to `true` by default?

vzh:
> I've renamed the key to `small-placeholders`.
> IIUC, you think it should be set to `true` by default?

Why not?

PR #636, opened 2020-05-12
Reduce TOPLEVEL structure and various clean-up's

vzh:
- Get rid of several unused functions (sometimes along with the files) and comments in `liblepton`, `lepton-attrib`, and `lepton-schematic`.
- Get rid of the `TOPLEVEL` argument in several `liblepton` functions or replace it with other structures.
- Replace read-only fields defining initial world sizes for page view objects with `define`s.
- Move `s_hierarchy.c` to `lepton-schematic`'s tree, since its functions are used only there.
- Add a new open file function, `schematic_file_open()`, to differentiate behavior when loading backups in GUI and non-GUI tools.
- Get rid of using `chdir()` in liblepton (hopefully, it breaks nothing).
- Replace `TOPLEVEL`'s field `show_hidden_text` with same per page view settings.  Thus the user can now show/hide text individually for different pages.


graahnul-grom:
Not doing `chdir()` when opening a page breaks calls
to `component-library()` and `source-library()` in
local `gafrc` files (where paths are almost always
specified as relative).
This can be remedied by placing `chdir()` just before
the call to `f_open_flags()` in `schematic_file_open()`.
Could you please check it?

[chdir.diff.txt](https://github.com/lepton-eda/lepton-eda/files/4622053/chdir.diff.txt)

graahnul-grom:
Test schematics:
```
chdir-test-dmn
├── 1/
│   ├── symcache_1/
│   │   └── bat2_dmn.v.sym
│   ├── 1.sch
│   └── gafrc
└── 2/
    ├── symcache_2/
    │   └── c_dmn.v.sym
    ├── 2.sch
    └── gafrc
```

`gafrc` files:
```scheme
( format #t " ++++++++  [ 1 ]  gafrc +++++++  getcwd():  [ ~s ]~%" (getcwd) )
( reset-component-library )
( component-library "symcache_1" "symcache_1")
```
```scheme
( format #t " ++++++++  [ 2 ]  gafrc +++++++  getcwd():  [ ~s ]~%" (getcwd) )
( reset-component-library )
( component-library "symcache_2" "symcache_2")
```

[chdir-test-dmn.tar.gz](https://github.com/lepton-eda/lepton-eda/files/4622132/chdir-test-dmn.tar.gz)

vzh:
Thank you for your investigation! I'll better delete corresponding commits for now. For future, it would be nice to have some tests like yours in the suite.
Please let me know if you find any other issues in the commit set.

graahnul-grom:
> I'll better delete corresponding commits for now.

I wouldn't do that. Changing directories inside
library functions was a bad idea [*], IMHO.
We could examine all invocations of `s_page_goto()`
(which used to call `chdir()`) and decide whether
directory change is needed in each case, and
eliminate `chdir()` in `f_open_flags()` with
the patch above.
Otherwise, the changes looks fine. "Autosave"
backup works (though the dialog could be improved),
turning on/off invisible text selectively for
individual pages is very useful.

[*] among other "features" of `liblepton` design,
like writing messages to the application's log
(e.g. in `s_clib_get_symbol_by_name()`) :-(.

vzh:
> > I'll better delete corresponding commits for now.
> 
> I wouldn't do that. Changing directories inside
> library functions was a bad idea [*], IMHO.

Sure.

> We could examine all invocations of `s_page_goto()`
> (which used to call `chdir()`) and decide whether
> directory change is needed in each case, and
> eliminate `chdir()` in `f_open_flags()` with
> the patch above.

OK, I wasn't going to just drop the commits.  I postponed the changes until I have more time, since now I'm busy with several other things.  If you wish to continue working on the issue just now, please do.

> Otherwise, the changes looks fine. "Autosave"
> backup works (though the dialog could be improved),
> turning on/off invisible text selectively for
> individual pages is very useful.

OK, let's give them a try :-)

> [*] among other "features" of `liblepton` design,
> like writing messages to the application's log
> (e.g. in `s_clib_get_symbol_by_name()`) :-(.

Revisiting `clib` (and, ultimately, rewriting it in Scheme) is one of my goals.  For now, in order to reduce polluting of logs, we could just reduce the severity of some messages, e.g. use `log 'debug` where appropriate.  In C code, we could add `s_log_debug()`: the same way as `s_log_message()` is a wrapper for `g_message`, it might be a wrapper for `g_debug`. WDYT?

graahnul-grom:
It's not strictly about flooding the log, I'm talking about
any side effects, which library functions should not cause.
Library function should perform its task (one, and only one
task, by the way) and return data and error information to
the caller, which decides what to do with it.
And you're right, debug messages are OK, the idea with
`s_log_debug()` seems right.

vzh:
Agreed.  And this conforms to pure functional style of programming and Unix way :-)

PR #642, opened 2020-05-26
Update examples

graahnul-grom:
After font rendering has been fixed, some of our examples
start looking not so pretty, because text is slightly
shifted now and overlaps other elements.
Schematics in `RF_Am/`, `TwoStageAmp/`, and `lightning_detector/`
have been fixed, as well as some symbols used there.
Now examples looks fine in both `lepton-schematic 1.9.10`
and `lepton-schematic` compiled from the `master` branch:

RF_Amp (before):
0-rfamp-master
0-rfamp-q1-master

RF_Amp (after):
2-rfamp-master
2-rfamp-q1-master

TwoStageAmp (before):
0-TwoStageAmp-master

TwoStageAmp (after):
1-TwoStageAmp-master

lightning_detector (before):
0-lightning-master

lightning_detector (after):
1-lightning-master


graahnul-grom:
Those attributes couldn't be fit nicely into the title block frame, so,
considering that they do not carry any important meaning, I decided
to delete them.
Now I see that I can just resize the lower part of A4; apparently, I was
quite tired yesterday, утро вечера мудренее :-).
Fixed. Please check.

lightning

vzh:
Anything is OK now.

graahnul-grom:
should we fix all title block symbols in the library the same way?

vzh:
Rather yes, I believe :-)

PR #648, opened 2020-06-02
Fix more memory leaks reported by valgrind

vzh:


liblepton/src/geda_color.c
```diff
@@ -263,6 +263,8 @@ s_color_map_from_scm (GedaColor *map, SCM lst, const char *scheme_proc_name)
       map[i].enabled = TRUE;
     }
 
+    free (rgba); /* this should stay as free (allocated from guile) */
```
graahnul-grom, line 266: it takes some time to figure out why `rgba` should stay free and what does "free"
mean in this context: "free" as in "freedom", in "free speech" or in "free launch"?
turns out, it's about function`free()`. :-))

vzh: As in "free beer" ("Налетай, уплочено!") :-) The comment stolen from another place in the code.

Issue #649, opened 2020-06-07
Need to refresh libraries after adding symbols


zo0om:
...and I'm told also after editing them (@vzh), which I guess is more important because you could import an old version of a symbol by mistake.

A perfect system would open the save dialogue at the symbol library, and present an authentication dialogue to allow you write access. But maybe this would require you to declare your creation is a symbol before trying to save it, so "New" would have to have more than one option.

zo0om:
Oh, I didn't spot that you have the option of displaying _both_ symbols and schematics, so the above doesn't make a great deal of sense!

PR #656, opened 2020-06-19
Fix file filters in the "Save As" dialog

graahnul-grom:
Decide what filter to use based on file name of the current page,
if it is e.g. `file.sym`, activate filter "Symbols (\*.sym)" and
"Schematics (\*.sch)" for `file.sch`.
Previously, the last chosen filter was set in the dialog (for example,
it's quite confusing when you try to "Save As" an `sch` file, but
the filter is set to "Symbols (\*.sym)").

Thanks to Matthew (@zo0om) for reporting it on gitter.

Probably, affects issue #649 (I'm not sure I understand what 
that issue is about).

### DIFF
--- a/schematic/src/x_fileselect.c
+++ b/schematic/src/x_fileselect.c
@@ -46,6 +46,22 @@ add_filter (GtkFileChooser* filechooser,
 
 
 
+static gboolean
+filename_sym (const gchar* fname)
+{
+  return g_str_has_suffix (fname, ".sym");
+}
+
+
+
+static gboolean
+filename_sch (const gchar* fname)
+{
+  return g_str_has_suffix (fname, ".sch");
+}
+
+
+
 /*! \brief Creates filter for file chooser.
  *  \par Function Description
  *  This function adds file filters to <B>filechooser</B>.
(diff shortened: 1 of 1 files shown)


vzh:
Just a note.  Some users of some OSes used to use all caps or camel-case for file extensions :-)  It could be probably useful to match file names case-insensitively, e.g. using `g_utf8_strdown()` and such. I am not sure, how good `glib` is for this task. In Guile, for example, `scm_{from|to}_locale_string()` functions, not *utf8* ones, are usually used for processing of *file names*, which is TRT, I believe.

graahnul-grom:
Good point, thank you for the hint.
DOS and its malicious descendants... :-)
I've made dialogs work with `*.[sS][cC][hH]` and `*.[sS][yY][mM]`,
please check if it is OK now.

vzh:
It's OK with me, thanks.

noqsi:
DOS and its malicious descendants?

Nah, DOS and its malicious [ancestors](https://en.wikipedia.org/wiki/Compatible_Time-Sharing_System#File_system)!

graahnul-grom:
:)

PR #657, opened 2020-06-22
Fix coverity scan issue 204294 - set_guile_compiled_path()

graahnul-grom:
`set_guile_compiled_path()` in `liblepton/src/liblepton.c`:

Coverity scan analysis ([issue 204294](https://scan9.coverity.com/reports.htm#v39680/p11510/fileInstanceId=66025513&defectInstanceId=7729048&mergedDefectId=204294))
rightfully warns that the value of the `$GUILE_LOAD_COMPILED_PATH`
environment variable (obtained by `getenv()`) could contain anything
and should be checked somehow.
Moreover, existing code assumes that variable's value is small
enough to fit into a buffer of size `PATH_MAX` (typically, 1024).

- Dynamically allocate a large enough string to hold both Lepton
pre-compiled path and existing contents of the
`$GUILE_LOAD_COMPILED_PATH` variable.
- Check if that value consists of valid UTF-8 characters with
`g_utf8_validate()`.

The question is: what else can we do to ensure that env var value
is correct? check its size? if so, what size should be considered
as "maximum"?

### DIFF
--- a/liblepton/src/liblepton.c
+++ b/liblepton/src/liblepton.c
@@ -57,31 +57,38 @@ void liblepton_init(void)
 
 
 
-/*! \brief Add Lepton compiled path to Guile compiled paths env var.
+/*! \brief Prepend Lepton compiled path to Guile compiled paths env var.
  *  \note  To take effect, must be called before scm_boot_guile().
  */
 void
 set_guile_compiled_path()
 {
-  char* path = getenv ("GUILE_LOAD_COMPILED_PATH");
-  char buf[ PATH_MAX ] = "";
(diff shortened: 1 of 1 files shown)


vzh:
I'm not sure the second patch is relevant here.  Guile, AFAIK, uses locale encoding for path processing, so the same should be done for env vars defining paths.  If something is wrong with path, it should let the user know of it.  What `g_utf8_validate()` should do with paths, e.g. in the `cp1251` encoding?

graahnul-grom:
You're right. Removed last commit. Then, I have no idea how to
(and should we) check that path. Guile will do it, anyway, IIUC.

PR #659, opened 2020-06-25
Transform lepton-schematic code into a new GUI library, libleptongui

vzh:
Now, `lepton-schematic` is a Scheme script, allowing to play with it without recompilation.  It uses Guile FFI to call for C functions without modifying them.  Another innovation is using of the module `(srfi srfi-37)` for command line option processing.

I've tested precompilation, all command line options, and anything seem to work OK.  However, the code needs more testing on various platforms.

Hopefully, the new library will be reduced to a proper minimum and merged with `libleptonattrib`.


schematic/lepton-schematic.scm
```diff
@@ -0,0 +1,379 @@
...
+                              (lepton version)))
+
+
+(define (process-error-stack stack key args)
+  (match args
+    ((subr message message-args rest)
+     ;; Capture long error message (including possible backtrace).
+     (let ((long-message
+            (with-output-to-string
+              (λ ()
```
graahnul-grom, line 192: Excuse me, it's a bit much. As if crazy names like
`make-picture/vector` is not enough. :-(

`$ cat lambda-facepalm.scm`
```scheme
( define λ—это_стильно,модно,молодёжно! "facepalm" )
( define π-ζδϵτσ "excuse me my Greek" )

( define ( функция-шмункция )
  ( format #t "функция-шмункция: ~a~%" функция-шмункция )
  ( format #t "вы будете смеяться, но и это тоже работает.~%" )
  ( format #t "λ—это_стильно,модно,молодёжно! [~a]~%"
               λ—это_стильно,модно,молодёжно! )
  ( format #t "π-ζδϵτσ [~a]~%" π-ζδϵτσ )
)

( функция-шмункция )
```

```
$ guile lambda-facepalm.scm
функция-шмункция: #<procedure функция-шмункция ()>
вы будете смеяться, но и это тоже работает.
λ—это_стильно,модно,молодёжно! [facepalm]
π-ζδϵτσ [excuse me my Greek]
$ echo $?
0
```

graahnul-grom:
Pre-compilation works, except for `lepton-netlist`, though it is
not related to changes in this PR. Unfortunately, I can't remember
when exactly it started (not a long time ago), but guile cannot
longer produce working code for `lepton-netlist` in advance.
`make precompile` is OK, but all invocations (with any backend)
fail with random unrelated errors. Deleting `ccache/netlist/`
fixes it. Sometimes `ccache/netlist.go` must also be removed.
I tried different things (modifying pre-compile script), walking the
directory tree "depth first", compiling components in particular
order, selectively, etc. - without success.
Maybe, there are some circular dependencies in netlister code?

graahnul-grom:
> Hopefully, the new library will be reduced to a proper minimum and merged with libleptonattrib.

I wouldn't do that. `attrib`'s code is old, large and contains 3-rd party
components.
```
$ wc -l libleptonattrib/src/gtksheet_2_2.c 
    8741 libleptonattrib/src/gtksheet_2_2.c
```

As an application, it is unusable.
Let's keep it in a separate library.

graahnul-grom:
As with `lepton-attrib`, `liblepton-schematic` works only
when `$LD_LIBRARY_PATH` is explicitly set.

graahnul-grom:
Please run `git status` after `make`: some files should be added
to `.gitignore`.

schematic/lepton-schematic.scm
vzh: Oh, come on :-) λ macro is available in guile core since 1.9.2, included by popular request. Many people like it, it looks prettier and the code is shorter. Emacs has a convenient shortcut for insertion it.  You see, I do not use any other non-ASCII letters in the code.

vzh:
> Something has changed in the initialization sequence: when
> `GschemToplevel` is constructed, configuration is not
> loaded yet, and `renderer_load_font()` fails. Thus, setting
> custom schematic font has no effect.

Good point. I'll fix it soon.

vzh:
> Maybe, there are some circular dependencies in netlister code?

Probably. Hopefully, I'll find some time to look into this.

vzh:
> As with `lepton-attrib`, `liblepton-schematic` works only
> when `$LD_LIBRARY_PATH` is explicitly set.

Since a new library was introduced, `sudo ldconfig` should be run to fix it.  At least, this works for me on Debian.

vzh:
> > Hopefully, the new library will be reduced to a proper minimum and merged with libleptonattrib.
> 
> I wouldn't do that. `attrib`'s code is old, large and contains 3-rd party
> components.
> 
> ```
> $ wc -l libleptonattrib/src/gtksheet_2_2.c 
>     8741 libleptonattrib/src/gtksheet_2_2.c
> ```
> 
> As an application, it is unusable.
> Let's keep it in a separate library.

I wish to make the code usable, and get spreadsheet *inside* `lepton-schematic`. Let's see...

vzh:
@graahnul-grom, issues with `.gitignore` & rendering font should be fixed in last two commits.

graahnul-grom:
> I wish to make the code usable, and get spreadsheet inside
lepton-schematic. Let's see...

Please, no! It is that very "spreadsheet" parody which is unusable.
It's not worth bloating lepton-schematic codebase with almost nine
thousand lines of ancient ugly code, increasing it by 30%.
It's not worth our time and effort, we have many much more important
things to do. I would understand your enthusiasm if we were bombarded
with user requests for this particular feature and had nothing else
to spend our time on. Even then, we'd better make a spreadsheet using
`GtkTreeView` from scratch. And it's far from obvious if table-like
representation is the best way to edit multiple attributes.

schematic/lepton-schematic.scm
graahnul-grom: > included by popular request

...received from kindergartens, fair enough. :-) They just
don't know there are even more fancy-pancy looking letters, `Ы`, for example.
And every technically inclined young person is well aware that code with
embedded anime pictures is faster and more reliable: this vital feature
is already in active development, I'm sure.
It is evident that replacing `lambda` with `λ` has zero practical benefits.
The only result we'll get is damaged reputation.
If at all possible, please let's stick to ASCII. Thank you.

vzh:
I got your point.  I was not going to merge the code bases as is.  Some GUI functions/dialogs may be unified and reused just now, though.  For me, `libleptonattrib` is a playground for working with Scheme FFI for now.  I'll report if I get something useful, of course ;-)  And yes, my ultimate goal is using of recent (third party) GTK libraries in Scheme with FFI instead of what we have now.  The road is long and the task is not of much priority, so I don't think you'll see anything useful soon :-)

schematic/lepton-schematic.scm
vzh: In such a case, Guile, Guix, and other Guile-based projects already have such a bad reputation.  The next step would be to avoid using all such programs.  Don't you know that `lambda` (I'm avoiding using of that horrible UTF-8 letter, probably we have to disable UTF-8 in our project at all) is about *lambda-calculations*?  (Now, after thinking of it a bit more I would prohibit TeX and mathematics completely in Lepton as well, if I could.) You probably aware of some `Ы`-calculations and some programs that we could use instead of Guile?  Let's do it, I'm all for it :-)

Jokes aside (while they're sometimes sad), well OK, apart from this issue, if I'll use just plain `lambda`, are there any other issues in the code that prevent merging it?

graahnul-grom: Please, don't get me wrong. I'm trying to point out that this
particular innovation at this point in time has zero practical
benefits and might bring negative impact. That's all.

I'm not against Unicode or Guile or whatever else the (unfounded)
generalization was made about. The problem lies on a completely
different plane.

Imagine users watching our project. It's evident that there is *huge*
amount of work to be done (suffice it to say, we do not have decent
documentation). In this situation, developers who busy with frills,
bells and whistles couldn't make a great impression. I, for one,
would certainly be alarmed by seeing that.

I've tested the recent changes, configuration is loaded correctly.
I think we could merge it. Thank you.

graahnul-grom:
@vzh I would propose to use the `gui: ` prefix in git messages
when committing changes to `libleptongui` code.
It would allow us to be more expressive in remaining 50 - 5 - 1 = 44
characters, what do you think? :-)

schematic/lepton-schematic.scm
vzh: It's not about bells and whistles. It's just convenient and quick and I used this in new code only. Fair enough, let's stop this.
Thank you for the review!

vzh:
> @vzh I would propose to use the `gui: ` prefix in git messages
> when committing changes to `libleptongui` code.
> It would allow us to be more expressive in remaining 50 - 5 - 1 = 44
> characters, what do you think? :-)

Agreed :-) I think I'll use `gui` for C code and `schematic` or other module namespace for Scheme.

graahnul-grom:
>  schematic: Get rid of TOPLEVEL in x_window_new() and main_prog().
The change leads to loading of RC-files many times, for each new
toplevel window.  Should it be so?

Both new and old configuration files are loaded only once, even if
`x_rc_parse_gschem()` is called for each new window (`liblepton/src/g_rc.c`).

Please note that looking at the "code" in that file for more than 2-3
minutes could lead to a heart attack. I am dead serious. Take care.

graahnul-grom:
One thing I overlooked is the tabbed GUI initialization.
`x_tabs_init()` should be called after the configuration is read,
but before new window construction (`x_window_setup()`).
Now `schematic.gui::use-tabs` and all settings in the
`[schematic.tabs]` group are ignored.
The attached patch should fix it.

```diff
diff --git a/libleptongui/src/lepton-schematic.c b/libleptongui/src/lepton-schematic.c
index 281240052..fa1e3f741 100644
--- a/libleptongui/src/lepton-schematic.c
+++ b/libleptongui/src/lepton-schematic.c
@@ -142,17 +142,14 @@ main_prog (SCM file_list_s)
   /* Set up atexit handlers */
   gschem_atexit (i_vars_atexit_save_cache_config, NULL);
 
-  /* Now read in RC files. */
+  /* Parse custom GTK resource files: */
   g_rc_parse_gtkrc();
 
   /* Set default icon theme and make sure we can find our own icons */
   x_window_set_default_icon();
   x_window_init_icons ();
 
-  /* Initialize tabbed GUI: */
-  x_tabs_init();
-
-  /* Allocate w_current */
+  /* Create a new window and associated TOPLEVEL object: */
   w_current = x_window_new ();
 
   /* Enable rendering of placeholders */
diff --git a/libleptongui/src/x_window.c b/libleptongui/src/x_window.c
index 7dd1b5d11..9fdc38cf3 100644
--- a/libleptongui/src/x_window.c
+++ b/libleptongui/src/x_window.c
@@ -991,12 +991,12 @@ x_window_init_icons (void)
  */
 GschemToplevel* x_window_new ()
 {
-  GschemToplevel *w_current;
   TOPLEVEL *toplevel = s_toplevel_new ();
 
+  /* Load old (*rc files) and new (*.conf) configuration: */
   x_rc_parse_gschem (toplevel, NULL);
 
-  w_current = gschem_toplevel_new ();
+  GschemToplevel *w_current = gschem_toplevel_new ();
   gschem_toplevel_set_toplevel (w_current, toplevel);
 
   /* Damage notifications should invalidate the object on screen */
@@ -1004,6 +1004,9 @@ GschemToplevel* x_window_new ()
                        (ChangeNotifyFunc) o_invalidate,
                        (ChangeNotifyFunc) o_invalidate, w_current);
 
+  /* Initialize tabbed GUI: */
+  x_tabs_init();
+
   x_window_setup (w_current);
 
   return w_current;
```

[fix-tabgui-cfg.diff.txt](https://github.com/lepton-eda/lepton-eda/files/4873757/fix-tabgui-cfg.diff.txt)

vzh:
> > schematic: Get rid of TOPLEVEL in x_window_new() and main_prog().
> > The change leads to loading of RC-files many times, for each new
> > toplevel window.  Should it be so?
> 
> Both new and old configuration files are loaded only once, even if
> `x_rc_parse_gschem()` is called for each new window (`liblepton/src/g_rc.c`).

This has been fixed in the next commit (39df6e7).

> Please note that looking at the "code" in that file for more than 2-3
> minutes could lead to a heart attack. I am dead serious. Take care.

COVID-20 ? 8-0

vzh:
> One thing I overlooked is the tabbed GUI initialization.
> `x_tabs_init()` should be called after the configuration is read,
> but before new window construction (`x_window_setup()`).

As the branch has been already merged, I'm waiting for a new PR with fixes :-)

PR #660, opened 2020-06-25
Deprecate legacy gschem subroutines

vzh:
In this branch, most of legacy subroutines defined straight without `SCM_DEFINE` and replaced by actions are moved to the module `(gschem deprecated)`.
`gschem-exit` is one of them :-)


vzh:
I'm still waiting for your attention, lepton devs :-)  I have another branch depending on this changes, which resolves some C <=> Guile interdependencies in lepton code.  The idea is to optimally separate C and Guile code, which would allow `guild` compilation and facilitate native builds on some unfriendly but popular OSes.

graahnul-grom:
Could you please rebase the changes so I can test them (there are
conflicts after #659)?

vzh:
Done.

graahnul-grom:
> native builds on some unfriendly but popular OSes

MINGW? We [discussed](http://www.delorie.com/archives/browse.cgi?p=geda-user/2019/09/27/09:15:55) it on the geda-user mailing list.
The problem is that guile [still cannot be built](https://github.com/msys2/MINGW-packages/issues/3298) natively.

vzh:
It can. And there are pre-built guile versions for Windows (e.g. by Eli Zaretsky). It's just devs in the project *mingw-w64* don't want to use unreleased (patched, non-stock) versions. Previously, we built it using plain *mingw* (*mingw32*), and it worked.  AFAIK, people from the *guile-gi* project (Mike Gran) build Guile for amd64 as well.  BTW, he is the author of necessary patches for Guile 2.2, and maybe 3.0, and his patches may still live in one of guile's wip branch.  Anyways, plain Guile (Scheme) code requires nothing but working Guile, and allows to work with third-party C libraries.  It's the main idea: make our tools build without Guile (snarf etc.), use FFI, and just launch them in Windows using a pre-built Guile.  I'm working on it, and it's interesting :-)

graahnul-grom:
I'm not acquainted with Eli and Mike, direct links would be helpful.
Could you please explain me the connection between using/not using
guile during build, FFI and native Windows build.
From your message one can conclude that it's possible to build
guile (and lepton-eda?) on Windows natively. If so, it would be
great if you add a wiki page with detailed instructions and
announce it on the geda-user mailing list (in the thread I
linked above, where nobody succeeded in doing it).
BTW, I cannot build lepton under Cygwin after all that library
conversion stuff, did you try to do it recently? Even 1.9.11
release fails to build. I've tried to `--disable-attrib` but it
doesn't work anymore.

vzh:
I am sorry, I've misspelled one of the names: Eli Zaretskii.

His project, [ezwinports](https://sourceforge.net/projects/ezwinports/files/), contains many binaries (with sources) rebuilt using plain MinGW (mingw.org's 32-bit version, not 64-bit one).

[Guile-gi](https://github.com/spk121/guile-gi) is a project by Mike Gran.  Look through its `README` and `configure.ac`, and you'll find they support Windows.  So there is a hope... Guile-gi itself could be of interest for our project.

Please also see [this discussion](https://www.mail-archive.com/guile-devel@gnu.org/msg14709.html) for the issues when compiling guile with mingw-w64.  Some notable links from there:

- [the branch by Mike Gran](http://git.savannah.gnu.org/cgit/guile.git/log/?h=wip-mingw-guile-2.2) which fixes several issues and should build on Windows

- [these patches on GH](https://github.com/mkeeter/guile-mingw),  the author says in README it mostly works, there are only some  issues with building of documentation. I don't know, though,  which guile version was used there.

- The [bug report](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=22406) by @peter-b sheds some light on what's wrong with Guile code on Windows.


There is also our [minipack](https://github.com/vzh/minipack) project.  I cross-compiled the whole gEDA suite with it a few years ago.  Tried to repeat it on my recent Debian with newer mingw utilities these days with no success: I stuck on the stage of gettext compilation :-( No Guile involved yet.


I didn't try to build on Cygwin. (Didn't use Windows at all for quite a while.)

About FFI.

I suspect if our C code didn't use libguile functions, only guile binary as an interpreter would be needed to make the tools work on Windows, even prebuilt one by somebody else.  (We have to switch to using FFI anyway since SMOBs are deprecated and will be thrown away in some of upcoming Guile versions, but this is another story...)

graahnul-grom:
Thank you.

vzh:
BTW, many fixes by Mike Gran (@spk121) wrt Windows/Cygwin support have been included in Guile 3.0.  We have to think about supporting it, too.  The *testing* branch of the Debian distributive already includes that version. That's a sign ;-)

graahnul-grom:
I've managed to build Guile 3.0.4 (from git) on FreeBSD, add 
support for it to lepton and compile it.
Got one compiler (clang) warning that `scm_t_int8` is deprecated
(liblepton/src/scheme_object.c:2193).
At run-time there are many (20-30) warnings like this:
```
WARNING: (module name): imported module (lepton core gettext) overrides core binding `_'`
```

After setting env var `GUILE_WARN_DEPRECATED=detailed`
I see 2 other warnings:
```
Passing a string as a type-name to make-record-type is deprecated.
  Pass a symbol instead.
```
and
```
Calling `record-constructor' with two arguments (the record type
 and a list of field names) is deprecated.  Instead, call with just
 one argument, and provide a wrapper around that constructor if
 needed.
```

`make precompile` is OK, but issues tons of warnings like above,
about "possibly unbound variable" and "non-literal format string",
the latter for localized strings like `(_ eklmn)`.

vzh:
In 3.0, guilers decided to use `G_` for translations in Scheme code, I think mostly because `_` is used in `match` and syntax transformers.  I've started such transition in our project as well. Need to fix this anywhere in the code.  Hopefully, it's easy.  Other issues should be addressed as well.

vzh:
@graahnul-grom, please contribute your changes wrt support of Guile 3.0.
And, IMO, `guile.m4` is not required.  On correctly configured systems m4 macros from Guile itself should work out of the box (iskaropki) :-).

graahnul-grom:
#667.
I'm not sure about guile.m4, it helps when multiple versions are installed.
In practice, a correctly configured system is not always the case (e.g. #463).

vzh:
IIUC and/or IIRC, we just didn't used stock Guile macros from `guile.m4` at that time.  IMO now, any release of Guile (at least since 2.0) includes the file with suitable defaults and its macros can be used by other programs via autotools.  If it is not the case on some systems, I believe they are just not properly configured.

graahnul-grom:
I've just tried to remove `guile.m4` file. Then I'm unable to `configure`
with Guile 3.0 while 2.2 package is installed.
It says "error: found development files for Guile 2.2, but /tmp/DMN/gu3/bin/guile3 has effective version 3.0".
With `m4/guile.m4` file in place, I have no problems with either version.
Why? `$PKG_CONFIG_PATH` is set correctly.

PR #666, opened 2020-07-11
Fix the --disable-attrib configure option

graahnul-grom:


graahnul-grom:
https://github.com/lepton-eda/lepton-eda/commit/a85d470

It is a very useful option! :-)

vzh:
So we may just drop it: less options, less code, less work :-)

graahnul-grom:
Only along with lepton-attrib itself. That would be *really*
"less code, less work". :-)

PR #668, opened 2020-07-15
Get rid of the Scheme module (lepton core version) defined in C

vzh:
The code of the module `(lepton core version)` has been rewritten in Scheme using FFI. The following changes have been made:

  - `%lepton-version()` is no longer used in the code.
  - New C functions have been added to get available version info from C code without mucking up with shell utilities and substitution scripts.  Delayed evaluation (*lazy* approach) is used to get that data in Scheme code *once*.
  - Added procedure `lepton-version-data()` which can return the list of version strings by requested symbol names.
  - Added procedure `lepton-version-ref()` which retrieves one version data element by its symbol name, that is, exactly repeats the previous behaviour of `lepton-version()` when it was invoked with one symbol argument.
  - `lepton-version()` has been redefined in such a way that without arguments it returns a predefined string usually output by Lepton tools. When it is called with arguments, the first of them must be a format string to format output by means of `simple-format`, and all others must be predefined symbols used to retrieve necessary data the same way as for the procedures above.
  - General copyright info has been added to the list of available version data, which allows uniform output of copyright info.
  - The general `msg` symbol has been removed from the list of symbols for retrieving version info. `lepton-version()` without arguments does it instead.

Additional changes include:
- Code of all Lepton tools wrt getting version information has been reduced and simplified.

- Added documentation on functions of the module `(lepton version)`.

- Added tests for functions of the module `(lepton version)`.

- Improved, added, or fixed docstrings of several Scheme procedures.

- Obsolete procedure `gschem-version()` written in C has been deprecated and removed from C code along with its helper function `g_rc_rc_filename()`.

- Gettext domain definition has been moved from `lepton-schematic` to `(schematic core gettext)`.

- `lepton-schematic-version` has been moved to a specially crafted module `(schematic version)` and renamed to `print-lepton-schematic-version`.


graahnul-grom:
```
$ git grep -n lepton_version_message
liblepton/include/liblepton/prototype.h:261:char* lepton_version_message();
```

vzh:
First issue has been fixed, thank you for the point.
As for the second one, it will be fixed in an upcoming PR where I simplified/reduced the code a bit more :-)

Issue #653, opened 2020-06-11
Document compability


spth:
I am currently using gschem 1.8.2, the latest gschem version in Debian. Apparently the geda-gschem / geda-gaf maintainer moved to lepton, so (like probably many pre-fork gschem users) wonder:

* Will I be able to work with old gEDA files using lepton (I guess yes)?
* Will I be able to work with past-fork gEDA files using lepton (I have no idea)?
* Will I be able to work with lepton files using post-fork gEDA (I have no idea)?

It would be good to have a section on compability in the README file.

graahnul-grom:
Lepton EDA and gEDA uses the same file format, nothing has changed
in this regard, so
1) yes
2) yes
3) yes

Thank you for the good suggestion, we will add notes on compatibility
to the README file.

vzh:
One small addition to what @graahnul-grom has already said.  So far, we supported gEDA file format and are planning to do so in future since there are lots of symbols/schematics created using it.  If there will be some additions on the `geda-gaf` side (e.g., wrt attributes, or new file format version, and such), it depends on users of Lepton, on you, to point us to what you think is not supported but should be.  The same is true wrt other features in `geda-gaf` Lepton may not have.  We are not able to fully keep track of that project, so again, we depend on our users here and need your feedback.

Issue #284, opened 2018-12-02
Improve documentation


graahnul-grom:
1) Most of the `README` file content should be moved to
a new `INSTALL` file. This file should also include a link
to the `Lepton EDA and Cygwin` wiki page.
2) `README` should be extended to provide more information
about the `Lepton EDA` suite:
  - it was forked by main `geda-gaf` developers
  - it is compatible with (and direct replacement for) `geda-gaf`
  - list of new features: link to `NEWS` file
  - what systems are supported
  - binary packages available
  - how to build it (link to `INSTALL` file)
  - link to (wiki) page with screenshots (maybe embed 1-2 screenshots
in `README` itself)
3) Some documents should be included in the installation
(`share/doc/lepton-eda/`):
  - `README`
  - `NEWS`
  - `Configuration settings`, `Attributes` wiki pages (converted to HTML (?))
4) A new `Configuration` (HTML) document should be added;
add a new `Help->Configuration` menu item for it.
This document should contain:
  - description of `component-library` and `source-library` options and
where to place them. This is the most misunderstood subject. Wiki page
(`wiki/geda-gschem_ug-config.html`) is not clear enough on that.
  - links to all files (local wiki, man pages, maybe web pages)
with information on how to configure various tools in the `Lepton EDA` suite.

graahnul-grom:
One more item to add to the `README` file:
- `lepton-eda` can easily be extended through the rich (and very well documented) `Scheme` API (`GNU Guile`).

I'd also consider adding `guile` to the list of "topics" (keywords shown at the top of the `github` repo page).

vzh:
I've added `guile`, as well as `c` and `geda` to the topic list.

graahnul-grom:
I think it's also worth installing the Scheme API reference manual HTML files.
We already installing HTML man page, so why not?

vzh:
Agreed. While I myself like plain info documentation, some users could prefer HTML.

graahnul-grom:
Having four help items in the `Help` menu is excessive.
They all point to essentially the same outdated wiki documentation
pages, one click away one from the other.
I think just one item (`gEDA Wiki`) would be enough.

vzh:
The menu should be reviewed, I believe. We have to somehow refactor the old docs from gEDA wiki and merge it with our up-to-date documentation (if any). I think we could move the docs first to our wiki, and then probably to another repository.

graahnul-grom:
[...]

PR #674, opened 2020-07-28
Started creation of Lepton EDA reference manual

vzh:
Affects #284 
Affects #653

Added little overview and converted our wiki document about configuration into *Texinfo*.
In `lepton-schematic` the reference can be opened with `h m`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,8 +65,9 @@ Installation
 Dependencies
 ------------
 
-In order to compile gEDA from the distributed source archives, you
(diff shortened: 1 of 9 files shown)


vzh:
Thanks.
Let's get started and add info piece by piece.  This is probably a long road, but we've made a first step eventually :-)
The next steps, I think, should be converting of other valuable contents of our wiki and fixing up the above mentioned issues.

PR #689, opened 2020-10-14
Fix two issues in the tEDAx backend

### LINKED ISSUE #686: Spaces in tEDAx output strings from lepton-netlist should be escaped
Hi,
I am just about to package lepton-eda for the Mageia distro, for which I am the maintainer of pcb-rnd.
I have started to use lepton-eda myself and so far it is behaving well. By that I mean not crashing and losing work as has been happening for years with gschem!

However I have hit the above issue, where parametric footprints (in particular) which contain spaces between the parameters are passed un-escaped from the exporter resulting in missing footprints in pcb-rnd.

Since the main object of adding lepton-eda to our repository is for use in partnership with pcb-rnd as an alternative to gschem/pcb it would be great if this could be resolved.

Thanks,
Barry

vzh:
- Fix escaping of whitespaces as reported in #686 (Closes #686)
- Fix output of `nettag` attributes (reported on the *geda-user* mailing list)

### DIFF
--- a/netlist/scheme/backend/gnet-tEDAx.scm
+++ b/netlist/scheme/backend/gnet-tEDAx.scm
@@ -21,8 +21,12 @@
 ;; The tEDAx format is documented at http://repo.hu/projects/tedax/
 ;; --------------------------------------------------------------------------
 
-(use-modules (netlist schematic)
-             (netlist schematic toplevel))
+(use-modules (srfi srfi-1)
+             (lepton attrib)
+             (lepton object)
+             (netlist schematic)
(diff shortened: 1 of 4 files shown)


barjac:
Thanks, Looks fine now, but please see my comment in the "Spaces" bug before closing.

vzh:
Umm, well, let's apply the fixes now, and if you feel the issue with trailing spaces is really a bug, please open a new issue.  I'm not sure if this should be fixed for an only backend or for all attributes processing.

PR #692, opened 2020-11-18
Allow changing of net colors in lepton-schematic GUI

graahnul-grom:
Allow users to change a color of net objects
via the standard color selection combo box of the
"Object Properties" dialog in lepton-schematic GUI.


vzh:
Thanks!

Issue #687, opened 2020-10-09
Colour scheme changes saved to file linked in gschemrc are ignored


barjac:
I have edited the colour scheme and saved to a file, /home/baz/Geda/colours
I have created ~/.config/lepton-eda/gschemrc
containing only:
primitive-load "/home/baz/Geda/colors"
The colour code I changed, (just for 'net' at present) is correct in the 'colours' file.
What am I doing wrong?
This gschemrc is being read OK as any syntax errors in it stop the program launch.
I am using the dark theme, is this relevant?
Dark blue net on black b/g is impossible to use for me.

Thanks,
Barry

Edited....
Log messages (don't know where the date comes from but this was from git master on 5th October) git# is correct:

Lepton EDA/lepton-schematic 1.9.11.20200604 (git: 6e70b0c)

Loaded RC file [/usr/share/lepton-eda/system-gafrc]
Loaded RC file [/usr/share/lepton-eda/system-gschemrc]
Loaded RC file [/home/baz/.config/lepton-eda/gschemrc]

Loading schematic [/home/baz/Geda/TIMER3/timer3.sch]

vzh:
Hi Barry,
Could you please provide us with the contents of your files `gschemrc` and `colours` or just attach them here?  What value have you changed and how?

barjac:
Hi,
This is the colors file content:
`[baz@jackodesktop Geda]$ cat colors
(display-color-map '(
  (background           "#000000")
  (pin                  "#ffffff")
  (net-endpoint         "#ff0000")
  (graphic              "#00ff00")
  (net                  "#97ff05")
  (attribute            "#ffff00")
  (logic-bubble         "#00ffff")
  (dots-grid            "#bebebe")
  (detached-attribute   "#ff0000")
  (text                 "#00ff00")
  (bus                  "#00ff00")
  (select               "#ffa500")
  (bounding-box         "#ffa500")
  (zoom-box             "#00ffff")
  (stroke               "#e5e5e5")
  (lock                 "#bebebe")
  (output-background    #f)
  (freestyle1           "#308a8e")
  (freestyle2           "#d02f2f")
  (freestyle3           "#b93ac1")
  (freestyle4           "#cccccc")
  (junction             "#ffff00")
  (mesh-grid-major      "#1e1e1e")
  (mesh-grid-minor      "#171717")
))

(display-outline-color-map '(
  (background           #f)
  (pin                  "#ababab")
  (net-endpoint         "#00ab00")
[...]

PR #693, opened 2020-11-18
Clarify the color editor widget message

graahnul-grom:
Indicate that parentheses must be included in the
`gchemrc` command for loading a color scheme file.
Experience shows (#687) that people who don't acquainted
with Lisp might get it wrong.


vzh:
Thanks!

PR #700, opened 2020-11-27
Fix the (schematic attrib) and (schematic menu) modules

graahnul-grom:
Functions used in these modules are exported
from `liblepton`, not `libleptongui`:
- `s_attrib_uniq()`
- `s_attrib_add_entry()`
- `s_menu_add_entry()`

Change `dynamic-func()` calls accordingly,
use proper module: `(lepton ffi)`.


vzh (approved):
Thanks!

PR #701, opened 2020-11-29
Merge libleptonrenderer into liblepton

graahnul-grom:
Source code of either library cannot be used without
the other's library code. Both are strongly interdependent.
Artificial "separation" as it is done now, makes no sense
whatsoever. Sure enough, such a "brilliant" design leads
to nothing, but problems.

Since release 1.9.10 it's impossible to build Lepton under
`Cygwin`. Eliminating unnatural separation of code allows
us to [repair](https://github.com/lepton-eda/lepton-eda/issues/690#issuecomment-735249953) a broken `Cygwin` build.

Packagers report that the above nonsense breaks build
with certain flags, e.g. `-Wl,--no-undefined` (#690).
More problems to come, that's for sure.

Move libleptonrenderer's source code and header files
to liblepton, adjust the build system accordingly.
Later, header files should be reorganized and renamed.

Closes #690

PR #703, opened 2020-12-05
Make Lepton work under Cygwin again

graahnul-grom:
Libraries loaded by `dynamic-load()` Guile function
should be named differently under Cygwin.
Adjust the build system so that, depending on the platform,
appropriate names are substituted.

It is now possible to build Lepton under Cygwin
(which was fixed in #700 and #701) and run
the GUI tools. I've even managed to compile it
with GTK3 (though `--without-attrib`).

What have to be done yet, is parsing of the
`xxx_SHLIB_VERSION` strings, because a cygwin
library name has the major version number appended.
For example, if `LIBLEPTON_SHLIB_VERSION='7:0:0'`,
Cygwin name should be `cyglepton-7`.
It's better to automate it, so we don't have to change
Cygwin names with every release.

### DIFF
--- a/liblepton/scheme/.gitignore
+++ b/liblepton/scheme/.gitignore
@@ -2,4 +2,6 @@ Makefile
 Makefile.in
 test-suite.log
 lepton/core/gettext.scm
+lepton/ffi.scm
+lepton/log.scm
 *~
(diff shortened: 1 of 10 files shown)


vzh (approved):

vzh:
@graahnul-grom, could you please fix our wiki page about Cygwin?

graahnul-grom:
Yes.

graahnul-grom:
Done. Please check.

vzh:
Thank you!

Issue #696, opened 2020-11-22
Cannot execute actions via lepton-schematic command line (-c)


graahnul-grom:
For example (the same result with other actions):

```
# lepton-schematic -c '(&file-quit)'
ERROR: Found invalid lepton-schematic window smob #f
Please see log file for more information.
```

In the log window:
```
Backtrace:
          10 (primitive-load "/home/dmn/lepton/bin.tb/bin/lepton-sch…")
In ice-9/boot-9.scm:
    829:9  9 (catch #t #<procedure 80e957760 at ice-9/eval.scm:330:…> …)
In unknown file:
           8 (eval (begin (eval-string "(&file-quit)")) #<directory …>)
           7 (eval-string "(&file-quit)" #<undefined>)
In ice-9/boot-9.scm:
   2312:4  6 (save-module-excursion #<procedure 80e966380 at ice-9/e…>)
In ice-9/eval-string.scm:
     38:6  5 (read-and-eval #<input: string 80eb78070> #:lang _)
In ice-9/eval.scm:
    619:8  4 (_ #(#(#<directory (schematic action) 80b8d05a0> # …) #))
In unknown file:
           3 (%file-quit)
In ice-9/boot-9.scm:
   751:25  2 (dispatch-exception 0 misc-error (#f "Found invalid …" …))
In ice-9/eval.scm:
   626:19  1 (_ #(#(#<directory (lepton eval) 802f048c0> #<vari…>) …))
In unknown file:
           0 (make-stack #t)

Found invalid lepton-schematic window smob #f
```

It works in 1.9.11 release, but is broken in the master branch.

vzh:
Yeah, in the new code, window is not initialized if you use the `-c` option.  IIRC, lepton-schematic always exited after executing commands specified with `-c`.
Do you have any real life example where you would need using of actions without opening the main lepton-schematic window?

graahnul-grom:
> lepton-schematic always exited after executing commands specified with -c

That's not correct. Please try `-c` with 1.9.11 release.

graahnul-grom:
`-c` is a convenient way to make lepton-schematic execute
some guile script on startup.

vzh:
Ah, OK, I see.  Sorry for the noise.

PR #704, opened 2020-12-06
Fix processing of lepton-schematic -c and -s command line arguments

graahnul-grom:
@vzh Could you please explain (in technical terms)
to the general public, why does it fix the issue (#696)?
Taking into account [comments](https://github.com/lepton-eda/lepton-eda/blob/ea15637e5f3949fc6ed9aa615608e038dd895c70/libleptongui/src/g_window.c#L98)
in `g_dynwind_window()`.

Maybe a rush into FFI-ing and DBCS-ing everything
left and right, without considering whether it's
suitable or not in each particular case, was not
that good idea, after all? :-)
Here I can see (IMHO, of course) an example of,
so to say, a "refactoring backwards": instead of
keeping `lepton-schematic` initialisation code
together, it is now randomly scattered between C
and Guile code (with inconsistently named functions,
BTW), which is executed in different contexts.
Isn't it dangerous, WDYT?

### DIFF
--- a/libleptongui/src/lepton-schematic.c
+++ b/libleptongui/src/lepton-schematic.c
@@ -137,8 +137,6 @@ main_prog (SCM file_list_s)
 
   cwd = g_get_current_dir();
 
-  scm_dynwind_begin ((scm_t_dynwind_flags) 0);
-
   /* Set up atexit handlers */
   gschem_atexit (i_vars_atexit_save_cache_config, NULL);
 
@@ -210,5 +208,4 @@ main_prog (SCM file_list_s)
     x_widgets_show_log (w_current);
   }
 
-  scm_dynwind_end ();
 }


vzh:
Please look at the new comments in lepton-schematic that I pushed in #712.
Yes, working with fluids in C requires using of `scm_dynwind_begin` and `scm_dynwind_end`, as specified in the Guile reference manual.  Removing those guards may lead to some dangerous behavior, while fixes this particular case, due to multithreading essence of Guile.

vzh:
About refactoring backwards and inconsistency:
This was my WIP interrupted by some claims to fix many other issues.
I tried to find the way to get rid of SMOBs, deprecated in Guile 3.0, and move low-level Scheme code out of Lepton libraries.  Seems I've found it (thanks to your nice criticism ;-)).
Another issues I try to solve now is proper compilation of Scheme code (without various tricks like pre-compilation or using of `primitive-eval`), getting rid of RC parsing code on C side, and some means of IPC (Guile callbacks working with Gtk signals).

Well, DBCS is something we discussed before, thank you again ;-)

Please check if #712 solves the issue and works at all for you.

graahnul-grom:
Exactly. It looks like a work interrupted somewhere midway.
Good to know that is indeed so. :-)
I'll check #712 a bit later today.
Thank you.

graahnul-grom:
Yes, #712 did it. Thank you!
Perhaps, `scm_t_bits window_smob_tag`
could be removed from `g_window.c`.

vzh:
I was sure I had removed it completely,  I'll try once again :-)

PR #705, opened 2020-12-06
Make lepton-attrib prompt to save changes on exit

### LINKED ISSUE #702: lepton-attrib doesn't prompt to save changes on exit
`lepton-attrib` built from the current master branch
(git: 716a01f) does not prompt to save changes if you:
- `Edit->Add new attrib column`
- Edit existing values

However, the prompt is displayed if you:
- `Edit->Delete attrib column`
- Select one of the items in the `Visibility` menu

graahnul-grom:
Track text value of the current cell, and if it's
modified after the cell is deactivated, mark the
schematic as changed, so that `lepton-attrib` prompt
the user to save changes on exit.

I do not set the "changed" flag in `s_toplevel_add_new_attrib()`,
because initially the new column is empty.

Closes #702.

It seems that I've discovered yet another bug: after deleting
an attribute column and saving changes, all invisible attributes
becomes visible. Could anyone confirm?

### DIFF
--- a/libleptonattrib/src/attrib.c
+++ b/libleptonattrib/src/attrib.c
@@ -95,6 +95,19 @@ gboolean attrib_really_quit(void)
   eda_config_save (cfg, NULL);
 
 
+  /* Deactivate the current cell to trigger "deactivate" signal.
+   * This allows changing of the sheet_head->CHANGED flag in the
+   * on_deactivate() handler function if needed.
+  */
+  for (int i = 0; i < NUM_SHEETS; ++i)
+  {
+    if (sheets[i] != NULL)
+    {
+       gtk_sheet_set_active_cell (sheets[i], -1, -1);
+    }
+  }
+
+
   if (sheet_head->CHANGED == TRUE) {
     x_dialog_unsaved_data();
   } else {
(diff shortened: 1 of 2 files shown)


vzh:
Yeah, it sets both `visibility` and `show_name_value` fields to 1.
Moreover, it can change values of some attributes: try to open *examples/gTAG/gTAG*.sch*, change something & save. Then use git to see what has changed.

vzh (approved):
Thanks!

graahnul-grom:
So far, I can't see that it *changes* values. Only visibility is reset.
It seems that I've found the problem, will try to check that idea tomorrow.

graahnul-grom:
@vzh Could you please take a look at #711?

vzh:
I meant, do just:
> lepton-attrib examples/gTAG/gTAG*.sch

that is `gTAG.sch` and all its subschematics together, and delete, say, the column `comment` or do any another change then save.
Then look if the `device` attributes of symbols `in-1.sym` and `out-1.sym` change.  On my system, one file even gets killed (`gTAG-ucont.sch`).  I discovered this behavior long ago, it doesn't belong to recent changes.

PR #706, opened 2020-12-06
Install Lepton reference manual images into the info directory

graahnul-grom:
`make install` should copy image files into the info
directory, too (usually, `/usr/share/info/`).

Thank you, Bdale (@bdale), for reporting this on gitter.


vzh (approved):
Thanks!

PR #711, opened 2020-12-09
Fix lepton-attrib code that removes an attribute column

graahnul-grom:
Problem: after deleting an attribute column and saving
changes, all invisible attributes becomes visible.

Such a data mangling resulted from destroying
the cell data without saving its contents first.
It was reconstructed with default values, leading
to messed up schematic with ALL attributes'
visibility reset to "visible" and "value only"
after save.

Fix it by saving the cell data and restoring it after
an attribute is deleted.


vzh (approved):
Yes, the code now works better, thanks!

PR #713, opened 2020-12-09
Backward-compatible color-map.scm file

### LINKED ISSUE #707: Please provide backward-compatible color-map.scm file
In PR #672 `color-map.scm` was moved and converted to module.
Though we strive to be as compatible as possible wrt existing code
(keeping old modules, carefully converting configuration), in this
case nothing has been done.
Every script dealing with attributes or symbols is likely to use it,
in particular `color-map-name-to-index()` and
`color-map-name-from-index()` functions.
It would be nice, if we provide support for such code, so that
expression `( load-from-path "color-map.scm" )` and at least
the above two functions could still work, ideally displaying
a warning with info that it's deprecated and a hint what
should be used instead.

graahnul-grom:
Add `color-map.scm` file with functions:
  - `color-map-name-to-index()`
  - `color-map-name-from-index()`

It makes code calling these functions and
`(load-from-path "color-map.scm")` work again.
Warn that this file is deprecated upon loading.

Closes #707.

### DIFF
--- a/liblepton/scheme/color-map.scm
+++ b/liblepton/scheme/color-map.scm
@@ -15,23 +15,11 @@
 )
 
 
-; Function: color-map-name-to-index()
-;
-; Returns an integer index given one of a color
-; symbols like 'background, 'net, 'text, etc.
-; Returns #f (false) if wrong symbol is specified.
-;
 ( define color-map-name-to-index
   ( @ (lepton color-map) color-map-name-to-index )
 )
 
 
-; Function: color-map-name-from-index()
-;
-; Returns one of a color symbols like 'background,
-; 'net, 'text, etc., given an integer index.
-; Returns #f (false) if wrong index is specified.
-;
 ( define color-map-name-from-index
   ( @ (lepton color-map) color-map-name-from-index )
 )
--- a/liblepton/scheme/color-map.scm
+++ b/liblepton/scheme/color-map.scm
@@ -1,0 +1,38 @@
+; Lepton EDA
+; Backward-compatible color-map.scm file
+; Copyright (C) 2020 Lepton EDA Contributors
+; License: GPLv2+. See the COPYING file
+
+( use-modules ( lepton color-map ) )
+
+
+; Function: color-map-name-to-index()
+;
+; Returns an integer index given one of a color
+; symbols like 'background, 'net, 'text, etc.
+; Returns #f (false) if wrong symbol is specified.
+;
+( define ( color-map-name-to-index sym-name )
+  ; return:
+  ( assoc-ref %color-name-map sym-name )
+)
+
+
+; Function: color-map-name-from-index()
+;
+; Returns one of a color symbols like 'background,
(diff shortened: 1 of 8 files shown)


vzh:
@graahnul-grom, why not just use those functions from the imported module, defining them as public?

graahnul-grom:
I have not found those functions in the new module.
There are some local functions with similar names,
but I'm unsure of what are they for.
Anyway, it's a couple of lines, not a big deal.
You don't like my functions? :-))

graahnul-grom:
I'm joking.
You're right, I'll move these functions to the module
and export them.
One thing that looks strange to me is this: they return
their argument if it's incorrect (e.g. index out of bounds).
I think in that case it's better to return `#f`.

vzh:
They do almost the same (grow from the same root) and can be easily made non-local.  Why duplicate their code? Just to show your DBCS is better than mine ;-) (Please stop trolling me.)

graahnul-grom:
We'll continue trolling tomorrow. :-)
К сожалению, мне уже пора спать.
Спокойной ночи, Володя!

graahnul-grom:
What about returning `#f` from ` color-map-name-toindex()`
and ` color-map-name-from-index()` if argument is invalid?
Another thought: shouldn't we make color map files
(`*-colormap-*`) self-sufficient by adding
`( use-modules (lepton color-map) )` line?

liblepton/scheme/color-map.scm
```diff
@@ -0,0 +1,50 @@
...
+; Function: color-map-name-from-index()
+;
+; Returns one of a color symbols like 'background,
+; 'net, 'text, etc., given an integer index.
+; Returns its argument if wrong index is specified.
+;
+( define color-map-name-from-index
+  ( @ (lepton color-map) color-map-name-from-index )
+)
+
```
vzh, line 38: You can just add:
> (use-modules ((lepton color-map)
>              #:select (%color-name-map
>                        color-map-name-to-index
>                        color-map-name-from-index)))

instead of all this.
Comments could go to `(lepton color-map)` itself as comments or docstrings (it's better since users could reference them then in interactive mode).  And here could be just reference to the module, something like: Please see the code of the module bla-bla ... use the command `,describe function-name` in interactive mode to get the function description.

vzh (changes requested):

vzh:
> What about returning `#f` from ` color-map-name-to-index()`
> and ` color-map-name-from-index()` if argument is invalid?
> Another thought: shouldn't we make color map files
> (`*-colormap-*`) self-sufficient by adding
> `( use-modules (lepton color-map) )` line?

I don't mind if it breaks nothing.

graahnul-grom:
Thank you!
Please check if everything is OK now.
BTW, please fix a typo in the `NEWS.md`:
"1.9.13 (upcoming)", not "1.9.12 (upcoming)".

vzh:
Applied and fixed, thanks.
Please remove your stale branches.

PR #716, opened 2020-12-10
Rename the index file in the Lepton EDA Reference Manual

### LINKED ISSUE #715: Lepton manual files on case-insensitive file systems
There are two generated `html` files:
- `index.html` (lower-case "i"): table of contents (start page)
- `Index.html` (upper-case "i"): index page

On filesystems that do not distinguish lower- and
upper-case file names, `make` overwrites one or
the other file. Thus, on Windows/Cygwin, for example,
the table of contents is lost.
In the Scheme manual, there are "concept index",
"variable index" and "function index", no problems.
The "Index" page should be renamed somehow.

graahnul-grom:
Generated HTML files for index and the table
of contents have similar names: only case of
the first letter is different ("Index.html"
and "index.html"). On file systems that do not
distinguish lower- and upper-case file names,
"make" overwrites one or the other file.
Fix it by using a new name: "Topic Index".

Closes #715.

### DIFF
--- a/docs/manual/geda-file-format-spec.texi
+++ b/docs/manual/geda-file-format-spec.texi
@@ -1,4 +1,4 @@
-@node gEDA file format, Index, Development, Top
+@node gEDA file format, Topic Index, Development, Top
 @chapter gEDA/gaf File Format Document
 @cindex gEDA file format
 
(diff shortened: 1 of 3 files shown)


vzh:
Well, I think it's the time to release 1.9.13.

graahnul-grom:
I think, yes. I'm glad that we've fixed the issues with
`lepton-schematic -c`, `color-map.scm` and Cygwin.
Thank you very much! Very nice work.

PR #721, opened 2020-12-25
Get rid of the module (schematic core builtins)

vzh:
- Changes in `libleptongui` and `liblepton`:
  - Less C files, global C defines, and Scheme-C interlacement:
  - Unified deprecation syntax in the module `(gschem deprecated)`.
  - The module `(schematic core builtins)` has been removed.
  - `main-prog()` and some other functions have been rewritten in Scheme and simplified (no g_free()`, `scm_dynwind_*()`, etc. 

- Better FFI code (for @graahnul-grom):
  - Less function definitions in the code.
  - FFI definitions are mostly done in specialized modules.
  - Syntax rules are used to reduce and simplify definitions.
  - Consistent function naming: foreign C functions and their Scheme counterparts are same named.

Merry Christmas to those who celebrate it today :-)


graahnul-grom (commented):
The names of the actions in active tab's context menu
should be changed, too: `file-new` => `&file-new`, etc.

[tab_menu_actions_names.diff.txt](https://github.com/lepton-eda/lepton-eda/files/5741745/tab_menu_actions_names.diff.txt)

graahnul-grom:
> Consistent function naming

Thank you very much :-)

schematic/lepton-schematic.scm
```diff
@@ -265,42 +179,146 @@ Run `~A --help' for more information.\n")
...
+          (gtk_icon_theme_append_search_path (gtk_icon_theme_get_default)
+                                             icon-dir)
+          (loop (cdr sys-dirs))))))
+
+
+(define (open-log-window)
+  (let ((cfg (path-config-context (getcwd))))
+    (when (string= (config-string cfg "schematic" "log-window")
+                   "startup")
+      (x_widgets_show_log window))))
```
graahnul-grom, line 229: `window` is undefined here.
If `schematic::log-window` is set to "startup", lepton-schematic exits with:
```
Backtrace:
           5 (primitive-load "/home/dmn/lepton/bin.tb/bin/lepton-sch…")
In ice-9/eval.scm:
   293:34  4 (_ #(#<directory (guile-user) 800731140> ()))
    619:8  3 (_ #(#(#(#<directory (guile-user) 800731140>) ()) #<…> …))
    155:9  2 (_ #(#(#(#<directory (guile-user) 800731140>)) #<geda…>))
   223:20  1 (proc #(#(#(#<directory (guile-user) 800731140>)) #<g…>))
In unknown file:
           0 (%resolve-variable (7 . window) #<directory (guile-user…>)

ERROR: In procedure %resolve-variable:
Unbound variable: window
```
The attached patch fixes this.

[log_window_on_startup.diff.txt](https://github.com/lepton-eda/lepton-eda/files/5741942/log_window_on_startup.diff.txt)

libleptongui/src/i_callbacks.c
```diff
@@ -49,7 +34,8 @@
  *  since there is a call: widget = NULL, data = 0 (will be w_current hack)
  *  \todo This should be renamed to page_new perhaps...
  */
-DEFINE_I_CALLBACK(file_new)
+void
+i_callback_file_new (GtkWidget *widget, gpointer data)
```
graahnul-grom, line 38: What is the `widget` parameter in all `i_callback_XXX()` functions?
It doesn't used anywhere.

graahnul-grom (approved):
Well done!
It's nice that you got rid of `DEFINE_I_CALLBACK` and
`DEFINE_G_KEYS` macros. Maybe we should add a note
to CONTRIBUTING.md that macros are absolute evil.
That was certainly a very bad programming style.
`lepton-schematic` startup code is now held in one place,
excellent.

libleptongui/scheme/schematic/builtins.scm
```diff
@@ -45,48 +47,62 @@
...
-  (%file-image))
+  (run-callback i_callback_file_write_png "&file-image"))
 
 (define-action-public (&file-script #:label (G_ "Run Script") #:icon "gtk-execute")
-  (%file-script))
+  (run-callback i_callback_file_script "&file-script"))
 
 (define-action-public (&file-new-window #:label (G_ "New Window") #:icon "window-new")
-  (%file-new-window))
+  (i_callback_file_new_window %null-pointer %null-pointer))
```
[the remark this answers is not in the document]
vzh: Because the callback doesn't use the `data` argument.  The check for non-null current window is unnecessary here.

schematic/lepton-schematic.scm
vzh: Thanks, I overlooked this during refactoring.

libleptongui/src/i_callbacks.c
vzh: While it is not used in Scheme, it may be used when you want to run the callbacks from C.   Some callbacks are indeed used in C, in the toolbar button creation code, and I've unified some of them and removed duplicates.  Since they are connected to glib signals such as "clicked" or "toggled", we have to use the standard pattern for those signals.  So the `widget` parameters for those callbacks are buttons utilizing the signals.  I tried to make all i_callbacks callable from both Scheme and C, though did not managed to do this for some of them yet since it requires more work on resolving C and Scheme code interdependence.

vzh:
@graahnul-grom, thank you! I've applied your patches and added info to CONTRIBUTING.md.

PR #723, opened 2020-12-29
Small improvements in lepton-attrib GUI

graahnul-grom:
- Add status bar and show color legend there, designating
the meaning of different attributes' colors (visibility).
This information should be somewhere in the main window,
anyway; GUI app describing its user interface elements in
the `--help` output, is... not quite **G**UI. :-)
Please suggest a better way how to do it (now it's just
four strings in different colors).
- Display the current file name in the window's title, and
if the file is changed, prepend an asterisk to its name.
Show "Multiple files" if more than one are opened.
- Do not show the one and only `GtkNotebook`'s tab.
- Explicitly display slotted components in row headers,
e.g. show `U1 (slot 3)` instead of `U1.3`.
- Display components as slotted only if they have the
`slot` attribute.
- On `File->Export CSV`, show the overwrite confirmation
dialog if already existing file is selected.
- Default response in the `Delete Attribute` dialog:
focus the "No" button by default.
-  Improve the `About` dialog, use `GtkAboutDialog`, like
in `lepton-schematic`.
- Fix modification status update on `Ctrl+S`.

old:
attrib_main_wnd_old

new:
attrib_main_wnd_new

old:
attrib_about_dlg_old

new:
attrib_about_dlg_new

[a ~~sample schematic~~ clever circuit](https://graahnul-grom.github.io/sch/index.html#slots) used in screenshots:
slots

### DIFF
--- a/libleptonattrib/include/prototype.h
+++ b/libleptonattrib/include/prototype.h
@@ -159,6 +159,7 @@ void x_dialog_about_dialog();
 void x_dialog_export_file();
 
 /* ------------- x_gtksheet.c ------------- */
+void x_gtksheet_set_saved();
 void x_gtksheet_init();
 void x_gtksheet_add_row_labels(GtkSheet *sheet, int count, STRING_LIST *list_head);
 void x_gtksheet_add_col_labels(GtkSheet *sheet, int count, STRING_LIST *list_head);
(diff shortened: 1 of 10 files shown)


vzh:
Nice, thanks.
I would add a couple of changes from [here](https://github.com/vzh/lepton-eda/commits/tb_310_attrib_gui_improve).

graahnul-grom:
Thank you!

vzh (approved):

PR #694, opened 2020-11-18
Fix the close confirmation dialog's title

graahnul-grom:
Show "lepton-schematic" in the title, instead of "\<unknown\>".
I wonder why nobody noticed this before. :-)

PR #724, opened 2020-12-31
Several fixes in lepton-schematic GUI

graahnul-grom:
- Fix window title in several dialogs to be in line with
other dialogs:
  - `Save Color Scheme As...` => `Save Color Scheme`
  - `Edit slot number` => `Edit Slot`
  - `Autonumber text` => `Autonumber Text`

- Fix title of the page close confirmation dialog: show
"lepton-schematic" instead of "\<unknown\>" (as in #694)

- Add missing title to some dialogs:
  - Clear log confirmation dialog
  - Load backup dialog

- Set default response to "No" in file overwrite
confirmation dialogs for:
  - `File->Save As`
  - `View->Color Scheme Editor->Save As`

- Fix save color scheme error dialog: actually show it :-)


vzh (approved):
Good catch in the last commit of the series :+1:

PR #725, opened 2021-01-01
Fix: cancel bus drawing with the right mouse button

graahnul-grom:
The bus drawing tool is supposed to behave like the
net tool: a click with the right mouse button should
cancel the current operation. It is even reflected in
the toolbar button's hint: "Add buses mode. Right mouse
button to cancel". However, the right mouse button does
nothing in the bus mode. Fix it.

I invite everybody to test the proposed patch.
For some strange reasons, it works for me. :-)
Happy New Year, friends!


vzh (approved):
I also noticed the bug before.  Let's give the fix a try.

PR #729, opened 2021-01-29
Fix lepton-schematic --command

graahnul-grom:
This command line option was introduced in 1.9.12 (7ca7c7d)
as an alias for `-c EXPR` (execute Scheme `EXPR`), but
now it's missing. Fix it, and document it in the
lepton-schematic man page.

I doubt that it's worth switching to `srfi-37` to process
command line options. It is ugly,
it can't understand `--option VALUE`, only `--option=VALUE`
and unable to handle the case when required argument
is missing, launch `lepton-schematic -c` and you get:

```
Backtrace:
           5 (primitive-load "/home/dmn/lepton/bin.tb/bin/lepton-sch…")
In ice-9/eval.scm:
   293:34  4 (_ #<directory (guile-user) 800731140>)
    155:9  3 (_ #(#(#<directory (guile-user) 800731140>)))
In srfi/srfi-37.scm:
   169:19  2 (next-arg)
   132:41  1 (invoke-option-processor #<srfi-37:option names: (#\c …> …)
In unknown file:
           0 (scm-error misc-error "args-fold" "Missing required ar…" …)

ERROR: In procedure scm-error:
In procedure args-fold: Missing required argument after `-c'
```

In this case it only raises an exception (look at the source code),
which is undocumented.

### DIFF
--- a/libleptongui/docs/lepton-schematic.1.in
+++ b/libleptongui/docs/lepton-schematic.1.in
@@ -36,7 +36,7 @@ Scheme files.
 \fB-s\fR \fIFILE\fR
 Specify a Scheme script to be executed at startup.
 .TP 8
-\fB-c\fR \fIEXPR\fR
+\fB-c\fR \fIEXPR\fR, \fB--command\fR=\fIEXPR\fR
 Specify a Scheme expression to be evaluated at startup.
 .TP 8
 \fB-h\fR, \fB--help\fR
--- a/schematic/lepton-schematic.scm
+++ b/schematic/lepton-schematic.scm
@@ -118,7 +118,7 @@ Options:
   -q, --quiet              Quiet mode.
   -v, --verbose            Verbose mode.
   -L DIR                   Add DIR to Scheme search path.
-  -c EXPR                  Scheme expression to run at startup.
+  -c EXPR, --command=EXPR  Scheme expression to run at startup.
   -s FILE                  Scheme script to run at startup.
   -V, --version            Show version information.
   -h, --help               Help; this message.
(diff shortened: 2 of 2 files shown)


vzh:
[srfi-37](https://srfi.schemers.org/srfi-37/srfi-37.html) conforms to:
- [POSIX guidelines for command line options](https://pubs.opengroup.org/onlinepubs/007904975/basedefs/xbd_chap12.html#tag_12_02) and
- [GNU guidelines](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html)
Please note no one mentions long options with space as separator. Long options with the equal sign is GNU extension to what POSIX specifies.  The authors of `getopt long` decided to extend it even more.  However, I like `srfi-37` more, because (quote): *Unlike ‘getopt-long’, it supports repeated options and any number of short and long names per option*.

You could take a look at these [Mail archives - srfi-37](https://srfi-email.schemers.org/srfi-37/), especially:
- https://srfi-email.schemers.org/srfi-37/msg/2762603/
- https://srfi-email.schemers.org/srfi-37/msg/14514666/
- https://srfi-email.schemers.org/srfi-37/msg/15639519/

As for `scm-error`, the [reference implementation of srfi-37](https://srfi.schemers.org/srfi-37/srfi-37-reference.scm) does not have it.  It's Guile guys who decided to do it in such a way.  I agree with you, the backtrace looks ugly.  It would be enough to have just the error message and exit with non-zero status.

vzh (approved):

graahnul-grom:
Thank you for the links. Indeed, there's no `--`
options in `POSIX`, and `GNU` only mentions
`--name=value`.

> Unlike ‘getopt-long’, it supports repeated options and any number of short and long names per option

I don't think these are so vital features that should force us to
abandon the clean and reliable `getopt-long` interface.

vzh:
AFAIR, that *clean and reliable* interface forced me to write an additional processing level for `lepton-netlist` namely wrt processing repeated options.  That is why now I would prefer srfi-37 just for that task.

graahnul-grom:
Well, OK.  The options processing in `netlist` is something... :-)
If `srfi-37` will help to simplify it, then yes, let's use `srfi-37`!

PR #730, opened 2021-01-29
Fix a link in the reference manual

graahnul-grom:
Fix a link to the Scheme API manual (for HTML output):
`lepton-scheme` => `lepton-scheme.html`

### DIFF
--- a/docs/manual/lepton-cli-shell.texi
+++ b/docs/manual/lepton-cli-shell.texi
@@ -5,7 +5,7 @@
 @code{lepton-cli shell} provides an interactive Scheme Read-Eval-Print
 Loop (REPL) for automating processing of schematic and symbol
 files. It is designed to be used with the Lepton EDA Scheme API.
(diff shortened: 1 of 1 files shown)


vzh:
This will break the link in the `info` format.  I often use it if I need to quickly find something in the documentation.  Here should be something like `@ifhtml`, `@ifinfo`, `@iftex`, etc. to fix this in all (at least in three) cases.

graahnul-grom:
Sorry, I forgot  about the other output formats!
In `info` manual that link works correctly.

graahnul-grom:
Hopefully, I've fixed that - using [@inlinefmt](https://www.gnu.org/software/texinfo/manual/texinfo/html_node/Inline-Conditionals.html).
Is it correct to place it inside `@xref`?
Please take a look; somehow, it works for me. :-)

graahnul-grom:
`travis` today is painfully slow.

vzh (approved):
Looks OK and works for me as well. Thank you!

graahnul-grom:
Great!

PR #733, opened 2021-02-02
Avoid displaying backtraces on wrong command line options for lepton-schematic. 

vzh:
Instead, show a useful prompt.  Discussed in #729.

Another minor fix is a crash when the following code is used:
> lepton-schematic -c '(quit)'

Now it works OK (strangely, it doesn't lead to exit from the program, while if I use `(primitive-exit 0)` instead it does.

### DIFF
--- a/schematic/lepton-schematic.scm
+++ b/schematic/lepton-schematic.scm
@@ -25,7 +25,7 @@ exec @GUILE@ -s "$0" "$@"
 
 (use-modules (ice-9 match)
              (srfi srfi-1)
-             (srfi srfi-37)
+             (lepton srfi-37)
              (system foreign))
 
 ;;; Load and initialize liblepton library.
--- a/schematic/lepton-schematic.scm
+++ b/schematic/lepton-schematic.scm
@@ -135,7 +135,7 @@ Lepton EDA homepage: ~S\n")
 
 ;;; Parse lepton-schematic command-line options, displaying usage
 ;;; message or version information as required.
-(define (parse-commandline*)
+(define (parse-commandline)
   "Parse command line options.  Return the list of non-option
 arguments which should represent the list of schematics to open."
   (reverse
@@ -180,21 +180,6 @@ Run `~A --help' for more information.\n")
       (exit 1))
     (lambda (op seeds) (cons op seeds))
     '())))
-
-;;; Catch for errors in parse-commandline*().
-(define (parse-commandline)
-  (catch 'misc-error
-    parse-commandline*
-    (lambda (key . args)
-      ;; Throw away the first arg (function name: args-fold), and
-      ;; all the rest.
-      (let ((message (second args))
-            (option-name (third args)))
-        (format #t (G_ "ERROR: ~A.
-Run `~A --help' for more information.\n")
-                (apply format #f message option-name)
-                (basename (car (program-arguments))))
-        (exit 1)))))
 
 
 ;;; Load GTK resource files.
--- a/schematic/lepton-schematic.scm
+++ b/schematic/lepton-schematic.scm
@@ -129,7 +129,7 @@ Lepton EDA homepage: ~S\n")
           (car (program-arguments))
           (lepton-version-ref 'bugs)
           (lepton-version-ref 'url))
-  (primitive-exit 0))
+  (exit 0))
 
 
 ;;; Parse lepton-schematic command-line options, displaying usage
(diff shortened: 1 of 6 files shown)


graahnul-grom:
> primitive-exit() unwinds only C stack

Amazing, how didn't I notice this before. :-)
- `primitive-exit()` - unwinds `C` stack
- `exit()` - unwinds `Scheme` stack

graahnul-grom (approved):
Yes, the code in `parse-commandline()` solves the problem.
It is done relying on knowledge of undocumented internals,
assuming that the only exception thrown is the 'misc-error
(with certain arguments in particular order) in case of absent
required argument.
Generally speaking, it's not the best practice.
Let me once again express my opinion, that it's better
to stay with `getopt-long`. This issue shows that `srfi-37`
is not quite finished yet, lacking support for such a trivial
case like absent arguments.

graahnul-grom:
OK, let's use `srfi-37`. :-)

Could you please try to reproduce one thing I've just discovered?
On my system, in `File->REPL` certain GUI command lead to crash, like
`&options-show-log-window()` or `&file-new()`, while others do not,
e.g. `&options-grid()`.

`scheme@(guile-user)> (&options-show-log-window)`
```
[xcb] Unknown request in queue while dequeuing
[xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
[xcb] Aborting, sorry about that.
Assertion failed: (!xcb_xlib_unknown_req_in_deq), function dequeue_pending_request, file xcb_io.c, line 165.
Abort trap
```

vzh:
GLib is completely thread safe... GTK+, however, is not thread safe... (https://developer.gnome.org/gdk3/stable/gdk3-Threads.html)

You must call `g-thread-init` and `gdk-threads-init` before executing any other GTK+ or GDK functions in a threaded GTK+ program... you must surround the callback with a `gdk-threads-enter`/`gdk-threads-leave` pair... (https://www.gnu.org/software/guile-gnome/docs/gdk/html/Threads.html)

A few years ago I tried to work in the background terminal REPL with the help of `guile-gnome` and used `gdk-threads-enter/leave` for that.  They freeze the GUI until `leave` is run.  Not very convenient, eh?  Now I avoid doing so.  I run REPL only for non-GUI stuff.

Did you know that any REPL uses a new thread?

graahnul-grom:
Thank you, I see, you run `(begin-thread (lepton-repl))`.

vzh:
About *undocumented internals*...

We might make them documented, or better, avoid them at all by using another approach: just add the code of the srfi-37 module, be it the reference implementation I referenced before, or the module from Guile, to Lepton and fix it there as we wish :-)  The reference implementation is in public domain, the Guile one is under GPL v3, so anything is OK, I believe.

graahnul-grom:
Nice idea. Indeed, why do we have to wait for someone
to fix what we need? It's just one module. After all, we have
`m4/guile.m4` from the guile 2.2.6 package, and I further
modify it in Fedora RPM...

schematic/lepton-schematic.scm
```diff
@@ -1,11 +1,12 @@
 #!/usr/bin/env sh
 export GUILE_LOAD_COMPILED_PATH="@ccachedir@:${GUILE_LOAD_COMPILED_PATH}"
+export GUILE_LOAD_PATH="@LEPTON_SCHEME_MODULE_DIRECTORY@:${GUILE_LOAD_PATH}"
```
graahnul-grom, line 3: Now we have 3 (three) `$PREFIX/share/lepton-eda/scheme`
entries in the `%load-path`. :-)

graahnul-grom:
There are `^L` symbols in the `srfi-37.scm` file.
Where did they come from?

graahnul-grom:
Very nice improvement of `srfi-37`!
:+1:

vzh:
> There are `^L` symbols in the `srfi-37.scm` file.
> Where did they come from?

The character was historically used as *Form feed* or *New paper* command for line printers, see e.g. [here](https://www.linuxquestions.org/questions/linux-newbie-8/how-to-remove-%5El-control-characters-from-text-files-4175426865/). Another use was converting plain text to Postscript/PDF by an interpreter like *ghostscript*.  There's lots of the `^L` characters in Guile sources. Remember, the initial version of `srfi-37` has been written back in 2002.

schematic/lepton-schematic.scm
vzh: Well, yes. But at the time it is needed, before initialization of `liblepton` extension there is only one defined in that code.  I could not managed to make the module work any other way.  It doesn't wish to be loaded from `primitive-eval` for some reason.

Issue #732, opened 2021-01-30
Organize the repository


graahnul-grom:
Now that the dust has settled, we have time to look at the
repository layout. It's far from perfect. "Inconsistency"
is the keyword here.

Let me propose the following changes:

- Move `lepton-cli` source code into a hierarchy similar to
that of the other components: `cli/src/`, `cli/docs/`, etc.
**UPDATE:** fixed in #740

- Do not mix symbols with other stuff. Split up the contents
of the symbols directory, e.g.:
  symbols/
  |-- docs/
  |-- scheme/
  |-- sym/

**UPDATE:** fixed in #751

BTW, do we really need the `nc.doc` satiric document?
**UPDATE:** fixed in #761: moved to Lepton manual

- Remove the `docs/toplevel` directory with the only file
which contains the only link to the wiki main page.
**UPDATE:** fixed in #750

- Create `docs/news/` and move `NEWS-1.9.2.txt` there.
**UPDATE:** fixed in #734

- Move `libleptonattrib/design` to `docs/`.
**UPDATE:** fixed in #734

- Clean up the mess in `utils/`. All that utils/scripts
separation is obsolete. We do not have "utils" or "scripts",
we have the *components* of the suite, which are installed
into the system `bin/` and `lib/` directories.
Here they are, in no particular order:

1) `liblepton`
2) `libleptongui`
3) `libleptonattrib`
4) `schematic`
5) attrib  [**UPDATE:** => `utils/attrib` #763]
6) netlist  [**UPDATE:** => `utils/netlist` #742]
7) cli  [**UPDATE:** => `utils/cli` #740]
8) tragesym  [**UPDATE:** => `utils/tragesym` #738]
9) sch2pcb  [**UPDATE:** => `utils/sch2pcb` #735]
10) archive  [**UPDATE:** => `utils/archive` #739]
11) symcheck  [**UPDATE:** => `utils/symcheck` #739]
12) symfix  [**UPDATE:** => `utils/symfix` #739]
13) embed  [**UPDATE:** => `utils/embed` #739]
14) schdiff  [**UPDATE:** => `utils/schdiff` #739]
15) pcb_backannotate  [**UPDATE:** => `utils/pcb_backannotate` #739]
16) renum  [**UPDATE:** => `contrib/renum` #756]
17) refdes_renum  [**UPDATE:** => `utils/refdes_renum` #739]
18) upcfg  [**UPDATE:** => `utils/upcfg` #739]

[...]

PR #739, opened 2021-02-10
Move utilities into separate directories under utils/

graahnul-grom:
@vzh What do you think about moving `lepton-netlist`
back to `netlist/` and `lepton-cli` to `utils/`?

Affects #732.


vzh:
I'd prefer to keep both utilities in `utils/`.  The only thing I'm unsure is where to store backends.  Test suites could be merged and moved say to `build-aux` as `automake` suggests, but backends is a netlister property not very well integrated with `lepton-netlist` so I'm in doubt.

vzh:
Thank you!

graahnul-grom:
> I'd prefer to keep both utilities in utils/.

Don't you consider `netlist` as one of the core
components of the suite? :-)

You mentioned in #732 that common guile files could be
moved to some top-level directory (say, `scm/`).
Maybe it's worth moving backends there, too.

vzh:
> Don't you consider `netlist` as one of the core
> components of the suite? :-)

Well, I think the main program in the suite is `schematic`.  All other utilities just do something useful of its results: fixing old schematics, automating changes in schematics, producing documents, images, netlists for drc, simulation, pcb layout.  Anything of this could be more tightly integrated into the program itself.  I don't understand why the user has to do many *cli* steps to do this during designing a project while we have all means to simplify this by integrating the functionality of any of the utilities into `schematic`.  Imagine using a shortcut for producing the graph of your SPICE simulation instead of running `lepton-netlist`, and then `gnucap` or `ngspice`, and then some other program to see it.  The only thing all those tools matter is automating of documentation creation (e.g. by using makefiles).  That is why most of the tools are useful (not all, I believe). `lepton-netlist` is useless without `lepton-schematic`.

BTW, for uniformity, I would move even `lepton-schematic` to `utils/` and then rename it to, say, `bin` or `tools` or `programs` :-)
 
> You mentioned in #732 that common guile files could be
> moved to some top-level directory (say, `scm/`).
> Maybe it's worth moving backends there, too.

I'm not sure if it is worth though we could try it :-) (Guile, for instance, has the `modules/` directory for its modules.  Other Scheme files are in other places.)

graahnul-grom:
> I don't understand why the user has to do many *cli* steps
> to do this during designing a project while we have all
> means to simplify this by integrating the functionality of
> any of the utilities into `schematic`.

Yes, we have rich and well-documented Guile API.
And I have easy-to-customize "external tools" menu for that.

> Imagine using a shortcut
> for producing the graph of your SPICE simulation instead
> of running `lepton-netlist`, and then `gnucap` or `ngspice`,
> and then some other program to see it.

No need to imagine. It is reality. :-)
I've been doing exactly this for several years now:
I select a net(s), press a button and immediately see
window(s) with graph(s) drawn by the ngspice 'plot' command.

> The only thing all those tools matter is automating...

Exactly. And this is great that we have them, and not
one big integrate-them-all all-in-one program.

> BTW, for uniformity, I would move even `lepton-schematic`
> to `utils/` and then rename it to, say, `bin` or `tools`
> or `programs` :-)

Not sure I understand why do we need this.

> I'm not sure if it is worth though we could try it :-)
> (Guile, for instance, has the `modules/` directory for its
> modules. Other Scheme files are in other places.)

So, we need to compose a list of modules we should move
to that dir? Or just move all the `liblepton/scheme` there?

vzh:
...
> No need to imagine. It is reality. :-)
> I've been doing exactly this for several years now:
> I select a net(s), press a button and immediately see
> window(s) with graph(s) drawn by the ngspice 'plot' command.

I mean this could be better integrated and available for newbies without the need of complex setup in the first place.

...
> Not sure I understand why do we need this.

Never mind, not a big deal :-)

> So, we need to compose a list of modules we should move
> to that dir? Or just move all the `liblepton/scheme` there?

I think about moving all scheme modules there, including `gui` ones, and placing the `backend/` directory in `netlist/` although the files there are not modules.

graahnul-grom:
>> No need to imagine. It is reality. :-)
>> I've been doing exactly this for several years now:
>> I select a net(s), press a button and immediately see
>> window(s) with graph(s) drawn by the ngspice 'plot' command.

> I mean this could be better integrated and available for newbies without the need of complex setup in the first place.

Yes, I agree.
I wish it were less complex. I never thought that someday
someone except me would want to use this stuff. :-)

PR #745, opened 2021-02-15
Rename object color accessors and actually use them in the code

vzh:
Amend the code to get rid of direct using the `color` field of the `LeptonObject` structure.


graahnul-grom (approved):
I found yet another place where accessors
should be used: `libleptongui/src/o_path.c`,
`o_path_draw_rubber()`, `o_path_draw_rubber_grips()`.

[o_path.c.diff.txt](https://github.com/lepton-eda/lepton-eda/files/5995902/o_path.c.diff.txt)

vzh:
Fixed, thanks.

PR #743, opened 2021-02-15
Rename OBJECT and GedaObject structs to LeptonObject

vzh:
Continue with leptonization and finish work on renaming for this particular struct.

PR #746, opened 2021-02-16
Rename PAGE and GedaPage structures

vzh:
Changes like in #743 though for pages.

### DIFF
--- a/liblepton/include/libgedaguile_priv.h
+++ b/liblepton/include/libgedaguile_priv.h
@@ -62,8 +62,9 @@
  * For more information on dynamic wind, see the Guile Reference
  * Manual.
  *
- * The remaining functions in this module allow you to convert Lepton EDA
- * #LeptonObject and #PAGE structures to and from Scheme values ("smobs").
+ * The remaining functions in this module allow you to convert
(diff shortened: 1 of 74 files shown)


graahnul-grom (approved):
Are you going to also rename source files
(e.g. remove `geda_` and `gschem_` prefixes,
or change them)?

vzh:
Yes, I am. Haven't decided what to do better yet: rename by dropping the prefixes or replace `geda` with `lepton`, `gschem` with `schematic` or something like this.  I believe it's a subject for discussion.  WDYT?

graahnul-grom:
I think in most cases the prefix is redundant, probably
only makes sense when there are conflicts in file names.
Even header files are installed to dedicated subdirs,
`include/libleptonattrib/`, `include/liblepton/`, so
shouldn't clash with anything.

vzh:
OK, let's get rid of them then.

bdale:
dmn <notifications@github.com> writes:

> I think in most cases the prefix is redundant, probably
> only makes sense when there are conflicts in file names.

I agree.

Bdale

vzh:
OK, thanks.  I've fixed my issue with `string.h` mentioned somewhere else by renaming it to `str.h`.

PR #748, opened 2021-02-17
Do not hard-code build date and time in lepton-renum exe

graahnul-grom:
It's not necessary.
Should aid "reproducible builds".

### DIFF
--- a/utils/renum/src/lepton-renum.c
+++ b/utils/renum/src/lepton-renum.c
@@ -396,5 +396,4 @@ void printver()
 	printf("This is lepton-renum, an advanced refdes renumber utility for Lepton EDA's lepton-schematic.\n");
 	printf("Version %s.  Lepton EDA version %s.%s\n",GRVERSION,
                PACKAGE_DOTTED_VERSION, PACKAGE_DATE_VERSION);
-	printf("Compiled on %s at %s\n",COMP_DATE,COMP_TIME);
 	}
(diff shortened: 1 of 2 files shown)


utils/renum/src/lepton-renum.c
```diff
@@ -394,7 +394,6 @@ void printhelp()
 void printver()
 	{
 	printf("This is lepton-renum, an advanced refdes renumber utility for Lepton EDA's lepton-schematic.\n");
-	printf("Version %s.  Lepton EDA version %s.%s\n",GRVERSION,
+	printf("Version %s. Lepton EDA version %s.%s\n",GRVERSION,
```
vzh, line 397: I think the author added two spaces here not without reason.  Some people just used to it because it is recommended to do so in plain text manuals processed by roff or texinfo.  See for example the texinfo manual ([here](https://www.gnu.org/software/texinfo/manual/texinfo/html_node/Not-Ending-a-Sentence.html) and the next section).  Text processors like TeX/LaTeX just make a bit longer spaces between sentences.  And two spaces look more attractive in plain text.

vzh (changes requested):

utils/renum/src/lepton-renum.c
graahnul-grom: OK, reverted extra space.
Looks like a typo.

graahnul-grom:
What about movig it to `contrib/`?
`lepton-refdes_renum` is identical, plus,
it has unit tests.

vzh:
I don't mind.

vzh (approved):

PR #751, opened 2021-02-18
Reorganize symbols/

graahnul-grom:
Make the contents of the `symbols/` directory
look like this:
```
symbols/
|-- docs/
|-- scheme/
|-- sym-gnetman/
|-- sym-verilog/
|-- sym-vhdl/
|-- sym/
  |-- 4000/
  |-- 74/
  |-- allegro/
  |-- ...
```

That is, we do not mix symbols with files of other types.
This will allow us to simplify installation commands
in `Makefile`s.

Affects #732.


vzh:
:+1:

vzh (approved):

PR #754, opened 2021-02-18
Delete superfluous gafrc files in netlist analog examples

graahnul-grom:
Remove unneeded (and erroneous) `gafrc` files,
which were, apparently, mistakenly copied/pasted
from the `verilog` example.


vzh (approved):
The better approach would be to make the example self-sufficient by adding appropriate symbols to the local library.  Some users don't use stock symbol libraries at all.

graahnul-grom:
Yes, like we did with out `examples/` in #479.
IMHO, the more important task would be to make
`liblepton` and `netlist` unit tests self-sufficient
in this sense (if at all possible).

PR #756, opened 2021-02-18
Move lepton-renum to contrib/

graahnul-grom:
Affects #732.


contrib/renum/src/Makefile.am
```diff
@@ -1,4 +1,4 @@
-bin_PROGRAMS = lepton-renum
+noinst_PROGRAMS = lepton-renum
```
vzh, line 1: I cannot remember/understand why it should be so. On my system, I cannot install things from `contrib/` even if I specify `--enable-contrib` on the configure stage.

graahnul-grom: I can't remember, too.
It seems `contrib/` tools were never meant to be installed:
look at the git history of `Makefile.am`'s.
`noinst_` has always been there.

vzh: OK, I see.  That pretty clear means the `--enable-contrib` option is useless and the `noinst_` prefix is abused since it was designed for supplementary tools that help to build other tools, if I understood correctly from autotools' documentation.

vzh (approved):

contrib/renum/src/Makefile.am
vzh: And, apropos, I recall why I agreed with that status of tools in `contrib/`.  Because their names still conflict with those in the `geda-gaf` package.

graahnul-grom:
We should amend the `--enable-contrib` description
`in README.md` ("Noticeable configure options"):

This option enables build ~~and installation~~ of several contributed tools...

vzh:
Yup

PR #757, opened 2021-02-19
gui: Fix behaviour of the lower combobox in the Component select dialog.

vzh:
I broke it myself in the commit e542a0a before 1.9.12.  While
getting rid of a wrong and deprecated type was OK, I shouldn't
have changed the number of callback arguments as the signal
"changed" requires exactly two.

### DIFF
--- a/libleptongui/src/x_compselect.c
+++ b/libleptongui/src/x_compselect.c
@@ -683,10 +683,12 @@ compselect_callback_filter_button_clicked (GtkButton *button,
  *  that the requested behavior for the next adding of a component has
  *  been changed.
  *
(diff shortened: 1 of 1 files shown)


graahnul-grom (approved):
And I thought what was that...
I try to "include component as individual objects",
but it works only the *second time* I place a component.
Thank you.

vzh:
And this fix is worth of a new release.  WDYT?

graahnul-grom:
It's not a serious bug. It is rather an annoyance.
Though we could make a release nevertheless.
But I would complete #732 before that.

vzh:
OK.

PR #758, opened 2021-02-19
Rename files and some functions dealing with component objects

vzh:
- Files `geda_component_object.X` have been renamed. The `geda_` prefix has been dropped.
- Several functions have been renamed to match the new object structure name.
- More checks for invalid use in component object functions.
- The component object field 'component_embedded' has been moved to `COMPONENT` and renamed. 
- Added accessors for this field to no longer use it directly.
- Many little fixes: comments, whitespaces, indentation.

### DIFF
--- a/liblepton/include/Makefile.am
+++ b/liblepton/include/Makefile.am
@@ -1,5 +1,6 @@
 libleptonincludedir = $(includedir)/liblepton
 libleptoninclude_HEADERS = \
+	liblepton/component_object.h \
 	liblepton/defines.h \
 	liblepton/edaconfig.h \
 	liblepton/edaerrors.h \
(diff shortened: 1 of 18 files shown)


graahnul-grom:
`geda_component_object.c => component_object.c`

While renaming files, didn't you consider dropping the
prefix from the functions/structures names as well?
Or do you think it's safer to use some prefix for the
identifiers in source code? (After all, it's not C++, where
we could use namespaces).

graahnul-grom (approved):

vzh:
I believe at least functions available for dlopening should have a unique prefix. And here it is `lepton`.

graahnul-grom:
Good point.

vzh:
Just found after renaming `geda_string.h` to `string.h` that such short names may conflict with common system headers which leads to compilation errors.  It seems it's safer to have a unique prefix for files in Lepton.  WDYT?

PR #770, opened 2021-03-09
Collapsible "Add Attribute" section in the "Edit Attributes" dialog

graahnul-grom:
Replace the static bottom "Add Attribute" frame in
the attributes editor dialog with a collapsible
section widget.
Most of the time this section is not used and,
on the other hand, often there's not enough room
to display all of the attributes (especially if
the "Show inherited attributes" check box is on).
The "expanded" state of the bottom section is
preserves across the dialog invocations and
program runs.

old version - static frame:
multiattr_old

new version - section expanded:
multiattr_new_1

new version - section collapsed:
multiattr_new_2

### DIFF
--- a/libleptongui/include/x_multiattrib.h
+++ b/libleptongui/include/x_multiattrib.h
@@ -69,6 +69,8 @@ struct _Multiattrib {
   GdkColor       not_present_in_all_text_color;
 
   gulong object_list_changed_id;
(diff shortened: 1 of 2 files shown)


vzh (commented):
:+1:

vzh (approved):

PR #774, opened 2021-03-19
Rewrite some functions from (lepton core object) using Scheme FFI

vzh:
Those functions are `object?()`, `object-type()`, `object-type?()`, `object-id()`.  Type checkers such as `arc?()` and the like have been redefined as well.  Plus, I added some more tests for the functions.  And docstrings you could refer in Lepton's REPL.

### DIFF
--- a/liblepton/scheme/Makefile.am
+++ b/liblepton/scheme/Makefile.am
@@ -141,6 +141,7 @@ TESTS = \
 	unit-tests/lepton-object-stroke.scm \
 	unit-tests/lepton-object-text.scm \
 	unit-tests/lepton-object-transform.scm \
+	unit-tests/lepton-object-wrong.scm \
 	unit-tests/lepton-os-basic.scm \
 	unit-tests/lepton-os-expand-env-variables.scm \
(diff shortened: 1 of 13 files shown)


graahnul-grom:
Are you using `'x` and `'a` in unit tests just
as invalid values?

```scheme
(test-assert (not (object-type? a 'x)))
(test-assert (not (object-type? 'a 'a)))
```

graahnul-grom:
Don't you think it's better to not use `geda-object`
in new names, e.g. `lepton-object->pointer` instead
of `geda-object->pointer` etc.?

graahnul-grom (approved):

vzh:
> Are you using `'x` and `'a` in unit tests just
> as invalid values?

Yes, exactly.

vzh:
> Don't you think it's better to not use `geda-object`
> in new names, e.g. `lepton-object->pointer` instead
> of `geda-object->pointer` etc.?

Currently, object smobs created in C code have prefix `geda`, hence the naming.  Since my intention is to eliminate smobs at all some time, I think new name would be better to introduce when the code will deal with foreign objects directly.

PR #778, opened 2021-03-23
Add accessors for picture object struct members

vzh:


graahnul-grom (approved):
Why don't you use accessors
for the `st_picture::pixbuf` field?

vzh:
1) I don't need them for Scheme
2) I've got just tired :-)  I think it can be done some time later

graahnul-grom:
:)

PR #779, opened 2021-03-23
Add accessors for the 'CHANGED' flag of LeptonPage.

vzh:
The functions are needed for Scheme FFI code.


graahnul-grom (approved):
> The functions are needed for Scheme FFI code.

IMHO, ideally, the `CHANGED` field should not be
accessed directly anywhere.

PR #782, opened 2021-03-25
Get rid of functions in o_embed.c.

vzh:
Embedding code has been splitted up and transferred to appropriate
files.  The number of functions for pictures has been reduced by
merging them.


liblepton/src/component_object.c
```diff
@@ -1514,3 +1514,78 @@ lepton_component_object_set_missing (const LeptonObject *object,
...
+ *  \param object The #LeptonObject to embed
+ */
+void
+lepton_component_object_embed (LeptonObject *object)
+{
+  LeptonPage *page;
+
+  g_return_if_fail (lepton_object_is_component (object));
+
+  page = lepton_object_get_page (object);
```
graahnul-grom, line 1533: Shouldn't we check here that `page != NULL`?

graahnul-grom (approved):

liblepton/src/component_object.c
vzh: Thank you for the catch!  Apparently, I've confused the func with `lepton_object_page_set_changed()` which does the check inside it.  I'll add the fix soon.

vzh:
@graahnul-grom, please check if anything is OK now.

graahnul-grom:
Thank you, everything is OK.
And indeed, it's possible to forget to `page-append!()`
a created component in script.
If we then call `set-object-embedded!()`, we'll trigger
the assertion in `lepton_page_set_changed()`,
for example:

```scheme
( define comp
    ( make-component/library
        "foo.sym"
        ( cons 500 500 )
        0
        #f
        #f
    )
)

; forget to call here: ( page-append! (active-page) comp )
( set-object-embedded! comp #t ) ; => assertion failed
```
log:
```
Component [foo.sym] has been embedded.
lepton_page_set_changed: assertion 'page != NULL' failed
```

PR #785, opened 2021-04-01
Rewrite three SCM_DEFINE'd arc functions in Scheme

vzh:
- The functions `arc-info()`, `make-arc()`, and `set-arc!()` have been rewritten using Scheme FFI
- Many C functions related to arc objects (accessors and other) have been renamed and now have the prefix `lepton_arc_object_`.
- Some of the C functions have been simplified.
- Lots of new tests for arcs have been added.

(Well, there's always room for improvements, e.g. I don't like docstrings I've written).


graahnul-grom (approved):
Also, I've tested recent changes (including this PR)
under `cygwin`. All works fine. Though some
`.gitignore` files needs to be amended.

PR #789, opened 2021-04-13
Several little clean-ups

vzh:
- Fix a comment in `st_object` definition.
- Get rid of unused enum values belonging to strokes.
- Get rid of two obsolete and unused fields of `st_object`.
- Fix one warning when compiling with gtk3 support.

### DIFF
--- a/liblepton/include/liblepton/object.h
+++ b/liblepton/include/liblepton/object.h
@@ -66,10 +66,6 @@ struct st_object
   /* 1 for up, -1 for down (vertial bus) */
   int bus_ripper_direction;             /* only valid on buses */
 
-
-  int font_text_size;                   /* used only with fonts defs */
-  GList *font_prim_objs;                /* used only with fonts defs */
-
   int whichend;    /* for pins only, either 0 or 1 */
   int pin_type;    /* for pins only, either NET or BUS */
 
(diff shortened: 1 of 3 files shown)


graahnul-grom (approved):
Did `gcc` warn you of unused structure's fields?

vzh:
It was `g++`.  I did something like `./configure --with-gtk3 CC=g++ && make CC=g++ all` after modifying this line in `configure.ac`:
> AX_GCC_FLAGS([-Wall -Werror=uninitialized -Wno-deprecated-declarations -std=c99])

to filter out gtk3 deprecated stuff.  There was one more error, though I decided not to deal with it yet.

graahnul-grom:
> `-Wno-deprecated-declarations`

I'm using `-DGLIB_DISABLE_DEPRECATION_WARNINGS=1`
in `CPPFLAGS` to filter out irrelevant warnings.

vzh:
Looking into it, I found another interesting macro: [GLIB-VERSION-MIN-REQUIRED](https://developer.gnome.org/glib/stable/glib-Version-Information.html#GLIB-VERSION-MIN-REQUIRED).  After about glib 2.61 there are tons of repeating deprecation warnings on my system.  I'm curious if we could suppress them by using it.

PR #791, opened 2021-04-14
Move two fields of LeptonObject to LeptonText

vzh:
- Move two fields of `LeptonObject`, namely `show_name_value` and `visibility`, to `LeptonText` and shorten the name of the former one.
- Move accessors to them to `text_object.c`.
- Leptonize names of some text object related functions.
- Rename the `st_text` struct to `_LeptonText`.


graahnul-grom (approved):
You're going to rename all of the `st_*` structs?

vzh:
Sure.  And as with some other changes, I already did it several times in various branches.  Just need to do it twice or more times to nail down the success.

PR #795, opened 2021-04-28
Rewrite object fill related functions using Scheme FFI

vzh:
- Introduce new type for object filling: `LeptonFill`.  It encapsulates all filling properties for which several fields of `LeptonObject` were previously used.
- Add accessors for those properties.
- As always, get rid of Scheme-in-C code related to object filling.
- Add more tests.
- Add docstrings to new Scheme functions.

### DIFF
--- a/liblepton/src/fill.c
+++ b/liblepton/src/fill.c
@@ -22,8 +22,8 @@
 #include "liblepton_priv.h"
 
 /*!
- * \file  fill_type.c
- * \brief Filling types.
+ * \file  fill.c
+ * \brief Functions for working with object filling.
  */
 
 
--- a/liblepton/include/Makefile.am
+++ b/liblepton/include/Makefile.am
@@ -22,7 +22,7 @@ libleptoninclude_HEADERS = \
 	liblepton/component.h \
 	liblepton/coord.h \
 	liblepton/forward.h \
-	liblepton/fill_type.h \
+	liblepton/fill.h \
 	liblepton/line.h \
 	liblepton/list.h \
 	liblepton/page.h \
--- a/liblepton/include/liblepton/fill.h
+++ b/liblepton/include/liblepton/fill.h
@@ -17,7 +17,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
  */
-/*! \file fill_type.h
+/*! \file fill.h
  */
 
 G_BEGIN_DECLS
--- a/liblepton/include/liblepton/fill_type.h
+++ b/liblepton/include/liblepton/fill_type.h
@@ -38,6 +38,68 @@ enum _LeptonFillType
 
 typedef enum _LeptonFillType LeptonFillType;
 
+typedef struct _LeptonFill LeptonFill;
+
+struct _LeptonFill
+{
+  LeptonFillType type;
+  int width;
+  int pitch1;
+  int angle1;
+  int pitch2;
+  int angle2;
+};
+
+
+
+LeptonFill*
+lepton_fill_new ();
+
+void
+lepton_fill_free (LeptonFill *fill);
+
+LeptonFillType
+lepton_fill_get_type (const LeptonFill *fill);
+
+void
+lepton_fill_set_type (LeptonFill *fill,
+                      LeptonFillType type);
+int
+lepton_fill_get_width (const LeptonFill *fill);
+
+void
+lepton_fill_set_width (LeptonFill *fill,
+                       int width);
+int
+lepton_fill_get_pitch1 (const LeptonFill *fill);
+
+void
+lepton_fill_set_pitch1 (LeptonFill *fill,
+                        int pitch);
+int
+lepton_fill_get_angle1 (const LeptonFill *fill);
+
+void
+lepton_fill_set_angle1 (LeptonFill *fill,
+                        int angle);
+int
+lepton_fill_get_pitch2 (const LeptonFill *fill);
+
+void
+lepton_fill_set_pitch2 (LeptonFill *fill,
+                        int pitch);
+int
+lepton_fill_get_angle2 (const LeptonFill *fill);
+
+void
+lepton_fill_set_angle2 (LeptonFill *fill,
+                        int angle);
+const char*
+lepton_fill_type_to_string (LeptonFillType type);
+
+LeptonFillType
+lepton_fill_type_from_string (char *s);
+
 gboolean
 lepton_fill_type_draw_first_hatch (int fill_type);
 
--- a/liblepton/include/liblepton/liblepton.h
+++ b/liblepton/include/liblepton/liblepton.h
@@ -32,7 +32,7 @@ G_BEGIN_DECLS
 #include <liblepton/defines.h>
 
 #include <liblepton/color.h>
-#include <liblepton/fill_type.h>
+#include <liblepton/fill.h>
 #include <liblepton/stroke.h>
 #include <liblepton/point.h>
 #include <liblepton/str.h>
--- a/liblepton/include/liblepton/object.h
+++ b/liblepton/include/liblepton/object.h
@@ -47,10 +47,8 @@ struct st_object
   /* Visible appearance of lines in graphical primitives. */
   LeptonStroke *stroke;
 
-  LeptonFillType fill_type;
-  int fill_width;
-  int fill_angle1, fill_pitch1;
-  int fill_angle2, fill_pitch2;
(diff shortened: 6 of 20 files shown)


graahnul-grom:
It seems that `pitch` value of 0 is not allowed.
Setting it to 0 gives many warnings like this:

```
CRITICAL **: m_hatch_polygon: assertion 'pitch>0' failed
```

And if I set it to 0 for *circle* objects, app crashes with
```
Floating point exception
```

Can you reproduce it?

graahnul-grom:
Aha, divide by zero in `calculate_initial_sweep()` (liblepton/src/m_hatch.c).

vzh:
Good point and a good cause for an argument check in that function.

graahnul-grom:
For boxes and paths there's just a warning (`m_hatch_polygon()`
has a check:`g_return_if_fail( pitch > 0 )`).
But `m_hatch_circle()` doesn't check its `pitch` arg and crashes app.

graahnul-grom:
What default value should be returned, then, from
`lepton_object_get_fill_pitchX()` (now it's 0)?

graahnul-grom:
And something has to be done in GUI, e.g. an error message
box, in case user tries to enter 0 in the `pitch` fields in the
object properties dialog.

vzh:
I think default width should be 10 as for lines, pitch should be 100, and angles 45 and 135. WDYT?

vzh:
And yes, some prompt in GUI would be nice to have.

graahnul-grom:
> I think default width should be 10 as for lines, pitch should be 100, and angles 45 and 135.

Yes.

liblepton/src/fill.c
```diff
@@ -0,0 +1,277 @@
...
+
+/*! \brief Get the width of a fill.
+ *
+ *  \param [in] fill The fill.
+ *  \return The width of lines of the fill.
+ */
+int
+lepton_fill_get_width (const LeptonFill *fill)
+{
+  g_return_val_if_fail (fill != NULL, FILLING_HOLLOW);
```
graahnul-grom, line 93: Should return `int`, not a `LeptonFillType` enum value.

liblepton/src/fill.c
```diff
@@ -0,0 +1,277 @@
...
+
+/*! \brief Get the first pitch of a fill.
+ *
+ *  \param [in] fill The fill.
+ *  \return The first pitch of the fill.
+ */
+int
+lepton_fill_get_pitch1 (const LeptonFill *fill)
+{
+  g_return_val_if_fail (fill != NULL, FILLING_HOLLOW);
```
graahnul-grom, line 121: The same as above: should return `int`, not a `LeptonFillType` enum value.

liblepton/src/fill.c
```diff
@@ -0,0 +1,277 @@
...
+
+/*! \brief Get the second pitch of a fill.
+ *
+ *  \param [in] fill The fill.
+ *  \return The second pitch of the fill.
+ */
+int
+lepton_fill_get_pitch2 (const LeptonFill *fill)
+{
+  g_return_val_if_fail (fill != NULL, FILLING_HOLLOW);
```
graahnul-grom, line 176: The same as above: should return int, not a LeptonFillType enum value.

vzh:
> It seems that `pitch` value of 0 is not allowed.
> Setting it to 0 gives many warnings like this:
> 
> ```
> CRITICAL **: m_hatch_polygon: assertion 'pitch>0' failed
> ```
> 
> And if I set it to 0 for _circle_ objects, app crashes with
> 
> ```
> Floating point exception
> ```
> 
> Can you reproduce it?

I can.  

What I cannot understand is what the code `((delta - pitch) / pitch * pitch))` does.  Looks like some rounding.
If I do change it to:
```
static gint calculate_initial_sweep(gint pitch, gint min_y, gint max_y)
{
  return min_y + pitch / 2;
}
```
it does visually mostly the same.  Hmm.

BTW, I found many assertions, for filling and stroke, are yielded when you add a new path :-/  Seems the defaults should be changed anyways.

graahnul-grom:
`min_y + pitch / 2` yields different result from the original
expression if used with not-so-round coordinates, e.g.:
```
max_y = 250;
min_y = 125;
pitch = 75;

int res1 = min_y + ( (delta - ( (delta - pitch) / pitch * pitch) ) / 2 );
int res2 = min_y + pitch / 2;
printf( "res1: [%d], res2: [%d]\n", res1, res2 );

res1: [187], res2: [162]

```

graahnul-grom:
Let's try to reproduce the algorithm.

pitch

```c
int calculate_initial_sweep( int pitch, int min_y, int max_y )
{
  int delta = max_y - min_y;
  return min_y + ((delta - ((delta - pitch) / pitch * pitch)) / 2);
}

int telepathie( int pitch, int min_y, int max_y )
{
    const int delta = max_y - min_y;

    int N = delta / pitch;
    int H = pitch * ( N - 1 );
    int delta_H = delta - H;
    int X = min_y + delta_H / 2;

    return X;
}

void tst( int pitch, int min_y, int max_y )
{
    int res1 = calculate_initial_sweep( pitch, min_y, max_y );
    int res2 = telepathie( pitch, min_y, max_y );

    printf( ".. res1, res2: %d %d => [%s]\n",
            res1, res2, res1 == res2 ? "v" : "X" );
}
```

graahnul-grom:
Though it gives wrong results in the corner case:
when `pitch > delta`.

graahnul-grom:
In that case I would return `delta / 2`.

```c
int telepathie( int pitch, int min_y, int max_y )
{
    const int delta = max_y - min_y;

    if ( pitch >= delta )
        return delta / 2;

    int N = delta / pitch;
    int H = pitch * ( N - 1 );
    int delta_H = delta - H;
    int X = min_y + delta_H / 2;

    return X;
}
```

vzh:
> Aha, divide by zero in `calculate_initial_sweep()` (liblepton/src/m_hatch.c).

The bug is present in the current master and I think it was there for years.

I believe, this can be solved in three steps:
- in the function `calculate_initial_sweep()` we could just get rid of `/ pitch * pitch` if `pitch == 0` by adding corresponding condition;
- prohibit setting of zero pitch in appropriate setters;
- initialize default values of `LeptonFill` members to values we discussed above.

ehennes775:
Hello,

Not sure what the end goal is, but the pitch cannot be zero because it will create an infinite loop in the line sweep algorithms.

int calculate_initial_sweep( int pitch, int min_y, int max_y )
{
  int delta = max_y - min_y;
  return min_y + ((delta - ((delta - pitch) / pitch * pitch)) / 2);
}
- The following calculates the maximum distance that needs to be hatched.
	md = (delta - pitch)

- The following calculates the actual distance that needs to be hatched, the perpendicular distance between the farthest hatch lines. Dividing by the pitch and then multiplying by the pitch makes the result an even multiple of the pitch by integer truncation.
	ad = md / pitch * pitch

- The following calculates the leftover distance on each side:
	l2 = delta - ad

- The following calculates the leftover distance on one side:
	l = l2 / 2

The following calculates the coordinate of the first pitch line:
	min_y + l

Cheers,
Ed

ehennes775:
Hello,

Not sure what the end goal is, but the pitch cannot be zero because it will create an infinite loop in the line sweep algorithms.

```c
int calculate_initial_sweep( int pitch, int min_y, int max_y )
{
  int delta = max_y - min_y;
  return min_y + ((delta - ((delta - pitch) / pitch * pitch)) / 2);
}
```

- The following calculates the maximum distance that needs to be hatched.
	md = (delta - pitch)

- The following calculates the actual distance that needs to be hatched, the perpendicular distance between the farthest hatch lines. Dividing by the pitch and then multiplying by the pitch makes the result an even multiple of the pitch by integer truncation.
	ad = md / pitch * pitch

- The following calculates the leftover distance on each side:
	l2 = delta - ad

- The following calculates the leftover distance on one side:
	l = l2 / 2

The following calculates the coordinate of the first pitch line:
	min_y + l

Cheers,
Ed

vzh:
@ehennes775, Thank you for explanation!  I'll add your comments to the code.  In the recently added commit da2fa26 I've added the check that pitch is non-zero.  If the check fails (which should no longer happen, I hope), it returns one half of delta.  The reason for this discussion was long standing bug crashing the app with `Floating point exception` when the user entered zero value for one of the pitches in the Options dialog.  Limiting values in the pitch entry will need more work.  For now, we're trying to make `liblepton`'s code more sane in this regard by just prohibiting appearance of zero pitch values in the underlying functions.  (Glad to see your comments, Ed :-) )

graahnul-grom (approved):

PR #796, opened 2021-05-05
Introduce Scheme module (lepton object type)

vzh:
- The new module includes functions for object type checking.
- Fixed a bug in the syntax rule `geda-object->pointer*()` that led to crashes of object type checking functions if an object of wrong type was used as the argument.
- Added many automatic tests including those making the above bug explicit.
- Added or improved docstrings for the functions in the new module.


liblepton/scheme/lepton/object/type.scm
```diff
@@ -0,0 +1,185 @@
...
+   ((net? object) 'net)
+   ((path? object) 'path)
+   ((picture? object) 'picture)
+   ((pin? object) 'pin)
+   ((text? object) 'text)
+   (else (error "Object ~A has bad type '~A'"
+                object
+                (integer->char (lepton_object_get_type pointer))))))
+
+(define (object-type? object type)
```
graahnul-grom, line 179: From the name `object-type?` I would guess that the function
takes one symbol argument and checks if it's a valid type name.
`object-type-is?()`?

vzh: Well, this was implemented long ago and was in the code and docs for years.  IMO, decisions on naming were not always TRT.  There are Scheme functions, for example, that have `object` as part of their names and those work with *any* Scheme objects.  More TLC is needed to develop a new naming scheme for lepton functions and gradually work on replacing and deprecation of old named ones.

PR #797, opened 2021-05-13
Rewrite Scheme-in-C functions related to linear primitives using Scheme FFI

vzh:
- New accessors for necessary object fields have been added.
- Much more tests have been implemented to keep the functionality intact and anything was tested before applying changes in the code.
- 7 functions have been rewritten.  They all have docstrings now.

### DIFF
--- a/liblepton/include/liblepton/object.h
+++ b/liblepton/include/liblepton/object.h
@@ -337,4 +337,7 @@ lepton_object_emit_pre_change_notify (LeptonObject *object);
 void
 lepton_object_emit_change_notify (LeptonObject *object);
 
+int
+lepton_object_get_whichend (LeptonObject *object);
+
 G_END_DECLS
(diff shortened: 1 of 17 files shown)


graahnul-grom (approved):
Do we have symbolic constants somewhere for the `whichend`
values, or just `0` and `1`?
Where is it documented what does 0 or 1 designate?

vzh:
[Here](https://lepton-eda.github.io/lepton-manual.html/pin.html#pin).  However, it's an obsolete work around for old versions of gEDA file format in which there was no convention of which pin side should be connectable.  There are old schematics in the repository where `whichend` is not used and C parser code specifies `-1` for the field in this case (AFAIR).  @peter-b simplified this convention on Scheme side.

vzh:
I am not aware of any symbolic constants for `whichend`.

PR #801, opened 2021-05-15
Rewrite text object related functions in (lepton object) using Scheme FFI

vzh:
The same changes as for other objects, e.g. in #797, though much more of them :-)


graahnul-grom (approved):
:+1:

PR #805, opened 2021-06-12
Fix visual representation of arcs.

vzh:
The one-liner fixes a long standing bug in arc representation on screen and in exported raster graphics reported several years ago by @NoSuchProcess for gschem on launchpad: [here](https://bugs.launchpad.net/geda/+bug/1504277).  There is a test symbol there to check how it looks like.

I attach here my own schematic and PNG files to test how it looks like before and after the change:
- [7400.sch.gz](https://github.com/lepton-eda/lepton-eda/files/6642469/7400.sch.gz)

- before:
7400-before

- after: 
7400-after

### DIFF
--- a/liblepton/src/edacairo.c
+++ b/liblepton/src/edacairo.c
@@ -297,7 +297,7 @@ eda_cairo_arc (cairo_t *cr, int flags,
   cairo_device_to_user_distance (cr, &offset, &dummy);
   cairo_device_to_user_distance (cr, &s_radius, &dummy);
 
-  do_arc (cr, s_x + offset, s_y + offset,
+  do_arc (cr, s_x + offset, s_y - offset,
           s_radius, start_angle, sweep_angle);
 }
 


graahnul-grom (approved):

graahnul-grom:
It might be related, at some zoom levels graphics elements
(lines, circles, etc.) gets shifted a bit, which makes schematic
look ugly.
For example, here's a component original look:
zoom1

Zoom out, and it becomes distorted:
zoom2

Zoom out one more time, and it's back to normal, etc.

graahnul-grom:
BTW, `zoom-gain` doesn't affect this behavior.

vzh:
Well, I think it's because the code tries to justify image on screen taking into account the current resolution.  The cairo code by @peter-b has been written to avoid one-pixel shift in case of small zooming.  We cannot deal with tools that do post-processing of already created screen image, with the tools that just process what the wm gives them.  The issue with one-pixel shift is visible in some cases because the renderer code needs sometimes to decide how many pixels to render *on-screen*, one or two, two or three.  (Really, looking at your pictures, I'm in doubt if I'm not wrong :-))

NoSuchProcess:
I'm not quite sure about the current lepton's arc renderer, but in the original gEDA renderer the bug at that time was as follows:
The renderer transforms all shape nodes in world coordinates to screen coordinates, then put them into integer positions, to prevent smoothing less-than-1px-width lines. This worked perfectly for lines, but there was an issue in case of the circles and arcs: these shapes are defined by center position and diameter. AFAIK cairo assumes that center of pixel at (x, y) has coordinates (x.5, y.5), so one should round the center position (and radius?) towards the 0.5 screen units, not the integer screen units. As far as I rememmer, these cracky connections occurs when diameter rounds to odd screen length.

To be noted, if you look at the down-oriented 7400 shape on the "after" image, you'll see that this particular shape has the radius of bow is approximately 1-pixel larger, than on the other shapes.

peter-b:
I know that when I was integrating the prototype hinting renderer that Peter Clifton wrote I introduced some hinting bugs. It's totally possible that the arcs/circles need tweaking. It's been 12 years though so my memory is hazy…

On 13 June 2021 12:04:26 BST, Vladimir Zhbanov ***@***.***> wrote:
>Well, I think it's because the code tries to justify image on screen
>taking into account the current resolution.  The cairo code by @peter-b
>has been written to avoid one-pixel shift in case of small zooming.  We
>cannot deal with tools that do post-processing of already created
>screen image, with the tools that just process what the wm gives them. 
>The issue with one-pixel shift is visible in some cases because the
>renderer code needs sometimes to decide how many pixels to render
>*on-screen*, one or two, two or three.  (Really, looking at your
>pictures, I'm in doubt if I'm not wrong :-))
>
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/lepton-eda/lepton-eda/pull/805#issuecomment-860192325

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

vzh:
@NoSuchProcess Thank you! Lepton's rendering code hasn't changed much since forking from `geda-gaf`.  If you have any ideas/code of how to improve it, it would be much appreciated as well as patches.
@peter-b I very much value your work and Peter Clifton's work on the renderer (not only on it, really, you drove the whole gEDA project at that time).  Really, it was the hard work, IMO.  After writing my previous message here I recalled Peter C. did the initial work.  Too late, I'm sorry for me not to mention him here :-)

PR #810, opened 2021-06-19
Fix 'make check' under root account

graahnul-grom:
When running 'make check' under root account, skip
two tests that tries to read config file with 000
permissions (and expected to fail).
However, root user still can read such a file, so
run those test only under normal user account.


vzh (approved):
JFYI, there is a simpler function: `zero?`.  `(zero? something)` is equal to `(eq? 0 something)`.  Not sure it is relevant here though :-)
Another little note. The `if` function usually has two branches.  While it is not so much relevant here as well, in some cases it may beat you, e.g. if you expect some result from it.   Therefore, in cases where you don't anticipate the result to be used somewhere else, using of `when` or `unless` is mentally more convenient as the users don't expect to have any results from them.

graahnul-grom:
Good point, `when` would be better here.

PR #812, opened 2021-06-22
Separate snap size and grid size status bar labels

graahnul-grom:
Separate snap size and grid size status bar labels, draw user's
attention by setting snap size label text decoration and color
when grid size is differs from the value set in configuration
(how many times did I screw up my symbols while grid size
was not set properly!).

old:

new:


vzh (approved):

vzh:
It would be very helpful, if you continued the effort a bit farther and added a couple of words about the feature into the manual :-) A PR is not necessary, you know.  The same wrt NEWS ;-)

PR #813, opened 2021-06-24
Rewrite object-connections() using Scheme FFI

vzh:
This was the last function defined in the `(lepton core object)` module.  Now, it has been rewritten in Scheme as all others functions from the module.  As my first naive approach of parsing *GList*s in Scheme appeared to be significantly slower than the corresponding C code, I utilized *memoization* for the function using the Guile facility named `object-property`, which restored the speed of test suite passing.

As always, some new tests have been added, and the docstring for the function has been added.

### DIFF
--- a/liblepton/scheme/lepton/object.scm
+++ b/liblepton/scheme/lepton/object.scm
@@ -41,6 +41,7 @@
             object-color
             set-object-color!
             object-component
+            object-connections
             object-embedded?
             set-object-embedded!
             object-id
@@ -174,7 +175,26 @@ Returns OBJECT."
   object)
 
 
-(define-public object-connections %object-connections)
+(define (object-connections object)
+  "Returns a list of other objects that are directly connected to
+OBJECT.  If OBJECT is not included in a page, raises an
+'object-state error.  The connections reported are independent of
+inclusion in components and includes only primitive objects such
+as pins, nets, or buses."
+  (define pointer (geda-object->pointer* object 1))
+
+  (when (null-pointer? (lepton_object_get_page pointer))
+    (scm-error 'object-state
+               'object-connections
+               "Object ~A is not included in a page."
+               (list object)
+               '()))
+
+  (let* ((gls (s_conn_return_others %null-pointer pointer))
+         (ls (glist->object-list gls)))
+
+    (g_list_free gls)
+    ls))
 
 
 (define (object-component object)
--- a/liblepton/scheme/lepton/object.scm
+++ b/liblepton/scheme/lepton/object.scm
@@ -33,6 +33,7 @@
 
   #:use-module (lepton color-map)
   #:use-module (lepton ffi)
+  #:use-module (lepton object foreign)
   #:use-module (lepton object type)
 
   #:export (copy-object
--- a/liblepton/scheme/lepton/object/foreign.scm
+++ b/liblepton/scheme/lepton/object/foreign.scm
@@ -1,0 +1,37 @@
+;;; Lepton EDA library - Scheme API
+;;; Copyright (C) 2021 Lepton EDA Contributors
+;;;
+;;; This program is free software; you can redistribute it and/or modify
+;;; it under the terms of the GNU General Public License as published by
+;;; the Free Software Foundation; either version 2 of the License, or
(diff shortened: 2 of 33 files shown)


liblepton/scheme/lepton/object.scm
```diff
@@ -33,13 +33,15 @@
 
   #:use-module (lepton color-map)
   #:use-module (lepton ffi)
+  #:use-module (lepton object foreign)
   #:use-module (lepton object type)
```
graahnul-grom, line 38: I think now you can safely remove
`#:use-module (lepton core object)`

graahnul-grom:
Memoize == "Memorize"?

liblepton/scheme/lepton/object/foreign.scm
```diff
@@ -0,0 +1,51 @@
...
+  ;; Return #f if the pointer is wrong.
+  (false-if-exception (pointer->scm (edascm_from_object pointer))))
+
+(define (geda-object-pointer? pointer)
+  (true? (edascm_is_object pointer)))
+
+(define (glist->object-list gls)
+  "Converts a GList of foreign object pointers GLS into a Scheme
+list.  Returns the Scheme list of objects."
+  (let loop ((gls gls)
```
graahnul-grom, line 46: Where can I find documentation on the obscure DB `loop` construct?

liblepton/scheme/netlist/schematic-connection.scm
```diff
@@ -100,8 +101,12 @@ Example usage:
                  (_ #\?)))
              args)))))
 
+;;; This is a helper object property to get object connections
+;;; once.
+(define object-conns (make-object-property))
+
 (define (connected-to? object1 object2)
-  (not (not (memv object1 (object-connections object2)))))
+  (not (not (memv object1 (object-conns object2)))))
```
graahnul-grom, line 109: Isn't double `not` superfluous here?

liblepton/scheme/netlist/schematic-connection.scm
```diff
@@ -100,8 +101,12 @@ Example usage:
                  (_ #\?)))
              args)))))
 
+;;; This is a helper object property to get object connections
+;;; once.
+(define object-conns (make-object-property))
```
graahnul-grom, line 106: How does this speed up the test suite?
BTW, now `make distcheck` runs for 6 minutes
on my machine instead of 5.

graahnul-grom (approved):
Congratulations on getting rid of
the `(lepton core object)` module! :-)

liblepton/scheme/lepton/object.scm
vzh: Yes, I'll do it in the next patchset.

vzh:
> Memoize == "Memorize"?

No, while some people try to write it in the latter form, the former one is a standard technical term that is known for decades in Lisp, and now in many other non-lisp languages.  Try to google for it, look at stackoverflow, or read the 5th chapter of "On Lisp" by Paul Graham.

liblepton/scheme/lepton/object/foreign.scm
vzh: The word `loop` is not a construct, its just what is called `named let` construct in Scheme/Lisp, the `let` construct that allows jumps by name like `goto` in other languages but much more efficient if it is correctly used.  Any other name may be used instead of `loop` and it will work anyways.

liblepton/scheme/netlist/schematic-connection.scm
vzh: Yes, in this particular case it is superfluous and may be omitted.  However, for user convenience, in case someone will debug the code in REPL, I wrote it so (for myself :-)). I believe, the guile compiler can deal with it nice and not profiled it to see how much it slows down the code ;-)

liblepton/scheme/netlist/schematic-connection.scm
vzh: Well, without this property, the code may work twice as long.  It is especially apparent with the `stack-torture.sch`.  Reset the branch to the state where the function `object-connections` has been actually rewritten, and you'll see what I'm talking about.

PR #815, opened 2021-06-25
Get rid of the module (lepton core deprecated)

vzh:
One more `core` module has been removed.  I'm not sure if I did it quite right.  Please let me know if anything could be improved, at least on C side.


liblepton/include/i_vars_priv.h
```diff
@@ -5,3 +5,16 @@ extern int default_keep_invisible;
...
+extern char _OBJ_BOX;
+extern char _OBJ_PICTURE;
+extern char _OBJ_CIRCLE;
+extern char _OBJ_NET;
+extern char _OBJ_BUS;
+extern char _OBJ_COMPONENT;
+extern char _OBJ_TEXT;
+extern char _OBJ_PIN;
+extern char _OBJ_ARC;
+extern char _OBJ_PLACEHOLDER;
```
graahnul-grom, line 20: Why do we have to re-define something in C at all?
In `liblepton/scheme/geda/deprecated.scm` just
`( define-public OBJ_LINE #\L )`, etc.

vzh: Because the variables with analogous names are still used in C code, and they were made up as such.  I first did as you suggested, and then thought for long time if this is a right decision.  Our C code should not conflict with Scheme.  Scheme code, since it uses the underlying `liblepton*` libraries should be sure the functions and objects are what it really relies on.  Thus, they should be the same in C and Scheme no matter how we work on them.  Imagine, for instance, me, adding some feature and changing the char name of some object in the process.  Nobody may notice this :-) I believe we surely will throw away the variables in time, though now we have to live with them.

Another approach would be generating code for those C and Scheme variables from same stem files having definitions for the variables.  If you do so, it'll be much better.  I had no time and wish to work in this direction.  Many other things are waiting for our TLC.  This may change some time, however :-)

graahnul-grom: OK, you're right.

graahnul-grom (approved):

PR #816, opened 2021-06-25
gui: Add gtk3 support in last modifications of lepton-schematic's bottom widget.

vzh:
The patch resolves the last warning in the build with g++ (I omit `GDK_SCROLL_SMOOTH` issue here as unresolved yet). The new code has been introduced in #812 so it needs corresponding changes for GTK3.

The changes I use in `configure.ac` to make sure our code is fully compatible with GTK3.0.0 and filter all other later GTK3 stuff are as follows:
```
AC_DEFINE(GDK_VERSION_MIN_REQUIRED, GDK_VERSION_3_0, [Ignore post 3.0 deprecations])
```

Of course, there are more things to be done in the gtk3 port code, and I have already done some work on this in my private repository.  I just believe this is a reasonable minimum we could start of and announce as a really working gtk3 port of Lepton.  Just GTK 3.0.0 yet :-)

Please let me know if something else should go in a (the?) new release which I'm planning to start working on just these days (hopefully, if nothing else in my private life stops me from that).

### DIFF
--- a/libleptongui/src/gschem_bottom_widget.c
+++ b/libleptongui/src/gschem_bottom_widget.c
@@ -492,11 +492,19 @@ set_snap_info_widget_color (GschemBottomWidget* widget,
 {
   g_return_if_fail (widget != NULL);
 
+#ifdef ENABLE_GTK3
+  GdkRGBA color;
+  gdk_rgba_parse (&color, color_name);
+  gtk_widget_override_color (GTK_WIDGET (widget->grid_snap_widget),
+                             GTK_STATE_FLAG_NORMAL,
+                             &color);
+#else
   GdkColor color;
   gdk_color_parse (color_name, &color);
   gtk_widget_modify_fg (GTK_WIDGET (widget->grid_snap_widget),
                         GTK_STATE_NORMAL,
                         &color);
+#endif
 }
 
 


graahnul-grom:
What's the issue with `GDK_SCROLL_SMOOTH`?

graahnul-grom (approved):

vzh:
The issue with `GDK_SCROLL_SMOOTH` is that gtk people could not managed to support so named `smooth` events from `XINPUT2` (IIRC) correctly, mostly due to ambiguity in the data delivered by the event procedures when the user starts smooth scrolling (the start point of scrolling). I've seen long discussions of programmers working on major browsers such as Firefox and Chrome on how to win this, and the answers of GTK developers who stated several years ago that they cannot do better and a better perspective is to wait until Wayland becomes common in distributions and then it will magically do TRT.  There are workarounds for the event in the code of the browsers and they were discussed for years.  I tried myself to get the events work right in the GTK3 version of `lepton-schematic` by trying to port the code of the browsers though there are still some bugs I could not win (yet?).  For example, in my code the canvas may shortly zoom in instead of zooming out when you start scrolling.  Support of the `smooth` events was basically intended for use in modern devices such as tablets etc (IMO).  However, in GTK3 it becomes the main mechanism for scrolling.

BTW, in GTK3 GUI I see zooming is somewhat better done with the scroll adjustment widgets rather than by processing of the scrolling events from mouse on the canvas.

graahnul-grom:
Bloody GTK3 and its "developers".
Of course, if their hands grow from the a$$, only Wayland
will help, mind you. Рекомендации лучших собаководов!
Не умеешь ни черта делать? Голова не соображает?
Ставь Wayland - всё заработает! :-))

vzh:
No, it's not GTK3 and its devs.  Other modern window managers / envs such as KWin and Plasma had the same issues and about during the same period of time.  It's a deficiency of the new X protocol invented for tablet devices.  Gtk devs just stated this fair.
And well, the devs of other apps have won the issue somehow, maybe by various tricks.  It's probably just me who couldn't :-)  OTOH, they struggled against it for years.  And a bit little different issue is how much is it now important for us... (Do you know anybody playing with gschem/lepton-schematic on phones/tablets?.. well, I'm just kidding.)

graahnul-grom:
*for years* (!)
What is that X protocol and how does it used in desktop (not tablet) apps?

vzh:
I mentioned it before, AFAIR it's `XINPUT2`.  Again, IIUC, the old scroll events such as `GDK_SCROLL_LEFT` etc. are considered obsolete in new GTK versions.  The new event `GDK_SCROLL_SMOOTH` has all the event info they delivered to the tools and a bit more since it is intended to be used for such devices as touch screens which, in turn, can deliver, say, two or three events of two fingers at once.  The inherent issue it has with non-touch devices is, IIUC, missing some info such as initial place and direction of moving at devices that cannot provide such info due to hardware or driver limitations.  Most of legacy PC's do not support such an extended initial information at hardware level, so the programmers had to work around this somehow in their projects.

As for *for years*, it's not so awful.  Look, for instance, at gEDA and Lepton.  We have some issues for decades while our project is not as big as GTK.

vzh:
Again, as X grows, some protocols are replaced with new ones.  Sometime `XINPUT` will die and be replaced with `XINPUT2`.  Some things will be thrown away or replaced by new code when/if Wayland gets the new native part of X.  I know, for example, it doesn't and is not going to support `XRANDR`.

vzh:
IMO, gtk has similar issues gEDA/Lepton have.  Not so many devs when there are lots of tools to support.  I think it is the reason they throw out some widgets/interfaces/code.  They delegate some things to third-party code, like cairo.  IMO, it's good in such a situation.  Look, we do the same :-)

graahnul-grom:
Maybe GTK has similar issues with Lepton,
but its developers is another matter.
Unlike them, we:
- maintain backward compatibility
- respect our users and take into account their opinions
- do not deliberately break API with every *micro* release in the name of some corporate bullshit

vzh:
Some notes:
- Our C API and ABI change much more significantly between releases.  Look at the versions of current `libgeda` and `liblepton` first.
- We **do** deliberately break our C API with almost all *micro* releases (`liblepton` et all).  It was never stable at least in our 1.9 series.  Surely, it  is not about our Scheme API, which may smooth things a bit.
- I believe there are much more GTK users to take into account their opinions than in our project.  It's not achievable at all in GTK, IMO.  Neither it applies to Lepton.  Please note, GTK is a much bigger project than Lepton.  For instance, even here, there are groups of people who are for rewriting everything in their favorite languages--C, Python, Scheme ;-)

I wouldn't like to be at the place of GTK devs now.  Please let's leave them alone.  I would need to have steel nervous to proceed today  if I would be one of them (not that I don't have such steel nervous, just maybe never noticed it :-)).

graahnul-grom:
Well, enough...
As if we have nothing else to discuss. :-)
Let's try to upload build results to coverity scan (please see my last message on gitter).

PR #817, opened 2021-06-26
Add basic attributes to symbols in sym/analog/

### LINKED ISSUE #784: Add basic attributes to basic symbols
Add `value` and `footprint` attributes to basic symbols.
There are basic symbols in `sym/analog/`; new user most likely start
their work with lepton by using them. The only attribute they have
(most of them) is `refdes`. What's the use for symbol which has
neither `value` nor `footprint` attribute? I propose to add these
attributes without bumping symversion.

by @graahnul-grom, moved here from #369

graahnul-grom:
- Add `value` and `footprint` attributes to the majority
of the basic symbols. Values are arbitrary and footprint
names are taken from the PCB (pcb.geda-project.org) library.

Remaining symbols I didn't touch:
1) darlington_NPN_PNP-1
2) inamp-1
3) jj-1
4) leddual-1
5) leddual-2
6) microphone-1
7) opamp-1
8) opamp-2
9) photo-resistor-1
10) photo-transistor-1
11) pwrxfmr-1
12) speaker-1
13) speaker-2
14) testpt-1
15) transformer-1
16) transformer-2
17) transformer-3
18) transformer-4
19) transformer-5
20) txo-1
21) CS5396-1 (what does it do in `analog/`?)
22) HA5221-1 (what does it do in `analog/`?)

- Copy symbols used in the `netlist` test suite to its test directory.
- Amend several title block symbols.

Closes #784.


vzh:
Looking at your changes I have to state two things we should eventually elaborate on, IMO:
- We need to specify somewhere in the Lepton manual what changes in symbols require their minor or major version change.
- We have to make one **integration** test suite instead of several ones for each tools and move netlist backend testing to it (I believe it should not depend on minor changes in the symbols in the repo).

I consider the question is important enough for a separate discussion on it, so any opinions matter and are encouraged.
WDYT, folks?

graahnul-grom:
> We need to specify somewhere in the Lepton manual what changes in symbols require their minor or major version change.

I think it should be
1) changes in graphics or attributes positions that lead to
changes in symbol's pins positions after translation
2) addition/removal of pins
3) changes in important attributes such as DEVICE
(not sure about other attrs)

graahnul-grom:
> We have to make one integration test suite instead of several ones

At least, we should make all test suites and example schematics
independent of symbols in the library (e.g. make a local copy
of symbols used).

vzh:
@graahnul-grom, do I understand correctly you've told this about **major** symbol version change?

I just wanted to say the changes made by you require at least bumping the minor version of the symbols.

graahnul-grom:
Yes, I think 1, 2, 3 is about major version changes.

graahnul-grom:
Many symbols in `sym/analog/` don't even have `symversion` attribs.

graahnul-grom:
I can't say yet what should constitute *minor* symbol changes. :)
What do you think?

vzh:
I think any changes in graphical stuff such as attribute placement, new lines in the body rectangle, colors etc. which do not affect the points you mentioned above, that is, the changes that do not break schematics with those changed symbols.

vzh:
I believe, if a symbol does not have the `symversion=` attribute, we should add it when changing.

graahnul-grom:
OK, I'll add missing `symversion` attrs (with what version? 0.1?)
and bump existing ones.

vzh:
Yeah, 0.1 would do.

vzh:
The change in the netlister test suite is really gross.  Have you simply copied all symbols to it, not only necessary ones?

graahnul-grom:
Just copied all of the symbols. :-)

vzh:
Just imagine how many debian lintian bugs we'll have... And how many different versions of same symbols we'll have in time.  Let's postpone this change, please.  At least not in this PR.

graahnul-grom:
I can run the [symcache](https://graahnul-grom.github.io/scm/symcache.scm.html) script and get rid of unused symbols
(why didn't I run it before?) On the other hand, if we postpone
making the netlist test suite independent of our symbols in the
library, the whole patch set becomes useless: no tests will pass.

vzh:
Eventually someone started to write proper scripts for Lepton :-) While it still has non-lisp DBCS (IMO), it's much more closer to what I would like to see in the project.  JFYI, using `let`s more then needed and side effects like `format`s make the code style imperative and non-functional, which means many things in it are wasted calculations.  Not that I ever write good code ;-)  I very appreciate such little scripts that show what can be done with Lepton.  I encourage you to publish the script at least on the geda-user list.  Please amend the branch using it.

The issue will be still present, anyways.  We have to somehow reduce the duplication of low-level stuff like symbols in the project, IMO.

graahnul-grom:
@vzh Removed commit where I add 3+ Mb of symbols. :-))
Using the *symcache* script, I've added only required
symbols to the netlist test suite.
I think we can tolerate a little overhead of having extra
copies of some symbols: IMO, it's much more important
to keep test suites and examples self-sufficient.

vzh:
Nice, thanks!  In future, I hope we'll do caches for test schematics automatically from the symbols in the repository, using something like your script.  I believe, it's easily achievable while someone still needs to spend their time on doing it.

vzh (approved):

graahnul-grom:
@vzh I publish my scripts here: https://graahnul-grom.github.io/#scm

PR #818, opened 2021-06-28
utils: Fix compilation warnings on launching Lepton Scheme tools.

vzh:
This was discussed recently on gitter.  The issue has been introduced in the last release 1.9.15.

The patch fixes the warnings as follows:
```
WARNING: compilation of /usr/local/bin/lepton-netlist failed:
no code for module (lepton ffi)
```

### DIFF
--- a/utils/archive/lepton-archive.scm
+++ b/utils/archive/lepton-archive.scm
@@ -61,8 +61,9 @@ exec @GUILE@ -s "$0" "$@"
 ;;;   Before each move, make sure that no overwrite of existing
 ;;;   files will occur.
 
-(unless (getenv "LIBLEPTON")
-  (add-to-load-path "@LEPTON_SCHEME_MODULE_DIRECTORY@"))
+(eval-when (expand load eval)
+  (unless (getenv "LIBLEPTON")
+    (add-to-load-path "@LEPTON_SCHEME_MODULE_DIRECTORY@")))
 
 (use-modules (ice-9 ftw)
              (ice-9 getopt-long)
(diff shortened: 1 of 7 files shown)


graahnul-grom (approved):
Thank you!
Could you please explain, why does it fix the warning?
How did you come to this solution?

vzh:
The short answer:  I just read Guile docs for years.

The long one:
Scheme as all Lisps is a very power language.  In the triangle `compile`--`eval`--`run`  it can do any sequence at any time, while lower level languages can do only some of them.  E.g. you can't run C code until you compile it.  The long story short, you really have to read Graham's observation reading about why the languages differ and why Lisps are most powerful among them.  Rant finished, and now just a couple of technical details I probably do not understand properly well myself :-).  I think Guile Scheme does several optimizations in diffrent circumstances.  When it compiles the code, it prefers to think you're rarely want to run macros to get some load paths, they are mostly predefined in Guile if you install code properly.  Anyways, it allows you to do anything.  You know, macros are evaluated before compilation, which makes them another, meta-programming level.  In Scheme, `eval-when` is such a macro, for which you can specify circumstances it should work in.  Some optimization code in Guile omitted the code that tried to add new directories to the Guile load path.  I just forced it to listen to me by defining when the code should be evaluated:  when it is compiled (expanded), when it is loaded, and when it is evaluated.  In short, always :-)

Well, maybe it's a bit an overhead here, though just for you and all reading this, it's just worth to see about Lisp: [William Byrd on "The Most Beautiful Program Ever Written](https://www.youtube.com/watch?v=OyfBQmvr2Hc)

graahnul-grom:
Thank you for the explanation

PR #822, opened 2021-07-05
Fix an error in the switcap netlist backend

### LINKED ISSUE #821: Switcap backend doesn't work completely, like it does in gEDA/gaf
Lepton's switcap backend doesn't work completely, like it does in gEDA/gaf by appending the test.ana file to the end of the netlist for it to be accepted as the proper input to the switcap program, sw.exe.

As stated in the .pdf document "gEDA/gaf Switcap Symbols and Netlister" by Dan McMahill, April 13th, 2003, found at http://web.mit.edu/~geda/arch/sun4x_59/versions/20060124/share/doc/geda-doc/switcap/switcap.pdf

"Figure 3 also shows an instantiation of an analysis block which directs the netlister to include the contents of the file test.ana in the output netlist"

The problem is that the command, lepton-netlist -g switcap -o example.scn ckt.sch clocks.sch analysis.sch
 generates the netlist file, example.scn, but only appends the following comment line,
 
/* reading analysis from "test.ana" */

to this example.scn netlist file, but FAILS to APPEND the contents of test.ana immediatly after this comment line, which is required to run the switcap simulation properly.

Please fix this backend problem with lepton.

Thank you,

Steven Astravas

graahnul-grom:
Closes #821.

### DIFF
--- a/utils/netlist/scheme/backend/gnet-switcap.scm
+++ b/utils/netlist/scheme/backend/gnet-switcap.scm
@@ -376,7 +376,7 @@
     (let ((fname (switcap:write-attrib package "file")))
       (display "\" */\n")
       (if (file-exists? fname)
-        (switcap:cat-file (open-input-file fname))
+        (switcap:cat-file (open-input-file fname) (current-output-port))
         (netlist-error 1 "ERROR: Analysis file ~S not found.\n" fname)))))
 
 (define switcap:cat-file


vzh (approved):
We definitely need a regression test for this.

graahnul-grom:
I've [tried](https://github.com/graahnul-grom/lepton-eda/commits/tb_355_switcap_fix_and_test) to add a test, but cannot make `make distcheck` work. Giving up.

PR #823, opened 2021-07-06
Fix legacy source library functions

vzh:
The functions are fixed in the following ways:
- The function `source-library-search()` no longer adds wrong directories with duplicated root name as reported by Karl Hammar on the geda-user mailing list.
- The functions `source-library()`, `source-library-search()`, `reset-source-library()` now return `%default-source-library` record.
- The functions check type of their arguments and raise `wrong-type-arg` error if they're wrong.
- Many new tests have been added for the functions.

PR #824, opened 2021-07-09
Fix environment variable expansion in source-library-search()

vzh:
While in #823 I fixed the function and added several tests, the test for env var expansion was incomplete: I only set a variable though never used it.
This commit set actually fixes expansion of environment variables in the function, adds tests for it, and includes some refactoring of the function.

Affects #823


vzh:
Merge this, since Karl Hammar confirmed this works on the geda-user list.

PR #825, opened 2021-07-13
Introduce new integration test suite (ITS) written in Scheme

vzh:
As a basis, I've taken the `test-driver.scm` from *autotools*.   I changed it a little to output results like our unit test suite (UTS) does.  Two schematics are now tested using the new test suite: `stack-torture.sch` and `stack_1.sch`. I intentionally omitted some tests yet (for output to stdout and verbose output).  They may be reintroduced some day, though it's an overhead at least for `stack-torture` tests which may be considered as a performance test.  After some fiddling with the code, I added the test for *switcap* example it its own directory as well.
Affects #822.


tests/netlist.test
```diff
@@ -0,0 +1,103 @@
...
+              "-g" backend
+              "-o" test-out
+              schematic-filenames)))
+
+    (when testdir (chdir cwd))
+
+    (test-eq EXIT_SUCCESS
+      (status:exit-val (system* "diff" test-out golden-out)))
+
+    (when (file-exists? test-out) (delete-file test-out))))
```
graahnul-grom, line 64: I think it would be useful for debugging test failures
to create temporary output files in the tests directory
(not in /tmp/) and retain them for failed tests.
Is it possible?

graahnul-grom (approved):

tests/netlist.test
vzh: Yes, sure, I thought of it.  I just had to start with something simple first as a proof of concept.

PR #827, opened 2021-07-21
Adapt three wiki documents under GFDL for Lepton manual

vzh:
- Adapted three GFDL'd documents for Lepton:
  - about interaction with PADS PowerPCB (`pads` backend and utilities)
  - about `cascade` backend
  - about `switcap` backend
- For this, several schematics and output files have been created/restored in their initial form.
- Added PNG and PDF images of the schematics/symbols for the PDF and HTML versions of the manual.
- Added/fixed links and missing info.
- Made some example schematics look better when exported to PDF or PNG.
- Suppressed building DVI docs on 'make distcheck'.  This was an unnecessary step which also required building EPS files that are larger than PDFs for same images and postprocessing them required two steps to make images not look blurred.
- Original wiki docs and images have been removed from the Lepton repository.

Affects #284


graahnul-grom:
After the build:
```
$ git status
?? docs/manual/lepton-manual.info-1
?? docs/manual/lepton-manual.info-2
```
And these files are installed into `share/info/`.
Where do they come from?

vzh:
@graahnul-grom, [texinfo splits large output info files](https://www.gnu.org/software/texinfo/manual/texinfo/html_node/_0040setfilename.html) and our output file became large enough to make this happen.  My fault.  I saw those files though ignored them for some reason and just deleted :-)

graahnul-grom (approved):

PR #830, opened 2021-09-27
Bump minimal GTK3 version to 3.2.0

vzh:
Several functions have been amended to support GTK 3.2.  Several warnings on deprecated have been fixed.


graahnul-grom:
When opening a file `lepton-attrib` crashes:

```
**
ERROR:gtksheet.c:3833:gtk_sheet_init: assertion failed: (FALSE)
Bail out! ERROR:gtksheet.c:3833:gtk_sheet_init: assertion failed: (FALSE)
Abort trap
```

A relevant part of `gtksheet` (4.3.5) code:
```
3827		has_css_parsing_errors = FALSE;
3828
3829		gtk_css_provider_load_from_resource (
3830		    provider, GTK_SHEET_CSS_RESOURCE_NAME);
3831
3832		if (has_css_parsing_errors) 
3833                    g_assert(FALSE);
```

vzh:
This is strange.  On my PCs I use at home and at work I never had such an error :-/
Anyway, we have to somehow debug and fix it. (Cannot promise I will do it soon or at all especially as I have hard times now (this is a poor translation from my thoughts in Russian, dunno if the expression I've chosen is valid in English).)

BTW, I found that `gtk_css_provider_load_from_resource()` is available since GTK 3.16.   The hint is: if we have such a dependency as a third party lib depending on that version, we have first to update our code for it :-) 

Thank you for the investigation!

vzh:
@graahnul-grom, could you please check what commit is the first one causing the issue (using bisect or so)?

graahnul-grom:
I'll try to bisect it this week and find out what's going on.

vzh:
Can I consider this as your approval? :-)

PR #840, opened 2021-10-29
Partial support of GTK3.4 functions

vzh:
Apart from deprecation warnings on color selection/chooser widget functions, most of other pre-3.4 function warnings have been eliminated as well as some warnings from lastly applied code.

### DIFF
--- a/libleptongui/src/x_event.c
+++ b/libleptongui/src/x_event.c
@@ -854,7 +854,15 @@ x_event_get_pointer_position (GschemToplevel *w_current, gboolean snapped, gint 
   width = gdk_window_get_width (window);
   height = gdk_window_get_height (window);
 
+#ifdef ENABLE_GTK3
+  GdkDisplay *display = gdk_display_get_default ();
+  GdkSeat *seat = gdk_display_get_default_seat (display);
+  GdkDevice *pointer = gdk_seat_get_pointer (seat);
+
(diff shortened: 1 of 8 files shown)


libleptongui/src/x_event.c
```diff
@@ -893,7 +901,15 @@ x_event_faked_motion (GschemPageView *view, GdkEventKey *event) {
   gboolean ret;
   GdkEventMotion *newevent;
 
+#ifdef ENABLE_GTK3
+  GdkDisplay *display = gdk_display_get_default ();
+  GdkSeat *seat = gdk_display_get_default_seat (display);
+  GdkDevice *pointer = gdk_seat_get_pointer (seat);
+
+  gdk_device_get_position (pointer, NULL, &x, &y);
```
graahnul-grom, line 909: Using ` gdk_device_get_position()` here has a strange effect:
please try to open add comp. dialog, select some symbol,
place cursor over the main window and do zoom in/out
with the mouse wheel.
Do you see the symbol jumping up and down when zooming, like I do?

graahnul-grom (approved):

libleptongui/src/x_event.c
vzh: Yes, it looks like it jumps a bit.  Maybe the issue with `GDK_SCROLL_SMOOTH` support I've already presented before.  I did some work on it before though cannot recall in which state I left it.  Need to re-check.

PR #841, opened 2021-11-01
Rewrite "lepton-cli config" in Scheme

vzh:
- `lepton-config` now does check if the specified directory exists
  and reports if it does not, returning non-zero exit status.

- Redundant arguments on the `lepton-cli config` invocation are
  no longer ignored.  Instead, the command prompts the user
  about what's happening and exits with non-zero exit status.

- The option `-p` with space before its argument is now processed
  the same way as without space.  Thus, several `XFAIL`ed tests
  have been fixed.

- The new utility now warns the user if the directories given on
  the command-line are missing.  Previously, it could output
  misleading messages about other errors, e.g., that the user may
  only specify one configuration store.

- An argument is now required by the options `-s` and `-u`, so
  using any combination of them in series without a file name
  argument leads to errors about wrong file names.


graahnul-grom:
I see that `-p` now *requires* an argument.
Is this intentional?

graahnul-grom:
On the other hand, `--project` does not.

vzh:
There are tests for the both options without arguments.  They do the same as you can see if you look at the tests.  Please look for  `'-p' without arguments` and `'--project' without arguments` in the `config.test` file.  The options are processed in the same block in the Scheme code so now they behave the same way.  (The set of tests may be different for historical reasons, though.) An only difference is *how* the user should specify the argument for the options.   `-p` can be now used in two forms, `-ppath` or `-p path`, while `--project` only as `--project=path`.  `-p=path` is now disallowed as an extension not conforming to POSIX and some other standards and due to its ambiguity (which directory name should be really used `path/` or `=path`?)  Me and you had a discussion on this before (#729). OTOH, the ambiguity of the expression `-p path` has been eliminated in the new code.  Indeed, is `path` a project directory or a group name here?  So I believe the change adds more consistency in the tool and the suite as a whole.

graahnul-grom:
All you have said about options is perfectly OK.
So, `-p` now must be used with an argument?

```
$ cat lepton.conf
[export]
dpi=144

$ lepton-config --project=. export dpi
144

$ lepton-config --project export dpi
144

$ lepton-config -p . export dpi
144

$ lepton-config -p export dpi
ERROR: Directory "export" does not exist.
```

I think, `--help` output and man page should be amended then.
Instead of
```
-p, --project[=PATH]  select project configuration
```
we should display
```
-p PATH, --project[=PATH]  select project configuration
```

graahnul-grom (approved):

vzh:
Yes, sure, thank you for the point!

PR #843, opened 2021-11-03
Fix two issues with gtk3

vzh:
- Add initial support for *smooth* scrolling and fix a warning about it.
- Fix issue with jumping of objects on zooming reported by @graahnul-grom in #840.


graahnul-grom (approved):
Excellent. Thank you!

PR #845, opened 2021-11-08
Allow attributes with empty values

### LINKED ISSUE #104: Empty attribute value should be valid
Right now, attributes with empty values are not allowed. But an empty string is a perfectly valid string, and can be useful for back ends driving downstream flows for which blank input is valid.

vzh:
The patch proposed by Karl Hammar on the **geda-user** mailing list.
Closes #104

### DIFF
--- a/docs/manual/attribute-net.texi
+++ b/docs/manual/attribute-net.texi
@@ -1,4 +1,4 @@
-@node net= attribute mini-HOWTO, Lepton EDA fonts, gEDA/gaf Master Attribute Document, Top
+@node net= attribute mini-HOWTO, Lepton EDA fonts, Master attribute list, Top
 @chapter net= attribute mini-HOWTO
 
 by: Ales Hvezda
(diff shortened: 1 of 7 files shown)


graahnul-grom:
If you really want to make such a drastic change, please
consider adding an option to control this behavior, e.g.:
`[schematic.attrib]::allow-empty-values`
(`false` by default, we can change it in the future).
Its description should explicitly warn that using empty
values will make schematics unusable in other versions
of Lepton and gEDA.
What do you think?

vzh:
What do you consider to be *drastic* in it?  In our tests, the change breaks only one test of a deprecated function.  I think it is an extension, not a limitation.  If the user wants his schematics to be supported in both gEDA and Lepton or in some previous versions of Lepton (which might be strange, IMO), he should not just use empty values of any attributes in his schematics.  We continue supporting schematics made in both gEDA and previous versions of Lepton. For me, it's similar to what we did when added support of some new attributes to `lepton-netlist` before.  If you see or could imagine any negative consequences of such a change, please let's discuss them here.  I'm reluctant to add new config settings just for such a small step forward. What **must** be really done is documenting of this change.

graahnul-grom:
Well, you're right. Maybe documenting the feature will be sufficient.

IMHO, adding an option for it allows one to be on the safe side:
those who want their schematics to be compatible might set it
to `false` to not accidentally add attributes with empty values.
Just a thought.

graahnul-grom (approved):

PR #852, opened 2021-12-02
Add and use accessors for window active page and its 'changed' flag in GUI

vzh:
This is a rather invasive change that allows getting rid of direct using of `LeptonToplevel` structure in many places of code and encapsulating it in several functions.  Some functions no longer use `toplevel` directly.  Several of them have been refactored as well. (I didn't fix `x_image_get_pixbuf()` yet as it requires more testing.)


graahnul-grom (approved):

vzh:
Thanks for prompt reviewing :-)

PR #854, opened 2021-12-05
Rewrite `lepton-cli export' in Scheme

vzh:
`lepton-export` has been rewritten in Scheme.  It's functions can be now, or are already, available in Scheme.  C functions used it the tool only, especially generic ones like related to compilation or lepton initialization, are no longer show-stoppers for further refactoring and improvement.


graahnul-grom (approved):
Well done!

vzh:
Thanks!

graahnul-grom:
I have a problem running lepton-netlist (latest master)
on schematics containing non-latin characters.
For example, in the following simple schematic,
lepton-netlist complains:
```
Throw to key `string-format' with args
`(string-page "Parse error: ~s"
("Parsing stopped at line 592:\n
Unexpected end-of-file in attribute list") ())'.
```
But if I remove one of the attributes with 'µA'
it runs without any errors.
@vzh Could you please try to run it on your system,
maybe something wrong on my side.

two

[two.sch.txt](https://github.com/lepton-eda/lepton-eda/files/7686802/two.sch.txt)

vzh:
@graahnul-grom, thank you for the report!  This is a definitely Scheme-vs-C string processing issue emerged in commit 485145f, which I found during bisecting.  I suspect it's about Scheme's `string-length()` vs C's `g_file_get_contents()` functions that count length of strings differently.  I suppose that the former one counts utf-8 characters while the latter seems to count just chars, i.e. bytes.  (Definitely, it's time to throw out the legacy geda format parser written in C.)  It's a regression IMO and we have to add a test for such a case, I mean using wider utf-8 symbols (non-just-ASCII) to our test suite.  Anyways.

PR #855, opened 2021-12-10
Fix regression in string->page()

vzh:
The code fixes a regression in Lepton discussed [here](https://github.com/lepton-eda/lepton-eda/pull/854#issuecomment-990014151).  The issue appeared in commit 485145f after rewriting `string->page()` in Scheme due to wrong assumption that any character in its `string` argument is one byte character.  It is not true for extended, non-ASCII UTF-8 set of characters.  Wrong string length calculation led to parsing errors.  This could be fixed with using string-to-bytevector functions from `(ice-9 iconv)`, e.g., with the code `(bytevector-length (string->bytevector str "utf8"))`.  However, as Scheme function `string->pointer()` converts strings to **NULL**-terminated pointers, it is enough to just notify `o_read_buffer()` about this by giving it the string length argument equal to `-1` and this solves the issue.

A test case with schematic provided by @graahnul-grom has been included in Lepton's ITS.


graahnul-grom (approved):
Thank you!
I almost lose sleep over this error before sharing the info with you.
Thinking that it's my fault in recent changes in my beloved BLPP
suite. I dismantled all of the new code and checked and debugged
several times new features' code, but cannot find any errors.
Thank you once again and certainly this evening I will drink to
your health. Hooray! :-)

vzh:
You're kidding as always :-)  Still looking forward to see your BLPP published somewhere.  Maybe I would be able to help somehow, and it could be land into the lepton suite in the end?  In any case, publishing it and attracting more attention to it would not harm for the project anyway, I believe.  Thanks for reviewing!

PR #856, opened 2021-12-10
Drop support for Guile 2.0

vzh:
Lepton's ITS already doesn't support Guile 2.0.  Since Guile 2.2 has been introduced a while ago and is widely adopted, it is time to reduce the burden on Lepton developers' shoulders and drop support for the previous stable version.  OTOH, it'll give us a chance of using more sophisticated modern Guile functions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ env:
 before_install:
   - sudo apt-get update
   - sudo apt-get install -y libgtk2.0-dev autopoint libgettextpo-dev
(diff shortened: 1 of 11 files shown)


bdale:
I fully support this change.  guile-2.0 is no longer even available in Debian stable, and my package builds for Debian use guile-3.0 these days.  So, guile-2.2 seems like a very reasonable "minimum version".

vzh:
OK, we all seem to have consensus on this.  I'll merge this now.  Please let me know if it breaks anything on your system(s).

PR #858, opened 2021-12-31
Improvements in color related code and a new GTK3 color chooser widget

vzh:
- The type `LeptonColor` has been amended so that, internally, it
  has the same structure as `GdkRGBA` which is used to represent a
  (possibly translucent) color in a way that is compatible with
  `cairo`'s notion of color.  More definitely, the field `enabled`
  of the `LeptonColor` type has been removed, and other fields
  that represent the color channels now have the type `double`.
  Instead of using the `enabled` field, disabling colors specified
  on Scheme side with the value `#f`, is now done in C by making
  them fully transparent, that is, by setting their alpha channel
  value to zero.

- A new color chooser widget for GTK 3 version of
  `lepton-schematic` has been introduced.  Apart from basic GTK
  functionality and a combobox for color selection, as in GTK2
  version, it has two buttons: *Apply* and *Back*.  The former is
  used to apply a selected color to immediately see the changes on
  the opened schematic page, and the latter is used to toggle
  between palette swatch view and color editor in the widget.
  Besides, it supports displaying translucent colors.

- Swatch rendering in the new color chooser and in some other
  widgets now also supports displaying non-opaque, translucent
  colors the same way the color chooser does, by showing such
  colors applied to a checkered background pattern.

- Functions, used for exporting a current color map for later
  including in Scheme RC scripts (typically, `gschemrc`), now
  support output of translucent colors by adding alpha channel
  values for colors that are not fully opaque.

- Color map creation code in export and GUI print functions has
  been unified.

- Provisions have been added against potential memory leaks in
  export functions in case they are called multiple times in
  Scheme scripts.

- Several build warnings have been eliminated due to the above
  changes.

- The minimal version of GTK3 has been bumped to 3.4.

**Palette swatch view**

palette-swatch-view

**Color editor view**

color-editor-view

### DIFF
--- a/libleptongui/src/color_edit_widget.c
+++ b/libleptongui/src/color_edit_widget.c
@@ -1,6 +1,6 @@
 /* Lepton EDA Schematic Capture
  * Copyright (C) 2018-2020 dmn <graahnul.grom@gmail.com>
- * Copyright (C) 2018-2021 Lepton EDA Contributors
+ * Copyright (C) 2018-2022 Lepton EDA Contributors
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -75,7 +75,7 @@ static void
 on_btn_apply (GtkWidget* btn, gpointer p);
 
 static void
-on_btn_back (GtkWidget* btn, gpointer p);
+on_btn_toggle (GtkWidget* btn, gpointer p);
 
 #else /* GTK2 */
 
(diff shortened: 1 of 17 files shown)


libleptongui/src/color_edit_widget.c
```diff
@@ -378,6 +493,52 @@ on_color_sel_changed (GtkColorSelection* csel, gpointer p)
...
+
+  g_return_if_fail (widget != NULL);
+  g_return_if_fail (widget->toplevel_ != NULL);
+
+  GtkColorChooser* chooser = GTK_COLOR_CHOOSER (widget->color_chooser);
+  GdkRGBA color;
+  gtk_color_chooser_get_rgba (chooser, &color);
+
+  int color_index = x_colorcb_get_index (GTK_WIDGET (widget->color_cb_));
+  x_color_set_display_color (color_index, &color);
```
graahnul-grom, line 514: Forgot to set outline color here?
`x_color_set_outline_color (color_index, &color);`

vzh: Thank you for the point, my oversight :-)

graahnul-grom:
What about renaming the "Back" button?
Something like "Toggle Editor"?

vzh:
I don't mind. Just tried to keep the button size minimized, thence the button text.

graahnul-grom (approved):

PR #859, opened 2021-12-31
Fix font output when exporting images from GTK3 GUI

vzh:
Previously, the code in question used some default font name and
this issue was only relevant for the GTK3 GUI.  Now, the font name
set up in configuration is used for exporting images.

### DIFF
--- a/libleptongui/src/x_image.c
+++ b/libleptongui/src/x_image.c
@@ -800,6 +800,14 @@ GdkPixbuf
                 "color-map", render_color_map,
                 NULL);
 
+  EdaConfig *cfg = eda_config_get_context_for_path (".");
+  gchar *fontstr = eda_config_get_string (cfg, "schematic.gui", "font", NULL);
+
+  if (fontstr != NULL) {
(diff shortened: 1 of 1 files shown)


graahnul-grom:
If the `schematic.gui::font` key is empty (`Sans` should be used
by default) exported image should look like this:

simple_gtk2

However, `File->Write Image...` produces image with  different font:

simple

If the `schematic.gui::font` key is set, export works as expected.
Could you please check it?

vzh:
No, on my system the output is identical with what I see on the canvas.  Anyways, I see your point.  The code used for GUI image exporting in the GTK3 port misses setting default values like default font name or junction size as can be seen in the images you've provided.  Other things broken I'm aware of is monochrome output when exporting from GUI.  Hopefully, it can be done in subsequent PRs.  And I would be glad if anybody files bugreports to not forget those issues.  I also hope this PR is a step forward and can be applied without solving yet any issues mentioned above ;-)

graahnul-grom (approved):

PR #864, opened 2022-01-08
Get rid of warnings on gtk_dialog_set_alternative_button_order() in GTK3 port

vzh:
AFAIK, there is no replacement function for it in GTK3.


graahnul-grom (approved):
Should we bump minimum required GTK3 version to 3.10?

vzh:
> Should we bump minimum required GTK3 version to 3.10?

I haven't checked yet what GTK version other functions that emit warnings on compilation depend on.  I have some more commits to review and publish.  Let's wait until I manage to submit them :-) Anyway, I believe we'll bump the minimum required version before the next release.

PR #870, opened 2022-02-13
Get rid of liblepton core modules

vzh:
This commit set makes our Scheme modules independent of `liblepton` Scheme-in-C interface.  All functions are now implemented as true FFI functions.  Some artifacts and dependence on Guile lib are still there.  In spite of it, we can now pre-compile our `lepton` modules dependent on `liblepton` without any special tricks, that is, by using just generic Guile compiler `guild`.


graahnul-grom (approved):
Great!

PR #880, opened 2022-03-16
Fix build compilation warnings when the USE_LDCONFIG variable is not defined on the ./configure stage

vzh:
This is a typical build case so the warnings are annoying.

### DIFF
--- a/liblepton/scheme/Makefile.am
+++ b/liblepton/scheme/Makefile.am
@@ -199,6 +199,7 @@ USE_GTK3=f
 endif !ENABLE_GTK3
 
 # Work-around for Debian and its derivatives.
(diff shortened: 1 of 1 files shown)


graahnul-grom (approved):
What kind of warning? I don't see any.
Is it Debian-specific?

vzh:
> What kind of warning? I don't see any.

E.g. [here](https://github.com/lepton-eda/lepton-eda/runs/5572248802?check_suite_focus=true) on the `make distcheck` stage, line 1309 and below, it reads several times:
```
Making all in scheme
/bin/bash: - : invalid option
...
```

> Is it Debian-specific?

Not sure.

PR #882, opened 2022-03-24
Finish porting lepton-attrib to the stable GTK version 3.24

vzh:
This is a last set of changes that eliminates several warnings on deprecated functions when compiling with GTK3 support.  I tested both GTK2 and GTK3 versions of the program and they seem to work OK.  There's a big room for improvements, though.  For example, I added  an only icon for the File->Save menu and declared all menu items translatable in the XML menu template (didn't test it though, and don't even know if/how it should work with *gettext*).


graahnul-grom (approved):
Why did you decide to place the 'about' and 'quit' menu items
in separate menu ('application')?

vzh:
I just started with a template from another program.

PR #888, opened 2022-04-02
The Scheme-in-C module (schematic core keymap) has been rewritten using Scheme FFI.

vzh:
This was the last Scheme-in-C module in Lepton.

The main innovation in this commit set is assigning a Scheme
callback on the C level for key processing in Scheme.  The current
solution is storing a necessary callback's pointer as a variable
in a variable in order it be defined once as some non-volatile
value.  Using a plain procedure in such a case could lead to
garbage collecting it at some point, which would lead to program
crashes.


graahnul-grom:
Calling `active-page()` in a function bound to a key crashes `lepton-schematic`:

```
$ cat gschemrc
( use-modules ( schematic window ) )

( define ( test_888 )
    ( active-page )
)

( global-set-key "F1" 'test_888 )
```

```
$ lepton-schematic
< and press F1 >

Backtrace:
           8 (primitive-load "/home/dmn/lepton/bin.tb/bin/lepton-schematic")
           7 (_)
In ice-9/boot-9.scm:
    152:2  6 (with-fluid* _ _ _)
In ice-9/eval.scm:
   293:34  5 (_ #(#(#(#(#<directory (schematic gui keymap) 803b7c8c0> #<pointer 0x805a640c0> #<pointer 0x805c71d90> #<pointer 0x800f4f500>)) #<pointer 0x805c71d90>) #<schematic-key F1>))
In ice-9/boot-9.scm:
    152:2  4 (with-fluid* _ _ _)
In ice-9/eval.scm:
    619:8  3 (_ #(#(#<directory (schematic action) 803b7c640> #<variable 8050f2f40 value: #<procedure no-last-action-error ()>> #<variable 8050f2f30 value: #<procedure eval-action!/recursive (a)>> test_888 #<variable 8050f2f50 value: #<procedure invalid-action-error ()>>) #<procedure test_888 ()>))
   293:34  2 (_ #(#(#<directory (schematic window) 803ac4f00>)))
    155:9  1 (_ #(#(#<directory (schematic window) 803ac4f00>)))
In unknown file:
           0 (_)

ERROR: Found invalid object (smob) #f
```

[out.log](https://github.com/lepton-eda/lepton-eda/files/8427370/out.log)

graahnul-grom:
It prevents crash, but still I cannot obtain and use an active page object.

```
Could not eval key "F1": misc-error (#f "Found invalid object (smob) ~S" (#f) #f)
```

I see `edascm_c_current_toplevel()` fail in `active-page()` function:
`EDASCM_ASSERT_SMOB_VALID` on line 77 in liblepton/src/scheme_toplevel.c.

Moreover, nothing is executed after the first call to `active-page()`
(if I add some code to `test_888()` after that).

vzh:
OK, I see, thanks.  I'll look into it further.

vzh:
@graahnul-grom, please test the most recent commit.  It seems to work on my system but I obviously cannot test it all thoroughly myself.

graahnul-grom:
Tested it. Now it works, thank you!
However, I see a strange behavior of the "Up" arrow key:
"Left", "Right" and "Down" do the usual thing - scroll the canvas,
but when I press "Up", current tab's header get the focus.
If I set `use-tabs=false` and press "Up", then the focus jumps to the toolbar.
Do you see this behavior, too?

graahnul-grom:
The focus is changed as described above even if the "Up"
key is rebound to some other action than `&view-pan-up()`.

vzh:
@graahnul-grom, thanks again! I seem to have found the issue. Hope to deal with it soon.

vzh:
Eventually, I had to stop playing with it :-)
In two words, the issue was in my superficial attention to the type of the key processing callback.  It should return a C boolean which is `FALSE` for an event if page view's parent widget has to process the event further.  In the previous version of the commit set I erroneously changed the type of the callback to `void`.  Moreover, I mistakenly interpreted the C expression `!scm_is_false(expr)` as `(not expr)` in Scheme while it is really just `expr` :-)
For example, if you press **Up**, the page view must process the key and stop the event processing since it is the part of the global keymap.  OTOH, processing of the **Tab** key which is not used in the keymap must be conveyed to the parent widget and lead to moving focus to the toolbar.
Hopefully, I've eventually managed things to behave properly.  Please test.

graahnul-grom (approved):
Now it works. Thank you.

PR #893, opened 2022-04-14
Support native compilation of Scheme files in the directory liblepton/ of the repository

vzh:


graahnul-grom:
It would be nice to have separate `make` targets
for compiling and installing `go` files, so that we
do not have to build/install them during development.
It would also assist packagers.

vzh:
Emm, probably I miss something in your point of view.  I'm not a
distribution packager, sorry.

From a user point of view, doing 
`./configure && make && sudo make install`
 is anything she would expect to make things working.  And
that's an established practice.  Adding any additional `make`
target is a little bit more inconvenient.

From a developer point of view, IMO, compiling `.go` files during
build is useful in that the developer can find issues that may be
not covered by Lepton's test suites.  Imagine a missing paren
somewhere in the not very often used part of a Scheme module in
the `schematic` namespace.  Or possible reporting of an unused
Scheme variable after refactoring.

Yeah, the compilation by `guild` increases the time of a build.
Please let me know of any other possible disadvantages in this
regard.

graahnul-grom:
It works - the tools uses precompiled `liblepton` Scheme files.

However, `make` (and `make distcheck`) failed with multiple
make jobs while compiling them on my system.
Did you try `make -jN` (`N > 1`)?

graahnul-grom:
Using only one `make` job, plus compiling Scheme files
*significantly* increases build time. IMHO, quite an expensive
way to catch potential errors undetected by the test suite.

While not introducing new `make` targets, we may think of
some `make` variable that allows skipping the generation
and installation of `go` files. WDYT?
e.g.:
```
$ QUICK_BUILD=yes make && make install
```

vzh:
> However, `make` (and `make distcheck`) failed with multiple make jobs while compiling them on my system. Did you try `make -jN` (`N > 1`)?

I always check PR branches with `make -j3 distcheck` first and have never hit any issues.  How do they exactly fail on your system? What error is output? (May it be related to the recent GH CI fails?)

graahnul-grom:
Attached the output of `make -j4`.

[BUILD.log](https://github.com/lepton-eda/lepton-eda/files/8534811/BUILD.log)

graahnul-grom:
Essentially, the error is "no code for module (lepton m4)".

graahnul-grom:
That error appears `N` times in the `make` output,
where `N` is the number of `make` jobs.

vzh:
@graahnul-grom, I've added two commits:
1) to solve your issue with `m4.scm`.  Please check if it works for you.
2) to solve the issue with long compilation.  After some thinking about it I still decided to opt your first proposal---do not compile `.go` files for now by default.  I've added the option `--enable-guild` to enable native compilation of Scheme files.  It can be changed to `--disable-guild` at any time in the future if we decide to build `.go` files by default.

graahnul-grom (approved):
Thank you!

PR #895, opened 2022-04-18
Get rid of warnings on deprecated functions in x_tabs.c when compiling using --with-gtk3

vzh:
Instead of using 'GtkAction's in the context menu of tabs, the
menu for GTK3 version of the widget has been reimplemented with
GSimpleAction.

gtk_action_*() functions have been deprecated since GTK 3.10.

### DIFF
--- a/libleptongui/src/x_tabs.c
+++ b/libleptongui/src/x_tabs.c
@@ -278,8 +278,19 @@ x_tabs_menu_create_item (GschemToplevel* toplevel,
                          const gchar*    icon_name);
 static void
 x_tabs_menu_create_item_separ (GtkWidget* menu);
+
+#ifdef ENABLE_GTK3
+static void
+x_tabs_menu_item_on_activate (GSimpleAction* action,
+                              GVariant *parameter,
+                              gpointer data);
+static void
+x_tabs_menu_item_on_activate_action (GtkMenuItem *item,
+                                     gpointer data);
+#else /* GTK2 */
 static void
 x_tabs_menu_item_on_activate (GtkAction* action, gpointer data);
+#endif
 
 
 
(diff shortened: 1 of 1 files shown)


vzh:
Hmm. GH docker CI system started segfaulting on random tests in our UTS.  Previously, I found this when rebased commits in the PR #893 onto the current master head.  Only removing parallel jobs for `make` by getting rid of `-j4` for `distcheck` revamped builds.  On my home PC anything works OK. What do we do, friends?

graahnul-grom:
Tested in VirtualBox (Ubuntu 20.04 - the same system as in CI)
both master and #893.
`make distcheck` runs fine regardless of the number of jobs.

vzh:
I found that `error 139` is a thing related to docker containers used by GH CI when running our jobs in Ubuntu.  They mean segfaults (`SIGSEGV`) for any reason.  One of the reasons I would come up with is some *out of memory* situation when running multiple jobs in parallel.  Just a shot in the dark.

graahnul-grom:
I've updated my local master branch and pushed it to my repo.
CI job completed [successfully](https://github.com/graahnul-grom/lepton-eda/runs/6076717161?check_suite_focus=true).

vzh:
You could try to push from this PR then? Or making an identical branch.

graahnul-grom:
Again, with this commit on top of master, CI [succeeded](https://github.com/graahnul-grom/lepton-eda/runs/6077008993?check_suite_focus=true).
Maybe that was some temporary failure on GH side?
Could you try to force push changes again to trigger CI run again?

vzh:
Again the same.  They seem to limit **my** jobs :-/

graahnul-grom:
What if you add a commit to this PR that reduces the number
of jobs in `.github/workflows/lepton-eda-ci.yml`? Worth a try, IMHO.

vzh:
It worked with `make -j2`.

BTW, in the last but one case I used `-j3` and instead of segfault with `error 139`, I saw some other error and the following message before:

> Inconsistency detected by ld.so: dl-close.c: 181: _dl_close_worker: Assertion `idx == nloaded' failed!

Looks like some racing condition with dlopening.  Maybe it's worth to try guile-3.0 in the gh container.  
(It's still enigma for me, though, why it breaks only with my branches, and starting at some point in time.)

vzh:
CI started to work for me as suddenly as it stopped before...

graahnul-grom:
Something was wrong with GH, most probably.

graahnul-grom (approved):
Didn't notice it before, but now I see that all of the icons in
tab's context menu are missing. But it's not because of this PR.
Something went wrong elsewhere.

vzh:
[This discussion](https://gitlab.gnome.org/GNOME/gtk/-/issues/1944) may shed some light on what's going on with menu icons in the gnome/gtk world.  There are some not deprecated means for supporting icons in menus, though they require more investigation and code.  My current effort is rather for getting rid of deprecation warnings than making UI prettier.

PR #896, opened 2022-04-21
Fix several minor issues in liblepton Scheme modules

vzh:
I've found the issues when using native compilation with guild as implemented in #893. They are related to eliminating of unused code / variables or, vice versa, exporting missing dependencies.


vzh:
The same gh/ci issue...

graahnul-grom (approved):

PR #905, opened 2022-05-17
Get rid of LeptonToplevel SMOBs

vzh:
Creating new **SMOB**s representing the `LeptonToplevel` structure
has been replaced with packing C objects to SCM values using FFI
functions as **SMOB**s are deprecated in Guile.  Tests in the
integration test suite have been altered accordingly.

The field `week_refs` has been removed from the struct
`st_toplevel` as no longer used.  Related accessor functions have
been removed as well.

The `LeptonToplevel` fluid used both in C and Scheme code is now
set in Scheme.  The name of the variable that provides it is
`%lepton-toplevel` and it is exported in the `(lepton toplevel)`
module.

Three procedures in the module `(lepton toplevel)`,
`%current-toplevel()`, `%make-toplevel()`, and `%with-toplevel`,
have been renamed.  The procedures had the prefix `%` as they were
previously defined in C code.  Now, when this is no longer the
case, the prefix has been dropped.

A new Scheme type, `<toplevel>`, has been added.  This is a
wrapped foreign pointer type in Scheme representing the
`LeptonToplevel` structure in C.  Three procedures in the module
`(lepton toplevel)` are now exported for transforming instances of the
type *to* or *from* pointers and for checking the type.  Those are
`toplevel->pointer`, `pointer->toplevel`, and `toplevel?`.

The C functions `edascm_with_toplevel()` and
`edascm_from_toplevel()` have been removed as no longer required.

The C function `edascm_current_toplevel()` is no longer available.
It used to set C toplevel fluid value in Scheme code.  Since the
toplevel fluid variable is now initialized in Scheme, the function
to convert its C sibling back to Scheme is superfluous.

The C function `edascm_c_current_toplevel()` is no longer exported
in Scheme FFI code.  The code of some Scheme modules as well as C
code of the tools `lepton-attrib` and `lepton-export` has been
modified so the tools do use an explicit value of the current
`LeptonToplevel` structure and the function is no longer required
for them to behave properly.


graahnul-grom:
When I bind a key to execute some function,
I've got an error on key press:

```
Could not eval key "F1": wrong-type-arg ("scm_i_foreign_call" "Wrong type argument in position ~A (expecting ~A): ~S" (1 "POINTER_P" #<toplevel-0x800f155c0>) (#<toplevel-0x800f155c0>))
```

It happens with some calls made in that function, e.g.
`active-pages()`, `active-pages()`, `make-page()` and,
probably, other functions, but not that many:

```scheme
; in gschemrc:
;
( define ( foo )
    ( format #t "foo()~%" )   ; OK
    ( active-page )           ; ERROR
    ( active-pages )          ; ERROR
    ( make-page "905.sch" )   ; ERROR
    ( log! 'warning "foo()" ) ; OK
    ( attribute? 'eklmn )     ; OK
    ( component? 'eklmn )     ; OK
    ( pointer-position )      ; OK
    ( make-box (cons 10 10) (cons 20 20) ) ; OK
)

( global-set-key "F1" 'foo ) ; could be any key or combination
```

If I make a menu item to execute the same function,
it works fine through the menu.
The same if I run it from the `:` prompt.

Tested on FreeBSD 13.1-STABLE with Guile 2.2.7
and Slackware Linux 15.0 with Guile 3.0.7.

vzh:
@graahnul-grom, thank you for the point.  Hopefully, this is fixed in the latest commit of this branch.

graahnul-grom (approved):
Thank you, it fixes the error.
Though I'm not sure I fully understand how;
and exact meaning of the last commit message either. :)

vzh:
Well, `%lepton-toplevel` is a Scheme fluid.  This means it can take any value and the program code can replace it with any value given under some circumstances defined by the programmer,  by me in this case.  In this case, I introduced a new Scheme type, `<toplevel>`, for wrapping C pointers of the type `LeptonToplevel`. While `%lepton-toplevel` was used to represent pure C pointers in the module `(lepton toplevel)`, it did otherwise in an only place, namely in the code changed in the last commit of this series.  There, I transformed the value it should be set to, to a newly defined structure, `<toplevel>`.  This was wrong because, well, from C point of view it is also a pointer, though a pointer to a SCM of a `<toplevel>` structure object, not a pointer baked in a Scheme SCM object representing a pure pointer the GUI code needed.  And the current code in the module `(lepton toplevel)`  requires *pure pointers* (so you can do `scm->pointer` and `pointer->scm` on them).  I am going to fix the code in future by making the `%lepton-toplevel` fluid not accessible for other code, that is, by not exporting it in the module, and amending functions working with `LeptonToplevel` so they would really work with its Scheme wrapper `<toplevel>`.  HTH.

Please don't hesitate to ask me more on this since I feel this explanation may be not fully intelligible :-)

graahnul-grom:
@vzh Thank you for the explanation!

PR #913, opened 2022-05-31
Gtk3 port: get rid of warnings on deprecated functions gtk_widget_override_color()

vzh:


libleptongui/src/gschem_bottom_widget.c
```diff
@@ -1218,22 +1178,32 @@ gschem_bottom_widget_set_status_text_color (GschemBottomWidget *widget, gboolean
...
 
 #ifdef ENABLE_GTK3
-  const GdkRGBA* color = NULL;
-#else
+  GtkCssProvider *provider = gtk_css_provider_new ();
+  GtkStyleContext *context = gtk_widget_get_style_context (widget->status_label);
+  if (active)
+  {
+    gtk_css_provider_load_from_data (provider,
+                                     "label { color: green; }", -1, NULL);
```
graahnul-grom, line 1186: Did you decide to postpone taking into account the status active color configuration option (schematic.status-bar::status-active-color)?

graahnul-grom:
Unfortunately, all attributes in the "Edit Attributes" dialog
are displayed with the same default black color: inherited
(should be gray) and individually selected (should be red).

libleptongui/src/gschem_bottom_widget.c
vzh: This was my oversight, sorry.  Hopefully it is fixed in the last commit here.

vzh:
On my system the Attributes Edit dialog looks like this.

Inherited attributes are displayed with the color of inactive elements:
attributes-editor-inherited

Selected attribs are not red like in GTK2 but dark-green (it may be dependent on the theme selected?):
attributes-editor-selected

libleptongui/src/gschem_bottom_widget.c
graahnul-grom: It works, thank you.

graahnul-grom:
Sorry, it turned out I have some GTK3 theme set in my config.
Now I see inherited attributes colored differently (when changing
to the default theme). But selected attributes still can't be
distinguished from normal ones on my system.

graahnul-grom (approved):

graahnul-grom:
All in all, the situation with GTK3 colors is strange.
With "disabled" color, if I trace the value of `color_insensitive`
in `multiattrib_update()` it is `rgb(146,149,149)`:



But in the Edit Attributes dialog it looks like violet.

And in the menu disabled items are dark gray:

vzh:
Just a shot into the dark: the issues may be due to using different GTK3's `StyleProvider` interfaces in various places of code.

PR #915, opened 2022-05-31
Gtk3 port: get rid of warnings on functions deprecated in Gtk 3.20

vzh:


graahnul-grom (approved):
With GTK3 and schematic.gui::warp-cursor = true
rotating the mouse wheel in both directions only
zooms *in*. Though it's not related to the changes in this PR.

vzh:
@graahnul-grom,

> With GTK3 and schematic.gui::warp-cursor = true
> rotating the mouse wheel in both directions only
> zooms in. Though it's not related to the changes in this PR.

I've bisected and found the issue was introduced in commit 52cdfc237c,
"gui[gtk3]: Initial support for smooth scrolling." :-)  Reverting the
commit would fix the issue.  However, I found this page:

https://gitlab.gnome.org/GNOME/gtk/-/issues/2048

It seems, direct working with `event->direction` is no longer possible
in Gtk3 due to two different event types.  A quote from there:

**Scroll events come in two flavors. Those with direction, and those
with deltas.**

I have to look into it a bit more...

PR #917, opened 2022-06-03
Get rid of SMOB machinery

vzh:
All deprecated gEDA *SMOB* machinery has been removed.  Now Scheme FFI code deals with C objects in Scheme.  Other related changes are described below.

- Three previously available SMOB types, `<geda-toplevel>`, `<geda-page>`, and `<geda-object>` have been replaced with three wrapped foreign pointer types, `<toplevel>`, `<page>`, and `<object>`, respectively.  In order to achieve this, the following changes have been done:
  - Three new modules with the definitions of the types and supporting functionality have been created: `(lepton toplevel foreign)`, `(lepton page foreign)`, and `(lepton object foreign)`
  - Other Scheme modules now use the above for transforming C pointers to/from Scheme pointers
  - Scheme functions dealing with them (`geda-something->pointer()` and such) have been renamed as well
  - All the functions in the above modules do now have due tests in the unit test suite
  - Functions in the `(lepton toplevel)` module are now tested in UTS as well
  - Function type check syntax rules in the above modules and their upper level modules have been unified, simplified, and moved to the above modules or `(lepton ffi)`
  - Functions dealing with hooks and the GUI check symbol feature code have been amended to support the new types

- Functions dealing with C `GList` lists in Scheme have been unified

- Locale is no longer set for Scheme code from C. All Lepton Scheme tools supporting localization already have the call to
`setlocale()` in their code.

- Lepton's C code has been amended as follows:
  - `f_open_flags()` has been renamed to `f_open()`, the previously defined `f_open()` just called `f_open_flags()`, so it has been removed
  - `s_toplevel_delete()` has been renamed to `lepton_toplevel_delete()`, it was an only so named function in `toplevel.c`
  - removed an unused argument in `x_rc_parse_gschem()`, the function is now called from Scheme
  - removed an unused callback `i_callback_file_new_window()`, its functionality was rewritten in Scheme a while ago
  - removed `edascm_c_current_toplevel()` as no longer used
  - `x_window_new()` no longer creates `LeptonToplevel` objects, nor it parses RC files

- `LeptonToplevel` objects are not created in C tests, it is no longer required as tested objects do not depend on the structure

- The module `(lepton toplevel)` now has a new procedure, `set-current-toplevel!()`, which may be used to unconditionally set the current toplevel fluid to a given value, which may be convenient for working with Lepton modules in REPL.

- Several new checks have been added for `make-page()`:
  - `make-page()` raises an error if current toplevel is not set
  - `make-page()` raises an error if if its filename argument is an empty string
  - added test for empty filename string in `make-page()`

- Added/fixed many docstrings and comments.

### DIFF
--- a/liblepton/scheme/lepton/object.scm
+++ b/liblepton/scheme/lepton/object.scm
@@ -1069,7 +1069,7 @@ where:
   y2 - Y-coordinate of bottom right of PICTURE.
   angle - rotation angle.
   mirrored - whether PICTURE object is mirrored."
-  (define pointer (check-object object 1 picture? 'picture))
+  (check-object object 1 picture? 'picture)
 
   (list (picture-filename object)
         (picture-top-left object)
@@ -1524,7 +1524,7 @@ component OBJECT."
   - Rotation angle.
   - Whether OBJECT is mirrored.
   - Whether OBJECT is locked."
-  (define pointer (check-object object 1 component? 'component))
+  (check-object object 1 component? 'component)
 
   (list (component-basename object)
         (component-position object)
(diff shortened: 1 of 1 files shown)


graahnul-grom (approved):
Отлично!

PR #914, opened 2022-05-31
Gtk3 port: get rid of warnings on deprecated functions gtk_widget_override_font()

vzh:

PR #922, opened 2022-06-15
Gtk3: Fix log and macro widgets fonts

vzh:
The bug was reported by @graahnul-grom in #914.

I'll repeat the link here: [gtk3 css issue](https://stackoverflow.com/questions/41257101/setting-gtkentry-font-from-pango-fontdescription).  A comment in the code says they broke gtk3 theming in gtk 3.22. Indeed, the fix found there (replacing *px* units with *pt*)  fixes font sizes in the widgets on my system.


graahnul-grom (approved):
Thank you, it fixes the font size problem.

graahnul-grom:
By the way, the *deliberately* break theming even in *micro* GTK3 releases.

PR #929, opened 2022-07-07
Improve slot editing action

vzh:
While playing with it in Scheme is fun, I still have to stop for
now :-) Further improvements may include adding appropriate hooks,
transforming the dialog into a widget that could be used both as a
dialog and as a widget in docks, etc.

So here is a summary of what has been done in the commit set so
far:

- The C function `o_slot_start()` has been thrown away.

- Previously, the dialog code used the function
  `o_select_return_first_object()` that returned the first
  selected object taking into account the order of objects in
  schematic file.  The action did nothing if the returned object
  was just a graphical primitive despite of the fact if the
  selection had components or not, it silently ignored user input
  in such situations.  The empty selection list was ignored as
  well.  Now, the list of objects is filtered and if it contains
  one component, it works with it.  As the C function is no longer
  necessary, the action does not use it any more.

- The contents of the function
  `o_attrib_search_object_attribs_by_name()` has been rewritten in
  Scheme and inlined in the code of the `&edit-slot()` action.
  Basically, it searches for an attribute of a selected component
  by name, traversing first its attached attribs and then its
  inherited ones.  It would be nice to eventually add a function,
  say `component-attrib(name)` that would *set in stone* such a
  canonical way of retrieving attributes from symbols because this
  is how they were obtained both in `lepton-schematic` GUI and
  in `lepton-netlist` for years.

- The slot editing action has been amended in that it now warns
  the user:

  - if no component objects to deal with is selected
  - if there are more than one component selected
  
  Lepton's generic message dialog is used to report such
  issues. Selection of non-component objects is admittable and
  does not affect the action.

- Slot edit dialog response callback is now assigned in Scheme.
  In time, it may be useful to replace it with a hook that could
  allow users to do their own actions when component slot has
  changed.  For example, swap components of a package when the
  slot of one of them has changed.

- The functions `slot_edit_dialog_response()` and
  `slot_edit_dialog()` have been splitted up into several
  functions to provide same functionality in Scheme.  No longer
  necessary C code has been removed.

- A separate Scheme module for Slot edit dialog code, `(schematic
  dialog edit-slot)`, has been added to localize the
  functionality.

- A wrapper function for the `g_signal_connect()` C macro has been
  added.  It is now available in Scheme and used in the new
  module.  I delayed its introduction for over two years as I was
  always in doubt if it is TRT so far.  Yet, other solutions seem
  to look much uglier for me, unfortunately.  The *response*
  signal of the Slot edit dialog is processed now by using this
  function.

- Improved/added doxygen comments of related C functions.

- The file `gschem_slot_edit_dialog.c` has been renamed to
  `slot_edit_dialog.c`.  The functions in it have not been renamed
  as I am still in doubt what prefix they should have :-) (I was
  inclining to add the `schematic_` prefix, not sure though.)

- As a side effect, a syntax macro returning pointers to foreign C
  callback functions has been added.  The macro just lazily calls
  for `dynamic-func()` and is not actually used as it was added as
  an intermediate step during the effort.  It may be useful in
  future, I hope (at least I have some not yet published stuff
  depending on it).

### DIFF
--- a/libleptongui/scheme/schematic/dialog/slot-edit.scm
+++ b/libleptongui/scheme/schematic/dialog/slot-edit.scm
@@ -25,31 +25,21 @@
   #:use-module (lepton ffi)
 
   #:use-module (schematic ffi)
-  #:use-module ((schematic selection) #:select (page-selection))
-  #:use-module (schematic window)
   #:use-module ((lepton object foreign) #:select (object->pointer))
 
   #:export (slot-edit-dialog))
 
 
-(define (slot-edit-dialog-response *widget response *window)
+(define (slot-edit-dialog-response *widget response *window component)
   (let ((accepted? (true? (slot_edit_dialog_response response))))
     (when accepted?
       (let ((*string (slot_edit_dialog_get_text *widget)))
         (unless (null-pointer? *string)
-          (with-window
-           *window
-           ;; The only one component must be selected.
-           (let ((component (car (page-selection (active-page))))
-                 (slot-string (string-append "slot=" (pointer->string *string))))
-             (o_slot_end *window
-                         (object->pointer component)
-                         (string->pointer slot-string)))))))
+          (let ((slot-string (string-append "slot=" (pointer->string *string))))
+            (o_slot_end *window
+                        (object->pointer component)
+                        (string->pointer slot-string))))))
     (slot_edit_dialog_quit *window)))
-
-
-(define *slot-edit-dialog-response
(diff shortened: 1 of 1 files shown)


graahnul-grom:
This is a half-finished work, as I see it.

graahnul-grom (approved):

vzh:
dmn ***@***.***> writes:

> This is a half-finished work, as I see it.

Yeah, there's a wide field for improvements.

What things would you like to be improved/fixed?

PR #934, opened 2022-07-21
Create lepton-schematic's toolbar widget in Scheme

vzh:
- A new function, `schematic_window_create_toolbar()` has been
  added.  It returns the created toolbar widget.
- C setters and getters for properties of the buttons have been added.
- A new module, `(schematic toolbar)`, now serves the toolbar
  creation.  Relevant functions have been moved to it.  The code
  has been refactored and splitted up so that the toolbar buttons
  and separators, groups of them and their properties, are now
  created and set in Scheme using separate procedures.
- A dedicated C file for toolbar functions has been introduced.
- **Glib** signals are now connected to toolbar buttons in Scheme.
- Three `GschemToplevel`'s fields storing radio button widgets
  have been eliminated.  The code has been amended to get widgets
  from the toolbar widget itself.
- Several functions have been renamed.
- Added/improved docstrings and comments for several functions.
- As an artifact of the code described above, a new function,
 `reference-pointer()` has been added to the `(lepton ffi)`
 module.  Its name reflects the fact that the procedure does the
 reverse action compared with the stock Guile procedure
 `dereference-pointer()`.  Its location in the `lepton` namespace
 will allow for dealing in Scheme with other C code where pointers
 to pointers are used, e.g. variables related to `g_error_*()`
 functions.


graahnul-grom:
The Scheme code tries to create toolbar even
if it's turned off in settings (`schematic.gui::toolbars=false`): there are warnings about failed calls to `gtk_toolbar_insert()`
(toolbar == NULL) in terminal.
It doesn't check the return value of `schematic_toolbar_new()` in `make-toolbar()`.

vzh:
@graahnul-grom, thank you!  I've added a couple of commits fixing the issue.

graahnul-grom (approved):

graahnul-grom:
I would also consider moving the "add text" button
to the left, so that three radio buttons are grouped
together (add net, add bus, select mode). WDYT?

vzh:
> I would also consider moving the "add text" button to the left, so that three radio buttons are grouped together (add net, add bus, select mode). WDYT?

Go for it.  Now you can test the result even without recompilation.  Just edit the installed Scheme file and go :-)

Issue #446, opened 2019-07-15
Homebrew on OSX: error: GIO 2.38.0 or later is required


craignied:
Compiling with Homebrew on OSX:

checking for gio-2.0 >= 2.38.0... no
configure: error: GIO 2.38.0 or later is required.

GIO isn't a module that I can find on Homebrew. Is it part of something else?

Thanks in advance

graahnul-grom:
@craignied Thank you for reporting this. `GIO` is a part of `glib`. We recently
bumped minimum required version of `glib` to 2.38.0. Do you have the `glib`
package installed?

craignied:
I have glib 2.60.5 installed:

brew info glib
glib: stable 2.60.5 (bottled)
Core application library for C
https://developer.gnome.org/glib/
/usr/local/Cellar/glib/2.60.5 (429 files, 15.2MB) *
  Poured from bottle on 2019-07-13 at 20:19:37
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/glib.rb
==> Dependencies
Build: meson ✘, ninja ✘, pkg-config ✔
Required: gettext ✔, libffi ✔, pcre ✔, python ✔
==> Analytics
install: 280,229 (30 days), 837,113 (90 days), 2,289,367 (365 days)
install_on_request: 15,309 (30 days), 42,956 (90 days), 105,085 (365 days)
build_error: 0 (30 days)

graahnul-grom:
Could you please try the following command:
```
 $ pkg-config --libs gio-2.0
```

craignied:
Still getting

checking for gio-2.0 >= 2.38.0... no
configure: error: GIO 2.38.0 or later is required.

graahnul-grom:
What is exactly the `./configure` command line you are using?
It would be helpful if you attach the full output of `./configure`
and the generated `config.log` file.
You could also try to disable the `GIO` check (just comment out two lines in `configure.ac`):
```diff
diff --git a/configure.ac b/configure.ac
index 9923799ea..40fd51ed1 100644
--- a/configure.ac
+++ b/configure.ac
@@ -79,8 +79,8 @@ PKG_CHECK_MODULES(GLIB, [glib-2.0 >= 2.38.0], ,
   AC_MSG_ERROR([GLib 2.38.0 or later is required.]))
 AC_DEFINE([G_DISABLE_DEPRECATED], [1], [Disable deprecated GLib features])
 
[...]

Issue #76, opened 2017-02-28
Segfault on OS X


noqsi:
While lepton compiles on OS X (version 10.9.5), it does not function properly

When I first start `gschem`, it does not display any of the widgets (see the screen capture below).  If I mouse over them, the widgets slowly reveal one by one.  This is also true of the menu widgets.  The scroll for the drawing area is also out of place.

After a short amount of use, I get a segmentation fault with the following logged errors:

```
2017-02-27 20:11:02.111 gschem[96694:507] *** WARNING: Method userSpaceScaleFactor in class NSView is deprecated on 10.7 and later. It should not be used in new applications. Use convertRectToBacking: instead. 
Invalid path [/usr/local/share/gEDA/sym/local] passed to component-library
Feb 27 20:12:11 Kakuzen.local gschem[96694] <Error>: CGContextRestoreGState: invalid context 0x7f8c33c99110. This is a serious error. This application, or a library it uses, is using an invalid context  and is thereby contributing to an overall degradation of system stability and reliability. This notice is a courtesy: please fix this problem. It will become a fatal error in an upcoming update.
Feb 27 20:12:11 Kakuzen.local gschem[96694] <Error>: CGContextSetAllowsAntialiasing: invalid context 0x7f8c33c99110. This is a serious error. This application, or a library it uses, is using an invalid context  and is thereby contributing to an overall degradation of system stability and reliability. This notice is a courtesy: please fix this problem. It will become a fatal error in an upcoming update.
```

![image](https://cloud.githubusercontent.com/assets/404801/23390169/a9880410-fd29-11e6-92a5-4b002bc33671.png)

vzh:
@noqsi After some googling I found [this](https://github.com/trizen/youtube-viewer/issues/115). Two links there point to gtk3 code and to documentation for class NSScreen used in apple products (and in quartz among others). One of the gtk2 artifacts in the new gtk3 code is still the method `userSpaceScaleFactor` reported on your console. It is deprecated in OS X 10.7 and later, as the error says. As far as I can see on the first page, using link to differences between gtk 3.22 and 3.89, nothing has changed even in the latter version. We even don't support gtk3, so the situation with first warning seems to be hopeless :-( I don't know if the two errors after it are someway bound to it, though.

gareth8118:
@noqsi couple of questions:

- did gEDA/gaf run on OS X for you? 
[...]

PR #939, opened 2022-08-06
Rewrite some Page select widget's callbacks in Scheme

vzh:
- C functions for translating action mode between C and Scheme have been added.
- Accessors for window's Page select widgets have been added.
- Several C callbacks for the widget have been rewritten in Scheme.
- A new module, `(schematic callback)`, has been added to allow for calling the callbacks from both Scheme and C.

### DIFF
--- a/libleptongui/include/gschem_toplevel.h
+++ b/libleptongui/include/gschem_toplevel.h
@@ -338,4 +338,10 @@ schematic_window_set_actionfeedback_mode (GschemToplevel *w_current,
 GList*
 schematic_window_get_place_list (GschemToplevel *w_current);
 
+GtkWidget*
+schematic_window_get_page_select_widget (GschemToplevel *w_current);
+
+void
+schematic_window_set_page_select_widget (GschemToplevel *w_current,
+                                         GtkWidget* widget);
 G_END_DECLS
--- a/libleptongui/include/page_select_widget.h
+++ b/libleptongui/include/page_select_widget.h
@@ -51,7 +51,8 @@ G_BEGIN_DECLS
 GtkWidget*
 page_select_widget_new (GschemToplevel* w_current,
                         GCallback page_new_callback,
-                        GCallback page_open_callback);
+                        GCallback page_open_callback,
+                        GCallback page_save_callback);
 void
 page_select_widget_update (GschemToplevel* w_current);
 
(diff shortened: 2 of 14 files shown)


graahnul-grom (approved):
Frankly, this doesn't look to me as an improvement or simplification.
Or is it just an intermediate step?
Anyway, why do you treat `i_callback_file_save()` differently
compared to other callbacks?

vzh:
> Frankly, this doesn't look to me as an improvement or simplification. Or is it just an intermediate step?

Yes, this is an intermediate code.  The road is long and its ultimate goal is to make `liblepton` and `libleptongui` independent of `libguile`.  Moving callbacks to Scheme and introducing accessors for C objects is just a part of the job.

> Anyway, why do you treat `i_callback_file_save()` differently compared to other callbacks?

Well, because it is used in many places of code compared to other callbacks.

graahnul-grom:
> make liblepton and libleptongui independent of libguile

+1

graahnul-grom:
> make liblepton and libleptongui independent of libguile

Though reducing the number of dependencies is always desirable,
in case of `libguile` it's not immediately obvious what
technical benefits it could bring to the users (and us).
It would be nice if we have this described in our issue tracker.

vzh:
JFTR, @dmcmahill raised several issues on gitter he stumbled upon while trying to make lepton work on Mac.  Unfortunately, I myself cannot do much to fix them.  While guile itself works on Mac, lepton, with its interlaced Scheme->C->Scheme-in-C code seems to introduce hard issues with dynamic libraries. Installation with non-default `--prefix` doesn't work and requires setting of the env var `GUILE_EXTENSIONS_PATH`.  Even so, it doesn't work. BTW, the issue #76 is one of the most visited, AFAICS.
Another point, using FFI is a recommended way to extend programs nowadays. Working with Guile garbage collection in C is worse than inconvenient and may introduce serious bugs, and that struck me many times. Is it enough for now? :-)

graahnul-grom:
Thank you. I understand that working with Guile in C is hard.
Unfortunately, I can't help with Mac stuff, too. I have no such hardware.
But there is hope, I believe, it seems that  geda-gaf works on Mac
(macport: https://github.com/macports/macports-ports/tree/master/science/geda-gaf), as I mentioned in #76, so lepton-eda probably
should work, too. But nobody tried it under macports so far
(in #76 and #446 it was `homebrew`).

vzh:
Thank you for the link. Maybe patches from there may solve at least some issues with porting lepton to Mac.

PR #943, opened 2022-08-22
Rewrite undo callbacks in Scheme

vzh:
- Two C callbacks, `i_callback_edit_redo()` and `i_callback_edit_undo()`, have been rewritten in Scheme.  As a result, two new Scheme functions, `undo!()` and `redo!()` have been introduced.  They are exported in the module `(schematic undo)`.
- Both callbacks do now check the config setting `undo-control` on the fly for current directory in order to do or skip the job.
- The fluid `%lepton-window` and functions related to it have been moved to a new module, `(schematic window global)`, in order to avoid circular dependencies between modules in the `schematic` namespace.
- Two global undo related variables have been eliminated.
- In order to facilitate the above, the `*current-window()` macro has been moved to `(schematic window global)`.  The macro is used to simplify the new undo procedures.

### DIFF
--- a/libleptongui/include/gschem_defines.h
+++ b/libleptongui/include/gschem_defines.h
@@ -70,10 +70,6 @@
  * constants, so be sure not to clash with those */
 #define LAST_DRAWB_MODE_NONE -1
 
-/* used in o_undo_callback */
-#define UNDO_ACTION             0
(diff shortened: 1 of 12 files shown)


graahnul-grom (approved):
sorry for the delay

PR #949, opened 2022-09-13
Get rid of the C function o_update_component().

vzh:
- The function `o_update_component()` has been rewritten in
  Scheme.  In a way, it updated components by using some magic
  when dealing with their primitives in C.  Hopefully, things got
  a little bit more clear after rewriting.

- A missing docstring for `make-component/library()` has been
  added.


graahnul-grom (approved):

libleptongui/scheme/schematic/builtins.scm
```diff
@@ -428,7 +428,89 @@ the snap grid size should be set to 100")))
...
+          (attribs (object-attribs current-component)))
+
+      ;; Force symbol data to be reloaded from source
+      (s_clib_symbol_invalidate_data
+       (s_clib_get_symbol_by_name (string->pointer basename)))
+
+      ;; Create new object.
+      (let ((new-component
+             (set-component-with-transform!
+              (make-component/library basename
```
graahnul-grom, line 456: `make-component/library()` could fail at this point, and
`set-component-with-transform!()` will throw an error:

```
Could not eval key "U": wrong-type-arg (#f "Wrong type argument in position ~A (expecting ~A): ~A" (1 component #f) #f)
```

It could happen if one opens another schematic from the directory
with `gafrc` calling `reset-component-library()` and setting up
different component library paths. Then there's a chance that components
user wants to update in previously opened schematic(s) are no longer available.

vzh: I managed to reproduce the issue, thank you!
Hopefully the last commit solves the issue.

graahnul-grom: Yes, thank you.

PR #951, opened 2022-09-18
Rewrite i_callback_attributes_*() functions in Scheme

vzh:
- A new Scheme procedure has been introduced in the module
  `(lepton object)`: `set-text-attribute-mode!()`.  The procedure
  sets a desired mode for an object and emits change
  notifications.

- A few C functions, in particular `i_callback_attributes_*()`
  ones have been rewritten in Scheme.
  
- A couple of no longer needed C functions have been thrown away.

### DIFF
--- a/docs/scheme-api/lepton-scheme.texi
+++ b/docs/scheme-api/lepton-scheme.texi
@@ -1378,6 +1378,13 @@ will be one of the following symbols:
 @end itemize
 @end defun
 
+@defun set-text-attribute-mode! text
+Sets the attribute mode of @var{text} to @var{mode} which should be
+one of the symbols @samp{'name}, @samp{'value}, or @samp{'both}.
+Returns the modified text object.
+@end defun
+
+
 @node Components
 @subsection Components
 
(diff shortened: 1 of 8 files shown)


graahnul-grom:
Both deleted functions, `o_attrib_toggle_visibility()` and
`o_attrib_toggle_show_name_value()`, used to call
`schematic_window_active_page_changed()`.
But now there is no visual indication of the "page changed"
status after "show name", "show value", "show both" and
"toggle visibility" actions.

graahnul-grom:
I hope this patch could fix it:

[0001-gui-Page-changed-on-attribute-modification.diff.txt](https://github.com/lepton-eda/lepton-eda/files/9608765/0001-gui-Page-changed-on-attribute-modification.diff.txt)

[0001-gui-Page-changed-on-attribute-modification.patch.txt](https://github.com/lepton-eda/lepton-eda/files/9608758/0001-gui-Page-changed-on-attribute-modification.patch.txt)

vzh:
@graahnul-grom, thank you for your review and patch!  However, I had a bit better solution just before I found yours :-) Please test.

PR #955, opened 2022-09-23
Get rid of i_callback_hierarchy_*() functions

vzh:
Three C callbacks, `i_callback_hierarchy_up()`,
`i_callback_hierarchy_down_symbol()`, and
`i_callback_hierarchy_down_schematic()` along with some other
related functions used in them, have been rewritten in Scheme and
removed.

To make it happen, the following changes have been introduced:
- A generic error dialog code has been factored out.
- Accessors for several fields of `LeptonPage` as well as for the
  static variable `page_control_counter` have been added and
  utilized in the code.
- The flags defined in C, `HIERARCHY_FORCE_LOAD` and
  `HIERARCHY_NORMAL_LOAD`, have been thrown out.

My first attempt to rewrite `i_callback_hierarchy_down_schematic()`
revealed some issues, and I decided to rewrite it step by step
starting from some Scheme code that almost wholly repeats its C
version, and then steadily refactor the function.

- Apart from simplifying the code, the new Scheme version has
  fixed not so obvious bug that was present in the original C
  callback.  The code wrongly incremented an attribute counter
  variable for `source=` attributes even when processing of
  attached attribs was done and it should be reset to zero.
  Instead of resetting the counter to start with the first
  inherited attribute, it was incremented again.  In case when
  there were attached `source=` attributes though the files
  mentioned in them were missing or unreadable, the code might try
  to deal with, say, the third inherited `source=` attribute
  instead of the first one.  This might be, for example, if there
  were two attached `source=` attributes.  In a similar case, when
  the number of inherited `source=` attributes was less than the
  number of attached ones, the user might think the program does
  the right thing by rejecting inherited attributes if there were
  errors in dealing with attached ones.  However, it was only a
  seeming behaviour.  The code has been amended to behave like
  `lepton-netlist` does in such cases: if there are attached
  `source=` attributes, only they are processed; if the list of
  attached `source=` attributes is empty, the inherited attributes
  are processed.

- Now the user is warned if no component is selected.  Previously,
  if no component were selected, or if the first selected object
  was not a component, the action silently returned without any
  visual effect.  The following improvements have been made:
  - Page selection is filtered to ignore non-component objects
    before processing the first selected component.
  - If no component is selected, the action warns the user by
    displaying an error dialog with a message about this.

One more bug have been fixed in the `&hierarchy-down-symbol()` action:

- Previously, if a symbol name was relative, searching the name
  amongst the names of active pages that had been "normalized",
  that is, made absolute in C code before creating the pages,
  would fail.  The issue has been fixed by "canonicalizing" symbol
  names in Scheme before searching for them.


graahnul-grom:
While I'm still testing the changes in this PR, it's worth noting
that while hierarchy->down_schematic displays error dialog
when no component is selected, hierarchy->down_symbol
does not. Also there is no error dialog if several components
are selected.

vzh:
OK, I've modified both actions to warn the user if no or several components were selected.  Please test.

graahnul-grom (approved):

libleptongui/src/x_dialog.c
```diff
@@ -86,6 +86,33 @@ int generic_confirm_dialog (const char *msg)
...
+ */
+void
+generic_error_dialog (const char *primary_message,
+                      const char *secondary_message)
+{
+  GtkWidget *dialog = gtk_message_dialog_new (NULL,
+                                              GTK_DIALOG_MODAL,
+                                              GTK_MESSAGE_ERROR,
+                                              GTK_BUTTONS_OK,
+                                              primary_message);
```
graahnul-grom, line 107: Here I get a compiler warning:
```
x_dialog.c:107: warning:
format string is not a string literal
(potentially insecure) [-Wformat-security]
```
I managed to fix it this way:
```diff
diff --git a/libleptongui/src/x_dialog.c b/libleptongui/src/x_dialog.c
index 8d3106dd8..672e102e3 100644
--- a/libleptongui/src/x_dialog.c
+++ b/libleptongui/src/x_dialog.c
@@ -104,7 +104,8 @@ generic_error_dialog (const char *primary_message,
                                               GTK_DIALOG_MODAL,
                                               GTK_MESSAGE_ERROR,
                                               GTK_BUTTONS_OK,
-                                              primary_message);
+                                              "%s",
+                                              primary_message ? primary_message : "");
   g_object_set (G_OBJECT (dialog),
                 "secondary-text", secondary_message,
                 NULL);
```

graahnul-grom:
If we move hierarchy-related error messages to dialogs,
shouldn't other use cases be considered, too?
- `down hierarchy` for selected symbol that has no sub-schematics
associated with it
- `hierarchy up` if there's no "upper" page for the current schematic
- `down symbol` if symbol cannot be found in the component library
Maybe I forget something else?
Though I think these error dialogs could be added later.

libleptongui/src/x_dialog.c
vzh: While I don't see such a warning when compiled with *gcc* or *g++*, here is my fix which seems to be more straightforward:
[0001-gui-Fix-the-warning-format-string-is-not-a-string-li.patch.txt](https://github.com/lepton-eda/lepton-eda/files/9690542/0001-gui-Fix-the-warning-format-string-is-not-a-string-li.patch.txt)

If you like it, I'll push it, otherwise, please commit your solution yourself, it also works OK for me :-)

vzh:
I considered also messages in status bar and even wrote some code for it though unfortunately it didn't look very nice for me, so I stopped this yet.  If you want to improve any of the actions, please go for it, or make an issue with feature request.

graahnul-grom:
Sure, dialogs or something else (status bar messages or whatever)
could be sorted out later.

libleptongui/scheme/schematic/builtins.scm
```diff
@@ -861,14 +862,209 @@ the snap grid size should be set to 100")))
...
+                          (string->pointer secondary-message))
+
+    (g_clear_error *error)))
+
+(define (hierarchy-down-filename filename *window *parent page-control)
+  (define use-tabs? (true? (x_tabs_enabled)))
+
+  (log! 'message (G_ "Searching for source ~S") filename)
+  (let* ((*error (bytevector->pointer (make-bytevector (sizeof '*) 0)))
+         (*child (s_hierarchy_down_schematic_single *window
```
graahnul-grom, line 907: If `s_hierarchy_down_schematic_single()` returns NULL here,
you eventually get a "dereference null pointer" exception.
`hierarchy-down-error-dialog()` -> `gerror-message()`:
because bytevector(0) is not a %null-pointer.

vzh:
@graahnul-grom, please look at the last two commits in the series.  I tried to solve the issues mentioned by you before: a warning by clang (IIUC), and the potential NULL pointer dereferencing.  I could not trigger the latter, though.

graahnul-grom:
@vzh Thank you for fixing it.

PR #962, opened 2022-10-18
Rewrite x_window_open_page() and some related functions in Scheme.

vzh:
- The recent chooser widget code has been amended as follows:
  - The callback for its signal `item-activated` is now assigned
    in Scheme.  It has been moved to the module `(schematic
    window)`
    for further code refactoring.
  - The maximal number of recent files which are displayed in the
     corresponding menu, is now processed in Scheme.  Unlike it
     was previously, if a `'config-error` error is raised while
     reading the config key `max-recent-files` the error is now
     reported to the log and is thus available in the log window
     of `lepton-schematic`.
  - Opening of a selected page is now handled in Scheme as well.

- A module for *File select* dialog functions has been created.
  It is called `(schematic dialog file-select)` and currently contains
  an only function, `file-select-dialog()`.

- The procedure `make-schematic-window()` now returns a `<window>`
  object representing the window created.

- A new procedure, `window-set-current-page!()`, has been
  introduced.  It replaces foreign calls to
  `x_window_set_current_page()` in Scheme code and works with
  `<window>` and `<page>` objects.

- A new Scheme function, `schematic-error-dialog()`, has been
  introduced to represent a generic error dialog window.  The C
  foreign function that launches the generic error dialog has been
  amended to allow for setting the title of the dialog.

- A new procedure, `window-open-page!()`, has been introduced.
  The calls to the foreign function `x_window_open_page()` have
  been replaced by calls to the Scheme procedure.  Finally, the
  function `x_window_open_page()` has been rewritten in Scheme and
  the function `x_window_open_page_impl()` has been renamed to
  `x_window_open_page()`.

### DIFF
--- a/libleptongui/include/prototype.h
+++ b/libleptongui/include/prototype.h
@@ -416,7 +416,8 @@ int generic_confirm_dialog(const char *);
 
 void
 generic_error_dialog (const char *primary_message,
-                      const char *secondary_message);
+                      const char *secondary_message,
+                      const char *title);
 
 char * generic_filesel_dialog(const char *, const char *, gint);
 
--- a/libleptongui/include/prototype.h
+++ b/libleptongui/include/prototype.h
@@ -535,7 +535,9 @@ lepton_menu_set_action_data (GtkWidget *menu,
                              gpointer action);
 void
 x_menu_attach_recent_files_submenu (GschemToplevel* w_current,
-                                    GtkWidget*      menuitem);
+                                    GtkWidget* menuitem,
+                                    gint max_items);
+
 /* x_multiattrib.c */
 void x_multiattrib_open (GschemToplevel *w_current);
 void x_multiattrib_close (GschemToplevel *w_current);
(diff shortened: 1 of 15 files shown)


graahnul-grom (approved):
Previously, the value of `max-recent-files`
has been checked to be `> 0`.
Now, if you set it to a negative value,
this warning is issued:

```
GLib-GObject-WARNING:
value "-123" of type 'gint' is invalid or out
of range for property 'limit' of type 'gint'
```

But it could be set set to `-1` and considered
to be a valid number: `gtk_recent_chooser_set_limit()`
accepts `-1` as a value for its `limit` parameter;
in this case the recent files menu would contain
*all* recent files.

Probably, it's better to limit possible values
to positive integers.

vzh:
Thank you for the point. I've amended the code to catch the case with non-positive integers and to not output a superfluous newline to log.

graahnul-grom:
@vzh Thanks!

PR #969, opened 2022-11-12
netlist: Enable loading Scheme scripts from %load-path.

vzh:
The issue pointed out by John Doty (@noqsi) on gitter.  He tried to load an add-on by its basename putting it to one of the lepton config dirs to no avail.  The commit here fixes only one part of the issue and its description is as follows:

Previously, scripts specified by the options `-m` and `-l` were required to be absolute file names or file names in the current directory.  It is no longer necessary and now the user can simply set the basename of a file to be loaded, possibly along with its relative path, and the name will be searched for in all paths set in the `%load-path` variable.  Thus, the directories set as user or system data paths and listed in the output of functions `user-data-dir` and `sys-data-dirs` can be used for these purposes.

The other side of the issue, AIUI, and it needs to be discussed, is whether we have to add support for directories dedicated for loading specific stuff.  E.g. directories for tool specific add-ons, for putting backends, and so on.

### DIFF
--- a/liblepton/scheme/netlist.scm
+++ b/liblepton/scheme/netlist.scm
@@ -37,6 +37,7 @@
   #:use-module (lepton gettext)
   #:use-module (lepton library)
   #:use-module (lepton log)
+  #:use-module (lepton file-system)
   #:use-module (lepton object)
   #:use-module (lepton page)
   #:use-module (lepton rc)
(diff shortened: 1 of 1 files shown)


graahnul-grom (approved):

graahnul-grom:
`make distcheck` results in 13 failed `sch2pcb` tests.

vzh:
Dunno why `make distcheck` worked on my home PC.
Well, GH CI reported errors and now I understand why.  `sch2pcb` is the only tool that uses `lepton-netlist` with `-m` in tests.  Fortunately, we have such tests now :-)

I changed the code so now `lepton-netlist` first looks for the file requested by `-l` or `-m` in the current working directory, and then, if nothing found, it looks in the directories set in `%load-path`.  Moreover, unlike in the previous version, it loads only the first file found and reports its name to log. If nothing found, it is reported, either.

graahnul-grom (approved):
:+1:

PR #970, opened 2022-11-14
Fix path to LINGUAS.

vzh:
During compilation, I saw warnings like this for years:

awk: fatal: cannot open file `../../.desktop-i18n/libleptongui.LINGUAS' for reading: No such file or directory

It seemed to be harmless to me as nobody reported it is not so until recently.  John Doty (@noqsi) has reported that on one of Debian derivations a like warning breaks compilation.

I'm not aware for sure of what the code I've changed does, just fixed the path to the files `LINGUAS` and the warnings disappeared.

### DIFF
--- a/m4/lepton-desktop-i18n.m4
+++ b/m4/lepton-desktop-i18n.m4
@@ -55,7 +55,7 @@ gettext is correctly installed, and rerun configure.])
     --localedir=$(DESKTOP_I18N_LOCALE_DIR)/share/locale $(DESKTOP_I18N_LANGS)'
 
   DESKTOP_I18N_LANGS_RULE='DESKTOP_I18N_LANGS = $[$](echo "{ printf \"--lang=%s \", \$[$]1 }" \
-    | awk -f - $(abs_top_builddir)/$(DOMAIN)/po/LINGUAS)'
+    | awk -f - $(abs_top_srcdir)/$(DOMAIN)/po/LINGUAS)'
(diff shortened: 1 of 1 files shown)


graahnul-grom (approved):
There's no more such warning when building (`make`),
though I can still see it in `make distcheck` output:
```
Making all in data
--- liblepton.xml ---
/bin/sh /home/dmn/lepton/lepton.git/lepton-eda-1.9.18/build-tools/desktop-i18n --create --gettext=/usr/local/bin/gettext --domain=liblepton  --localedir=../../.desktop-i18n/share/locale $(echo "{ printf \"--lang=%s \", \$1 }"  | awk -f - /home/dmn/lepton/lepton.git/lepton-eda-1.9.18/_build/sub/liblepton/po/LINGUAS) ../../../../liblepton/data/liblepton.xml.in liblepton.xml
awk: can't open file /home/dmn/lepton/lepton.git/lepton-eda-1.9.18/_build/sub/liblepton/po/LINGUAS
 source line number 2
```

vzh:
Thanks for the point.  I've added one more commit which hopefully fixes `distcheck` as well.  Please check if it does it for you.  Just replaced `abs_top_builddir` with `abs_top_srcdir`.

graahnul-grom:
> Just replaced abs_top_builddir with abs_top_srcdir.

Fixed. Thank you.

PR #971, opened 2022-11-14
Assign Page manager callbacks in Scheme

vzh:
- The contents of the function `x_window_set_current_page()` has
  been moved to Scheme thus allowing for working on tabbed and
  non-tabbed GUI code separately.

- Page manager callbacks are now assigned in Scheme.

### DIFF
--- a/libleptongui/include/page_select_widget.h
+++ b/libleptongui/include/page_select_widget.h
@@ -60,7 +60,7 @@ page_select_widget_update (GschemToplevel* w_current);
 void
 schematic_page_select_widget_set_callback (char *name,
                                            GCallback callback);
-void
+LeptonPage*
 pagesel_callback_selection_changed (GtkTreeSelection* selection,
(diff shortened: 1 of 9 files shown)


graahnul-grom (approved):
IMHO, it's not necessary to introduce a new function
(`set-window-page!`); it just takes pointers instead of fluids.
The work could be done in `window-set-current-page!()`,
avoiding xxx->pointer conversion in many places.
Or we could add more consistently named function like
`*window-set-current-page!()` dealing with pointers,
and call it in `window-set-current-page!()` (please
take a look at the attached patch for example).

[pr971.diff.txt](https://github.com/lepton-eda/lepton-eda/files/10098346/pr971.diff.txt)

vzh:
OK, I've applied your diff with little changes to master as the naming looks more consistent, thank you. The only tweaks are moving the function to the place before it's first called in the module and a bit of indentation.

PR #987, opened 2022-12-19
sch2pcb: Move processing of output file names to Scheme.

vzh:
Apart from processing output file names in Scheme, the program now
supports case insensitive schematic file extensions and reports if
schematic files are not regular or not readable.

### DIFF
--- a/tools/sch2pcb/lepton-sch2pcb.scm
+++ b/tools/sch2pcb/lepton-sch2pcb.scm
@@ -227,11 +227,16 @@
 
 
 (define (add-schematic schematic-name)
+  (define (basename-ci name)
+    ;; basename() is not case-insensitive, so just drop last 4
+    ;; chars (.sch).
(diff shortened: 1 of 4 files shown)


tools/sch2pcb/lepton-sch2pcb.scm
```diff
@@ -225,11 +227,21 @@
 
 
 (define (add-schematic schematic-name)
+  (define (basename-ci name)
```
graahnul-grom, line 230: Unused function?

graahnul-grom (approved):

tools/sch2pcb/lepton-sch2pcb.scm
vzh: Yup, my sloppy buggy half-crafted feature :-)  Or a mistake after rebasing...
Fixed anyway.

PR #989, opened 2022-12-20
schematic: Rewrite x_event_button_released() in Scheme.

vzh:
- Several new accessors for `GschemToplevel`'s fields have been
  added and utilized in the code.
- Accessors for GDK key masks and some fields of the structure
  `GdkEvent` have been added so they are now available in Scheme.
- A new callback function, `callback-button-released()`, has been
  added as a substitute for `x_event_button_released()`.
  
Thus, mouse button release event processing is now carried out in Scheme,
which allows for further refactoring.

PR #994, opened 2023-01-19
schematic: Rewrite x_event_button_pressed() in Scheme

vzh:
The changes are similar to those in #989 but for *button pressed* events.


graahnul-grom (approved):

graahnul-grom:
Recently I noticed strange behavior of copy/paste actions
(`Ctrl+C`/`Ctrl+V`). After some time working in schematic,
`paste` starts pasting not the last copied object, but some
other object(s) copied previously. And eventually it stops
working altogether, showing 'empty clipboard' messages
in the status line. Do you see anything like that?

vzh:
No, I haven't seen that.  If you'll find a reliable way to reproduce this behaviour, please open an issue about it.

vzh:
Looking through the code of `o_buffer_paste_start()` I found one case you might trigger it.  If you select and copy to the clipboard an invisible text object when the invisible text is not hidden, and then try to insert it from the clipboard when the text is hidden, you'll see just this message about empty buffer.  This is how our lowlevel C code works today.

vzh:
Playing with clipboard, I found several interesting things.  Stumbled upon them while working on code in parallel with drawing something in `lepton-schematic`.

One of them, if you cut/copy something in another window of your WM, say in your text editor, and try to paste the contents of the clipboard thereafter, you'll get the message about empty clipboard.  This is because our clipboard shares X selection with other apps though cannot process non-`LeptonObject` objects :-)  See `docs/wiki/geda-wip-clipboard.html` in the repo for more.

Another thing I noticed, GTK2 and GTK3 differently affect the behavior of `lepton-schematic`.  The former toolkit creates different processes when you launch the program more than once, while the latter attaches the next processes to the first one.  In some contexts this may also affect clipboard behavior.  Not sure, though.

Probably we have to restore the previous `Copy to buffer / Paste from buffer` functionality in addition to the clipboard thingy or do something other to prevent data losses :-/

PR #996, opened 2023-02-07
schematic: Rewrite x_event_motion() in Scheme.

vzh:
This PR continues the efforts undertaken in #989 and #994 on
moving mouse processing events related to canvas to Scheme.  The
changes include introducing a couple of new C accessors for
`GschemToplevel` fields, revealing more lowlevel C functions in
the module `(schematic ffi)`, dealing with motion events mostly in
Scheme, and getting rid of a couple Scheme-in-C constructs.  By
the latter I mean eliminating a couple
`scm_dynwind_begin()/scm_dynwind_end()` pairs along with
`g_dynwind_window()` and `g_scm_c_eval_string_protected()` calls.
Thus, the lowlevel C code dealing with `libstroke` no longer
depends on Scheme, and `eval-stroke()` is called from Scheme
module.  Please note, however, that `libstroke` is supported for
GTK2 port only.


graahnul-grom:
Everything works fine, except 2 failed tests in 'make distcheck'.

```
FAIL: component-library-command-grp (1)

Detailed information
====================
File "/home/dmn/lepton/lepton.git/lepton-eda-1.9.18/_build/sub/../../liblepton/scheme/unit-tests/lepton-object-component.scm"
Line 415

Failed form:
(test-assert (component? C1))

Expected value:
#f
Actual value:
#f
Expected error:
#f
Actual error:
#f



FAIL: component-library-command-grp (2)

Detailed information
====================
File "/home/dmn/lepton/lepton.git/lepton-eda-1.9.18/_build/sub/../../liblepton/scheme/unit-tests/lepton-object-component.scm"
Line 416

Failed form:
(test-assert (component? C2))

Expected value:
#f
Actual value:
#f
Expected error:
#f
Actual error:
#f
```

These seems irrelevant.
And why 'Expected value' == 'Actual value'?

graahnul-grom (approved):

vzh:
This looks strange, and on my systems I don't see those errors.  To prove that my code here broke something you could try bisecting starting from `master`.  However, in any case, the code does not change anything in `liblepton` and core Scheme functions, while functions depending on `libleptongui` are not currently tested by our unit test suite...  Just a guess, maybe the version of guile matters on your system?

... after looking into `docs/manual/cmd-component.sh` where the tested components are defined, it may be a race condition since those symbols are defined in that `cmd-component.sh` shell script.

As for *expected value* vs *actual value*, I suspect `test-assert` does not allow `#f` in principle, and the report is just a work-around for this function similar to other ones that really compare two values.  `test-assert` just tests the value it gets is not false (`#f`).

PR #997, opened 2023-02-11
Rewrite `o_buffer.c' in Scheme, refactor, fix and improve its functionality.

vzh:
Thinking about the issue discussed in PR #994 (strange behavior of
copy/paste actions), I decided to play around with the code of
copy/paste actions in REPL, which itself required rewriting in
some other language, and, at the same time, solve another issue:
move several Scheme-in-C hooks out of C into Scheme code.  While
I've not solved the former issue, and I'll mention my thoughts on
how to solve it below, here I present the work, fixes and
improvements:

- The function `buffer-paste-start()` used in the
  `&clipboard-paste()` action has been fixed in that it does no
  longer ignore buffer number given and will work with proper
  buffers when used in custom functions.  Previously, it always
  worked with clipboard due to error in its code.

- `lepton-schematic`'s buffer list has been reimplemented as a
  Scheme list of C `GList`s.

- The file `o_buffer.c` has been thrown out.  A new Scheme module,
  `(schematic buffer)`, has been introduced instead.  Now, it
  exports procedures that replace functionality of the C code
  previously defined in that file.  As of now, the functions and
  variables it exports are:
  - `%clipboard-buffer-id`---always 0,
  - `free-buffers()`---replacement for `o_buffer_free()`,
  - `paste-buffer()`---replacement for `o_buffer_paste_start()`,
  - `selection->buffer()`---replacement for both `o_buffer_cut()` and `o_buffer_copy`,
  - `o_buffer_init()` has no replacement since the buffers are no
    longer C arrays but items of a Scheme list.

  The new exported functions have comments and docstring,
  everything as it should be.

- Rewriting of the above functions required adding several
  accessors for several C structures and several helpers, as
  always.  While in general this increased the amount of the
  resulting code, it will pay off in future, I believe.  One
  example is the field `place_list` of the struct `GschemToplevel`
  which now allows for manipulating objects placed on canvas in
  Scheme.  There are more, of course.

- One call of `callback_cancel()` has been moved to Scheme.  It's
  a step forward towards revealing access keys that will help
  access to lower level Scheme-in-C functions hidden in depth
  under it.

- Paste operation does no longer change clipboard contents.  Only
  cut and copy operations do.

- The number of paste buffers is now 10 and the user can cut,
  copy, and paste to/from any of them by prefixing the usual
  commands `Ctrl-X`, `Ctrl-C`, and `Ctrl-V` with a numerical
  prefix from 0 to 9.  Zero prefix does the same that the
  keystroke without prefix does, that is, works with clipboard
  buffer.  Appropriate actions and keystrokes have been added for
  such dealing with buffers.  IMO, it's a temporary solution, for
  longer term I would better re-use `Y C` and the like shortcuts,
  see below why.

- Paste action now reports if there is no visible objects to
  insert.  There were situation when it did not work as expected.
  If the user selected a hidden text object in the "hidden text
  enabled" mode, and then tried to insert it afterwards when the
  hidden text was disabled, that is, there were no visible objects
  to insert, it used to report wrong message "Empty buffer" while
  the buffer contained only hidden text objects.  This has been
  fixed by outputting a slightly more meaningful message in such
  cases.

- Support for pasting objects by clicking on menu item `Ctrl-V`
  has been restored (or implemented? :-)).  I've spent quite much
  time to get why it didn't work so far.

### Future directions---what I believe should be improved in copy-paste

My thoughts aloud, please let me know what do you think of them.

Now again about the issue with clipboard copy/paste.  I've already
mentioned in the discussion in the PR #994 that clipboard can
change its contents under influence of other apps working with
copy-paste.  Thus, if you copy anything into primary X selection
from, say, your favorite text editor, and try to use `Ctrl-V` in
`lepton-schematic`, you'll suddenly find that your clipboard
buffer is empty as it doesn't support plain text objects for
import-export, only C Lepton objects.

To solve the issue, I suggest to restore first buffer as the main
buffer for cutting/copying/pasting in `lepton-schematic` and
provide shortkeys for it like the following shortkeys commented
out in `libleptongui/scheme/conf/schematic/keys.scm`:

```scheme
;(global-set-key "Y C" 'buffer-copy1)
;(global-set-key "Y U" 'buffer-cut1)
;(global-set-key "Y P" 'buffer-paste1)
```

add menu items for them, and clearly state in the documentation
what the difference between main buffer (1) and clipboard buffer
(0) is, noticing the value, deficiencies, and usage of both
buffers.

We could also add similar shortcuts prefixed with numbers to allow
for more buffers.  I've already done this in this commit set,
though I believe "2 Ctrl-V" should be replaced with "2 Y P"
etc. in the code.


graahnul-grom:
> if you copy anything into primary X selection
from, say, your favorite text editor, and try to use Ctrl-V in
lepton-schematic, you'll suddenly find that your clipboard
buffer is empty

This is true for X *clipboard*, not "primary X selection".
The terminology here is quite confusing:
X primary selection, X secondary selection, X clipboard
and also there is "selection" object in lepton-schematic. :)

> restore first buffer as the main
buffer for cutting/copying/pasting in lepton-schematic

Buffers do not work between different lepton-schematic instances.

graahnul-grom (approved):
So far, everything works fine; I can't reproduce
problems with clipboard I described in #994.
Thank you.

vzh:
> Buffers do not work between different lepton-schematic instances.

It maybe was so so far without changes in this commit set.  Or maybe GTK works differently on different platforms even with these changes?  Let's not guess yet.  On my home PC I can easily copy-paste objects between `lepton-schematic` window instances using `1 C-C` and `1 C-V` or `3 C-C` and `3 C-V`.  Please let me know if things do not work properly for you.

PR #998, opened 2023-02-11
Get rid of GschemToplevel's field 'last_drawb_mode'.

vzh:
The field introduced back in 2000 seems to do nothing these days.  I've looked into its usage while working on PR #997 and found it pretty useless.  The only place it was evaluated to do some decision was the code related to net rubberbanding in move mode.  I couldn't trigger any issues after removing it and replacing it directly with the `actionfeedback_mode` field.  It looks like someone forgot their slippers when they left.

If you have some strong objections, I'll remove the second commit here while leaving the first one intact :-)

### DIFF
--- a/libleptongui/include/gschem_defines.h
+++ b/libleptongui/include/gschem_defines.h
@@ -1,7 +1,7 @@
 /* Lepton EDA Schematic Capture
  * Copyright (C) 1998-2010 Ales Hvezda
  * Copyright (C) 1998-2014 gEDA Contributors
- * Copyright (C) 2017-2022 Lepton EDA Contributors
+ * Copyright (C) 2017-2023 Lepton EDA Contributors
  *
  * This library is free software; you can redistribute it and/or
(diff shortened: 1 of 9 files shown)


graahnul-grom (approved):
The same 2 failed tests as in #996.

vzh:
> The same 2 failed tests as in #996.

Try `distcheck` on `master` branch first.

graahnul-grom:
Same failed tests even with master branch.
Repeated under new user account in clean environment.

vzh:
As the tests work okay on Debian (on my PC) and Ubuntu (GH), the issue may be specific to your OS, maybe to your guile version.  Probably we need to skip those two tests for some software configurations.  BTW, you could try to launch the failed script from your `lepton-schematic`.  Hopefully, you'll get some hints then.

graahnul-grom:
The same with guile3 and single/multiple make jobs.
Will investigate further.
(No problems with 'make distcheck' in Linux virtual machine.)

PR #1008, opened 2023-03-17
sch2pcb: Rewrite sch2pcb_add_elements() in Scheme.

vzh:
Well, this is a partial rewriting (along with a little refactoring in C).  I tried to make things clear by doing changes step by step and trying to keep them intelligible in every commit, and to avoid errors which may appear due to overcomplicated program code and insufficient testing.

After thinking much about testing it, I'm going to add FFI siblings for C functions and test them using our UTS.  (Obviously, I cannot test every aspect manually myself.)  It's still error-prone and we need much more testing on the user side.  Maybe we have to roll a new release?

Hierarchy support proposed by Karl Hammar on the geda-user list is still in my TBD list.  However, with the current rate of reviews and bug reports on topic branches... Well, maybe there will be some volunteers who might test my code more often in future??

### DIFF
--- a/liblepton/src/sch2pcb.c
+++ b/liblepton/src/sch2pcb.c
@@ -1,6 +1,6 @@
 /* gsch2pcb
  * Copyright (C) 2003-2016 gEDA Contributors
- * Copyright (C) 2017-2022 Lepton EDA Contributors
+ * Copyright (C) 2017-2023 Lepton EDA Contributors
  *
  *  Bill Wilson    billw@wt.net
  *
@@ -1233,6 +1233,19 @@ sch2pcb_buffer_to_file (char *buffer,
  * pcb file will only have new elements.
  */
 void
+sch2pcb_verbose_report_new_file_element_required (PcbElement *el,
+                                                  gboolean is_m4)
+{
+  if ((sch2pcb_get_verbose_mode () != 0)
+      && !is_m4)
+    printf ("%s: need new file element for footprint  %s (value=%s)\n",
+            pcb_element_get_refdes (el),
+            pcb_element_get_description (el),
+            pcb_element_get_value (el));
+}
+
+
+void
 sch2pcb_add_elements (FILE *f_in,
                       FILE *f_out)
 {
@@ -1277,12 +1290,7 @@ sch2pcb_add_elements (FILE *f_in,
               || (is_m4
                   && sch2pcb_get_force_element_files ()))
           {
-            if ((sch2pcb_get_verbose_mode () != 0)
-                && !is_m4)
-              printf ("%s: need new file element for footprint  %s (value=%s)\n",
-                      pcb_element_get_refdes (el),
-                      pcb_element_get_description (el),
-                      pcb_element_get_value (el));
+            sch2pcb_verbose_report_new_file_element_required (el, is_m4);
             if ((sch2pcb_get_verbose_mode () != 0)
                 && is_m4
                 && sch2pcb_get_force_element_files ())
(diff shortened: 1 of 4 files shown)


liblepton/src/sch2pcb.c
```diff
@@ -1169,6 +1174,39 @@ pkg_to_element (FILE * f, gchar * pkg_line)
...
 
+
+FILE*
+sch2pcb_open_file_to_write (char *filename)
+{
+  FILE *f_out;
+
+  if ((f_out = fopen (filename, "wb")) == NULL)
+  {
+    fclose (f_out);
```
graahnul-grom, line 1185: `fclose( NULL )` will fail. Om my system it even causes `SIGSEGV`.

vzh: Thank you, indeed, I read somewhere that the function works differently on different systems, and may even segfault on NULL arguments.

graahnul-grom (approved):
Tested also with the latest Guile 3.0.9. And though I managed
to test changes in this PR (with the attached patch applied),
other code utilizing `system*()` is broken.
In 3.0.9 this function has been reimplemented with `posix_spawn(3)`,
here's the NEWS entry:
https://git.savannah.gnu.org/cgit/guile.git/tree/NEWS?h=v3.0.9#n8
On my system it always returns 127.
Even in simple REPL tests, like `( system* "ls" )`.
If possible, could you please test this on your system with 3.0.9?

[sch2pcb_guile309_system.diff.txt](https://github.com/lepton-eda/lepton-eda/files/11099960/sch2pcb_guile309_system.diff.txt)

graahnul-grom:
> Maybe we have to roll a new release?

IMO, it's not a good idea, given the current half-baked
state of things. It would be very difficult to support
such release.

> Well, maybe there will be some volunteers who might test my code more often in future??

Get Karl aboard! :)

All in all, straightforward translation of
lepton-sch2pcb from C to Guile is not an easy task.
And, as it turns out, perhaps this is not an optimal way.
Now it appears that gradually designing it from the
ground up would've been be easier, not to mention risks
of breaking something.

vzh:
> Tested also with the latest Guile 3.0.9. And though I managed to test changes in this PR (with the attached patch applied), other code utilizing `system*()` is broken. In 3.0.9 this function has been reimplemented with `posix_spawn(3)`, here's the NEWS entry: https://git.savannah.gnu.org/cgit/guile.git/tree/NEWS?h=v3.0.9#n8 On my system it always returns 127. Even in simple REPL tests, like `( system* "ls" )`. If possible, could you please test this on your system with 3.0.9?
> 
> [sch2pcb_guile309_system.diff.txt](https://github.com/lepton-eda/lepton-eda/files/11099960/sch2pcb_guile309_system.diff.txt)

I cannot promise I'll test it soon.  Recently, I've changed my day job, and my schedule changed as well.  Please bear with me.  Hopefully, things will settle down soon.

vzh:
> > Maybe we have to roll a new release?
> 
> IMO, it's not a good idea, given the current half-baked state of things. It would be very difficult to support such release.

Agreed, let's finish with sch2pcb first.

> 
> > Well, maybe there will be some volunteers who might test my code more often in future??
> 
> Get Karl aboard! :)

Sure, he's already in.

 
> All in all, straightforward translation of lepton-sch2pcb from C to Guile is not an easy task. And, as it turns out, perhaps this is not an optimal way. Now it appears that gradually designing it from the ground up would've been be easier, not to mention risks of breaking something.

Not easy but doable.  Doing it this way is my attempt to avoid situations we had with other lepton tools, in which bugs were revealed by users years after releasing them.

PR #1013, opened 2023-04-10
Add --disable-doc configure option

graahnul-grom:
It's off by default. When turned on, disables build
and install of manuals, wiki, readme files and man pages.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -6,13 +6,17 @@ if ENABLE_CONTRIB
 CONTRIB_DIR=contrib
 endif
 
+if ENABLE_DOC
+DOC_DIR=docs
+endif
+
 SUBDIRS = \
 	liblepton \
 	${ATTRIB_DIR} \
 	libleptongui \
 	tools \
 	symbols \
-	docs \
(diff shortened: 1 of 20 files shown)


vzh:
Internal `define`s would look better than those ugly `let*`s in the last commit.

And I would suggest to add the info on the new `--disable-docs` feature to some of the docs it is going to disable :-)

vzh (approved):

graahnul-grom:
Added `--disable-doc` to the manual
(2.7 - Installation/Configure options).

Changed `let` to `define`.
Frankly, can't see the difference here.
Both are ugly. Like anything else in guile; especially so
in combination with the dog's breakfast coding style (tm). :P

vzh:
I was reluctant to discuss this but you started first.  I felt the same when looked at your DBCS reminding me BASIC or FORTRAN though distastefully altered for lisp beginners (starting to imagine a mathematician writes a tutorial about differential equations for first-graders, one equation per one textbook).  For me, Racket style I mentioned in wiki looks awesome.  Nothing personal, though :-P

PR #1018, opened 2023-05-05
sch2pcb: Rewrite sch2pcb_update_element_descriptions() in Scheme.

vzh:
I've added a couple of tests to make sure it works properly.

IMO, the option `--fix-elements` is superfluous.  In the first place, it was introduced decades ago to aid transition from `gschem2pcb` to `gsch2pcb`.  I may be wrong, but today we could live without it.  If anything changed in schematics, their pcb files have to be modified according to the changes.  Please let me know if you have any different opinion.

### DIFF
--- a/liblepton/include/liblepton/sch2pcb.h
+++ b/liblepton/include/liblepton/sch2pcb.h
@@ -303,5 +303,6 @@ sch2pcb_get_verbose_mode ();
 void
 sch2pcb_update_element_description (FILE *f_out,
                                     char *buf,
-                                    PcbElement *element);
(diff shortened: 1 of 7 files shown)


graahnul-grom (approved):

graahnul-grom:
> IMO, the option `--fix-elements` is superfluous. In the first place, it was introduced decades ago to aid transition from `gschem2pcb` to `gsch2pcb`. I may be wrong, but today we could live without it. If anything changed in schematics, their pcb files have to be modified according to the changes. Please let me know if you have any different opinion.

I think you are right.

PR #1022, opened 2023-05-15
Selection of locked components in lepton-schematic

### LINKED ISSUE #651: Cannot select a locked object
Obviously you don't want to be able to edit it, but it's confusing for us newbies, so maybe it could produce a dialogue telling you it's locked, and possibly offering to unlock it? I suppose that should only happen if you click on an individual object rather than multi-selecting by dragging an area.

graahnul-grom:
Add new `&edit-select-locked` action; `Edit->Select Locked`
menu item and `E K` shortcut for it.
It selects locked components on current page one by one.
If there is none, shows a message box about that.
Information about currently selected component is shown
in the log and, briefly, in the status bar.
Implemented in new `(schematic sellock)` module.
Based on my [select-locked](https://graahnul-grom.github.io/scm/select-locked.scm.html) script.

Closes #651.

select-locked-0

select-locked-1

select-locked-2

select-locked-3

### DIFF
--- a/docs/manual/lepton-schematic.texi
+++ b/docs/manual/lepton-schematic.texi
@@ -310,6 +310,11 @@ that is, makes them not selectable by just clicking on them.
 @clicksequence{Edit @click{} Unlock} (@kbd{E Shift-L}) unlocks
 selected locked objects.  @xref{Editing schematics}.
 
+@item Select Locked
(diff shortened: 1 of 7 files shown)


vzh (approved):
Looks much better :-)

PR #1027, opened 2023-07-02
Simplify the toolbar creation code

graahnul-grom:
- *append* toolbar items by default; no need to supply index
- easily add radio buttons: automatically assign radio group
- add `radio-group-end()`: ends the group and sets the active button
- add `make-toolbar-separator()`
- make `g_action_eval_by_name()` accessible in Scheme
- add 2 new toolbar buttons: "Down Schematic", "Down Symbol"
- move the "Add Text..." button, so that radio buttons for
  adding nets, buses and select are grouped together.
What other actions should be added to the toolbar?
(I don't use it at all).

tb_369_toolbar

### DIFF
--- a/libleptongui/scheme/schematic/ffi.scm
+++ b/libleptongui/scheme/schematic/ffi.scm
@@ -476,6 +476,8 @@
             a_zoom_box_motion
             a_zoom_box_start
 
+            g_action_eval_by_name
+
             parse-gschemrc
             ))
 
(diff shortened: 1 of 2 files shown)


vzh (approved):
I don't use toolbar, either, so my opinion is not of much value here.  Probably buttons for all primitives could be added, dunno.  To provide them, we would have to draw some nice icons for them as peter-b did long ago for currently existing buttons. `g_eval*()` functions are subjects to be thrown out in future, though currently it doesn't matter.  Your work looks nice, both visual UI changes and the code, thank you!

graahnul-grom:
Спасибо на добром слове, Володя!

PR #1033, opened 2023-07-11
sch2pcb: Avoid test failures when component library functions are used in gafrc.

vzh:
Recently, I suddenly found, that Lepton test suite fails if I have in my `~/.config/lepton-eda/gafrc` the following line:
`(component-library-search "~/lepton/symbols")`

As it stands out, two of our tools, namely `lepton-export` and `lepton-shell` cannot survive `make distcheck` tests with this configuration setting.

The branch fixes this by exporting the `(lepton library)` module in both tools.

### DIFF
--- a/tools/cli/scheme/lepton-export.scm
+++ b/tools/cli/scheme/lepton-export.scm
@@ -1,7 +1,7 @@
 ;;; Lepton EDA command-line utility
 ;;; Copyright (C) 2012 Peter Brett <peter@peter-b.co.uk>
 ;;; Copyright (C) 2014-2016 gEDA Contributors
-;;; Copyright (C) 2017-2022 Lepton EDA Contributors
+;;; Copyright (C) 2017-2023 Lepton EDA Contributors
 ;;;
 ;;; This program is free software; you can redistribute it and/or modify
 ;;; it under the terms of the GNU General Public License as published by
@@ -21,6 +21,7 @@
              (system foreign)
              (lepton ffi)
              (lepton gettext)
+             (lepton library)
              (lepton page)
              (lepton rc)
              (lepton srfi-37)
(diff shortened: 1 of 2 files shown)


graahnul-grom (approved):

graahnul-grom:
Also `make distcheck` fails unless I add use-modules
`(geda deprecated)` to tools/archive/lepton-archive.scm
and tools/cli/scheme/lepton-cli.scm
(can't find `scheme-directory()`), and `(lepton color-map)`
to tools/netlist/lepton-netlist.scm (can't find `print-color-map()`).

vzh:
`print-color-map()` can also be used in some rc files.  There is no direct use of the function in the code elsewhere.  It would be nice to learn which settings could trigger such failures, at least on your system.  If the version `1.9.18` works OK there, `git bisect` could help.  I myself have no failures you've mentioned above.

vzh:
So I see two possible solutions or work-arounds wrt the issue you're talking about here:
- Make all tools reading `gafrc` during their initialization load necessary guile modules in order to provide the functions you've mentioned above.  The set of tools may be incomplete, just to fix the current issues you have, though we could fix other tools later as people report similar issues for them.  This option would need your report/PR to be solved.
- Just mention in our docs we don't support those functions in `gafrc` any more.  I could add such info myself if you wish.

WDYT?

graahnul-grom:
I'll file a PR.

Though `scheme-directory()` is deprecated, `print-color-map()`
is not. In the `gafrc` [documentation](https://lepton-eda.github.io/lepton-manual.html/gafrc.html) there's info on primitive-load()'ing
print color schemes, therefore, indirectly this function can be
invoked from a `gafrc` file.

And I'm not sure about deprecation of `scheme-directory()`.
It's just a convenient function which expands env vars.
Could be moved to the `(lepton os)` module and documented. 
WDYT?

vzh:
It could be, though I don't like it.  In a nutshell, it's already very simple and it would be better for users who're trying to be more familiar with Scheme to know its basics.  What's better, writing two functions, or one function and one `use-module` stanza?  How many users use `scheme-directory()` in their rc files?  You could ask about it in the `geda-user` ml (I won't do, I know somewhere about).  After all, there is this issue tracker.  Open a request about surviving the function here and we'll see, how many users really need it.

graahnul-grom:
I think, you are right. `add-to-load-path()` and `getenv()`
are simple enough, no need to introduce yet another function.

PR #1038, opened 2023-08-11
Avoid test failures: scheme-directory() and print-color-map()

graahnul-grom:
If these functions are called (possibly, indirectly) from
the user's `gafrc` configuration file, unit tests might fail
for some tools in the suite loading this file.

`scheme-directory()` is defined in the `(geda deprecated)`,
`print-color-map()` - in the `(lepton color-map)` module.

Add missing `use-modules` where appropriate.

### DIFF
--- a/tools/archive/lepton-archive.scm
+++ b/tools/archive/lepton-archive.scm
@@ -1,4 +1,4 @@
-;;; Copyright (C) 2019-2022 Lepton EDA Contributors
+;;; Copyright (C) 2019-2023 Lepton EDA Contributors
 ;;;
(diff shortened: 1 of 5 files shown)


vzh (approved):
For me, it's counterintuitive that, for example, `lepton-shell` or `lepton-netlist` require loading of color map related functions to work.  Fair enough, we'll deal with the functions later, after rc files won't be needed at all. (Hopefully, the backwards compatibility won't kill the project before.)

graahnul-grom:
> For me, it's counterintuitive that, for example, lepton-shell or
> lepton-netlist require loading of color map related functions to work.

Yes, it is.

PR #1041, opened 2023-08-24
schematic: Make toolbar callbacks remember <window> they are related to.

vzh:
Otherwise, some lower level functions used in the callbacks may be unfamiliar of the current window the callbacks are related to, and therefore fail.

On one of my systems, the issue appears after 1 to 3 clicks on the Undo toolbar button.  `lepton-schematic` crashes reporting that `current-window` is `#f` due to implicit using `current-window()` function in `in-action?()` that is used in the button's callback.

The commit proposed here solves the issue on my system.

### DIFF
--- a/libleptongui/scheme/schematic/toolbar.scm
+++ b/libleptongui/scheme/schematic/toolbar.scm
@@ -28,6 +28,8 @@
   #:use-module (schematic callback)
   #:use-module (schematic ffi)
   #:use-module (schematic undo)
+  #:use-module (schematic window global)
+  #:use-module (schematic window foreign)
 
   #:export (make-toolbar))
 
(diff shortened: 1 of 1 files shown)


graahnul-grom:
Clicking on the `undo` toolbar button:
```
In schematic/undo.scm:
     87:6  2 (callback-edit-undo #<pointer 0x82932d800> #<pointer 0x829223e00>)
In schematic/action-mode.scm:
    41:45  1 (in-action? _)
In unknown file:
           0 (scm-error wrong-type-arg window->pointer "Wrong type argument in position ~A (expecting ~A): ~A" (1 <window> #f) #f)

ERROR: In procedure scm-error:
In procedure window->pointer: Wrong type argument in position 1 (expecting <window>): #f
```

graahnul-grom (approved):

PR #1060, opened 2024-02-15
Enable orthogonal net connection to bus in lepton-schematic GUI

vzh:
By default, during connection a net connecting to a bus changes
its direction by 45 degrees and connects diagonally.  It would be
nice to enable orthogonal connections.  The solution is to enable
setting the size of diagonal net part in the configuration to
zero.  In this branch setting of `bus-ripper-size` to zero is
enabled, which along with the setting `bus-ripper-type=net` allows
for creating orthogonal net-bus connections.


vzh:
As nobody minds, I'll rebase and merge the branch.

PR #1061, opened 2024-04-06
lib: Allow for connections of nets to ends of buses

vzh:
It is convenient when you like to connect nets to buses without busripper symbols where one net starts the bus it connects to. Connection without busripper symbols can be enabled by setting the 'schematic::bus-ripper-type' configuration option to 'net'.

So far, net-bus connections were allowed only in the middle of buses, not to their ends.

AFAIK, there is no direct mention of such connections style in IEK though in my practice, I've seen lots of schematics done this way.


vzh:
Well, fair enough, as nobody cares...

PR #1036, opened 2023-08-10
gui[gtk3]: Fix dockable widgets displaying.

vzh:
Before this fix, some widgets, e.g. text properties, did not show in dockable view on my system, at least in GTK3 port.

PR #1062, opened 2024-05-03
Fix crashes of the GTK3 port of lepton-schematic when clicking on menus

### LINKED ISSUE #1058: segfault crash is lepton-schematic
I'm getting segfault in lepton-schematic when opening menu item in lepton-schematic only in gtk3 version.

vzh:
This branch fixes two types of crashes:

-   Crashes when clicking on items in tabs' menu or main menu items.
-   Crashes on closing one of open **lepton-schematic** windows, it
    doesn't matter whether with mouse click on the menu item or with
    a keyboard shortcut.

The issue was first [reported](https://github.com/lepton-eda/lepton-eda/pull/1036#issuecomment-1676992859) by @graahnul-grom in #1036.  Please
see the discussion there.  He triggered it [with Guile 2.2.7 on
FreeBSD 13.2-STABLE](https://github.com/lepton-eda/lepton-eda/pull/1036#issuecomment-1687053422).  And the same behavior was found [with Guile
3.0](https://github.com/lepton-eda/lepton-eda/pull/1036#issuecomment-1722495258).  Menus [worked correctly](https://github.com/lepton-eda/lepton-eda/pull/1036#issuecomment-1724315966) in Lepton 1.9.18 until the commit
b60bacb [was introduced](https://github.com/lepton-eda/lepton-eda/pull/1036#issuecomment-1736089446).  In that commit, the GTK3 port of Lepton
started to use the newer `GSimpleAction` interface for the menu
actions.  On my own Debian PCs, however, I [couldn't reproduce the
issue](https://github.com/lepton-eda/lepton-eda/pull/1036#issuecomment-1937466655) for some time until I [upgraded several packages](https://github.com/lepton-eda/lepton-eda/pull/1036#issuecomment-1939216154) including
`gcc`, `guile`, `gtk` libraries and their dependencies.

The program seems to have started crashing not because of Guile
since both its versions 2.2 and 3.0 behave the same way.  Besides,
the version of the Guile package on my system after upgrade
changed only its revision number.

I don't think it was upgrading GCC with its dependencies that
triggered the issue as another build system, cmake, is usually
used on FreeBSD.

Therefore I think the reason of crashes is more likely upgrading
the GTK libraries and/or their dependencies such as **libglib**
which meantime was upgraded from 2.74 to 2.78.  While the older
versions of the libraries somehow tolerated some wrong formed
macros specified in the code, the new code could become more
strict.

Some prerequisites:

-   `g_signal_emit_by_name()` is a macro which, by definition, can
    take various number of arguments.  That is why the compiler
    swallows this wrong function call.
-   Guile GC seems to affect it.  It looks like a race condition
    affected both by Guile GC and GTK functions dealing with
    callback arguments, which possibly causes use-after-free,
    double-free, or some other memory corruption which leads to
    segfault.

**The real issue** is the wrong usage of the function
`g_signal_emit_by_name()` with the `activate` signal defined for
`GSimpleAction` type objects.  The signal must have one parameter
in the call which may be `NULL`.  In the callbacks defined for
activating menu items, the function was called without any
parameters, which has been fixed in two cases: for the main menu,
and for the tabs popup menu.  As I stated before, the function is
a macro, and the C compiler tolerates such wrong calls while this
may have side effects when compiling using some software
combinations.

The second commit in the branch removes a duplicate signal
connection.  The function `lepton_action_create_menu_item()`
creating a menu item widget connects the "activate" signal to it
before returning the widget.  The code removed in the commit
connected the same signal to the returned widget second time.

Closes #1058

### DIFF
--- a/libleptongui/src/x_menus.c
+++ b/libleptongui/src/x_menus.c
@@ -254,11 +254,6 @@ schematic_window_create_main_popup_menu (GschemToplevel* w_current)
     GSimpleAction* action = g_simple_action_new (e.action, NULL);
     menu_item = lepton_action_create_menu_item (action, gettext (e.name), NULL);
 
-    g_signal_connect (menu_item,
-                      "activate",
-                      G_CALLBACK (&on_menu_item_activate),
-                      action);
-
     g_signal_connect (action,
                       "activate",
                       G_CALLBACK (g_menu_execute),


vzh:
I've inadvertently pushed to another trunk, and after repairing this branch is now really merged upstream.  I have only to manually deal with the issues it could close automatically.

PR #1081, opened 2025-04-28
hierarchical schematics, doc and symbols

felix-salfelder:
Both ports and parameters are part of the interface to a device. A schematic drawing often represents a device, currently with an implicit interface. I propose to streamline hierarchical circuit modelling based on existing work in the documentation, enabling the unambiguous explicit declaration of parameters and ports.

- extend the "schematic-port" section in the documentation
- add a "schematic-parameter" section following the pattern
- provide symbols that make use of attributes described therein.
 
NB: Parameters exist in Spice, and the proposed logic could be used in Spice netlisting, too. Current Spice netlisting workarounds and heuristics indeed require netlisting and fail to work with tools other than Spice (please discuss elsewhere).


vzh:
Felix, thank you!

PR #1096, opened 2025-09-18
sch2pcb: Get rid of three functions working directly with C file descriptors

vzh:
Three **sch2pcb** functions working directly with file descriptors
in C have been replaced with superior Scheme functions outputting
data to Scheme ports, which simplified code in several places.


vzh:
The branch has been merged into `master`.  However, GH cannot cope with this sequence of actions.

PR #1101, opened 2025-09-28
Fix two bugs in the callbacks processing mouse button events

### LINKED ISSUE #1100: lepton-schematic crashes
Latest version in both gtk3 and gtk2 of lepton-schematic.
When using middle mouse button to move schematic the entire app crashes.

Backtrace:
In ice-9/boot-9.scm:
  1755:12  7 (with-exception-handler _ _ #:unwind? _ # _)
In unknown file:
           6 (apply-smob/0 #<thunk 7fc83ebd0300>)
In ice-9/boot-9.scm:
    724:2  5 (call-with-prompt _ _ #<procedure default-prompt-handle…>)
In ice-9/eval.scm:
    619:8  4 (_ #(#(#<directory (guile-user) 7fc83ebd3c80>)))
In ice-9/boot-9.scm:
   2858:4  3 (save-module-excursion _)
  4408:12  2 (_)
In unknown file:
Exception thrown while printing backtrace:
In procedure primitive-call-ip: Wrong type argument in position 1 (expecting PRIMITIVE_P): #<procedure 7fc83ec734c0 (_)>

schematic/undo.scm:101:18: In procedure callback-button-released:
Current window is unavailable.

vzh:
- A syntax rule has been added to process mouse button events on
  canvas the same way.
- Mouse button callbacks crashes have been fixed.  The crashes
  happened since the current window fluid was undefined in the
  callback scope.
- Crashes in the mouse stroke processing code due to absent
  double-to-int coordinate conversion have been fixed as well.
  The issue emerged after rewriting the stroke processing code in
  Scheme.  To trigger it, the config key
  `schematic.gui::middle-button` has to be set to `stroke`.

Closes #1100

### DIFF
--- a/lepton-eda/scheme/schematic/window.scm
+++ b/lepton-eda/scheme/schematic/window.scm
@@ -352,7 +352,11 @@ zooming."
         (if (null-pointer? *page)
             ;; If there is no page, terminate event.
             TRUE
-            (process-func *canvas *event *window)))))
+            ;; Some underlying functions have to know what window
+            ;; they operate on.  Set the current window explicitly
+            ;; as it is not defined for C GLib callbacks.
+            (with-window *window
+              (process-func *canvas *event *window))))))
 
 
 (define (callback-button-released *canvas *event *window)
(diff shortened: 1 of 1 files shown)


drforbin:
Did not fix it...same problem!

PR #1103, opened 2025-10-03
Add support for running built Lepton tools without installation

### LINKED ISSUE #1084: unable to run tools from git repo
With git commit 06816166a

    $ ./autogen.sh
    $ mkdir build; cd build
    $ ../configure; make

(seems to work fine)

then

      $ tools/schematic/lepton-schematic
      ;;; WARNING: compilation of /home/.../tools/schematic/lepton-schematic failed:
      ;;; no code for module (lepton m4)
      [backtrace stuff]

Seems to be looking for other files?

     $ make install # maybe here?
     $ /usr/local/bin/lepton-schematic
     [compiling stuff] .. error
     $ rm ~/.cache/guile/ -rf
     $ /usr/local/bin/lepton-schematic
     [compiling stuff] .. error, still.
     $ /usr/local/bin/lepton-schematic
     ... undefined symbol in /usr/lib/x86_64-linux-gnu/liblepton.so

Ok ok, but:

    $ export LD_LIBRARY_PATH=/usr/local/lib
    $ /usr/local/bin/lepton-schematic
     ... undefined symbol in /usr/lib/x86_64-linux-gnu/liblepton.so, still

So I also tried

    root# ldconfig
    $ hash -r
    $ which lepton-schematic
    /usr/local/bin/lepton-schematic <= PATH looks ok.
    $ /usr/local/bin/lepton-schematic
    ... undefined symbol in /usr/lib/x86_64-linux-gnu/liblepton.so ...
    $ ldconfig -p |grep liblepton.so
        liblepton.so.11 (libc6,x86-64) => /usr/local/lib/liblepton.so.11
        liblepton.so.11 (libc6,x86-64) => /lib/x86_64-linux-gnu/liblepton.so.11
        liblepton.so (libc6,x86-64) => /usr/local/lib/liblepton.so
        liblepton.so (libc6,x86-64) => /lib/x86_64-linux-gnu/liblepton.so

I use an ordinary Debian system, and normally things involving autotools and ld.so just work. There is no INSTALL file in Lepton, and README.md just says "make install". This shouldn't be required, but sounds failsafe. It does not work for me. What am I missing?

TIA

vzh:
The branch adds support for running just built but not installed
Lepton tools after compilation.  While there are some issues yet
(e.g. **lepton-netlist** cannot find legacy backends, l10n doesn't
work properly in GUI tools), the basic issue has been fixed.
**lepton-schematic** and **lepton-attrib** GUIs work OK for me.
You can build and run the GTK2 ports of the programs while having
their GTK3 ports installed.

The following changes have been made to achive this:

- The tools now check whether they run from the installation
  directory or from another place (i.e. from the build directory).
  To make this work, a function returning the path a suite tool is
  launched from has been added.  Currently, this works only in
  Linux.  The hints how to do it for other OSes are described in
  code comments.  The tools will report if running without
  installation on an OS is not supported.
- Several variables defined previously in autoconf's *m4* files
  have been moved to Scheme code or removed:
  - The variable `LEPTON_SCHEME_DIR` has been removed and replaced
    with another variable defined in Scheme that returns the
    Scheme modules directory.
  - The variable `LEPTON_CCACHE_DIR` has been replaced with a
    Scheme variable.
  - The variables `bindir` and `LEPTONDATADIR` are inlined in the
    code of all Lepton tools.
  - The variable `BITMAP_DIRECTORY` has been removed as it is no
    longer needed: the bitmap path is determined and set during
    **lepton-schematic** start-up (and it is used only in that
    program).
  - The variable `PACKAGE_TARNAME` has been inlined in the Scheme
    code to define paths to the package directories.
- The variables `abs_top_builddir` and `abs_top_srcdir` have been
  made available in Scheme code.
- Two new functions have been introduced. One searching for system
  data paths, and one searching for system config (RC) files.  The
  deprecated variables `geda-data-path` and `geda-rc-path` are no
  longer used in any RC scripts.  The new config lookup functions
  are used instead in RC files.
- The `XDG_CONFIG_DIRS` and `XDG_DATA_DIRS` environment variables
  are now set to predefined values when running from the build
  directory, which allows for getting necessary program data from
  the repository subdirectory `lepton-eda` when the user runs GUI
  programs after building without installation.  The system RC
  files have been amended to support the new behavior.

The repository layout has the following amendments:
- The contents of `liblepton/lib` has been moved to `lepton-eda`
  directory in the root of the repository.
- The config files of `libleptongui` have been moved to the place
  where `liblepton`'s files reside.
- Symbol directories have been moved to `lepton-eda`.
- Symbol library autoload script has been moved to `lepton-eda` as
  well.
- Icons for **lepton-schematic** are now installed as symlinks on
  the configuration stage which makes several of them (not all)
  available for the program when running without installation.

GUI icons have been fixed for both GTK2 and GTK3 ports:
- The icons named `edit-undo` and `edit-redo` seem to be
  unavailable for the GTK2 port.  This has been fixed by
  replacing them with `undo` and `redo`, correspondingly.
- Replacing icons `gtk-goto-bottom` and `gtk-go-down` with
  `go-bottom` and `go-down` makes them displayed in the GTK2
  port of **lepton-schematic** as well as improves the look of
  the GTK3 port as the new icons are taken from the same theme
  as other icons in the toolbar (at least on my system).  The
  former pair looked differently than other stock icons.

-----

One special commit has been made to resolve a long standing issue
emerging every now and then for years, and I spent much trouble on
finding and solving it while working on the branch.  It caused
sporadic test suite failures on log rotation when doing `make
distcheck`.  Here is the full description from the commit which
workarounds the issue:

When several processes of a lepton tool, say **lepton-netlist**,
are running in parallel and producing many log files, they can
also try to delete same files while rotating logs.  It causes
misleading error messages like

> ERROR: Failed to load RC file [.../system-gafrc]: No such file or directory

as the log rotation starts from the RC script `system-gafrc`, and
the current RC handlers are not smart enough to show the real
culprit.

Moreover, the user is not always able to find an actual error even
in the log files as they might be deleted by another process that
started log rotation at the same time.

The implemented work-around for the issue is prohibiting the log
rotation when the environment variable `LEPTON_INHIBIT_RC_FILES`
is set, which is true during the tests.

Closes #1084

### DIFF
--- a/tools/script.in
+++ b/tools/script.in
@@ -93,8 +93,13 @@ is not supported for your OS, sorry.\n")
             ;; installed.
             (set! %load-compiled-path (cons %lepton-ccache-dir
                                             %load-compiled-path)))
-          (let ((%lepton-scheme-dir
+          (let ((%lepton-scheme-srcdir
                  (string-join (list %m4-abs-top-srcdir
+                                    %m4-package-name
+                                    "scheme")
+                              file-name-separator-string))
+                (%lepton-scheme-builddir
+                 (string-join (list %m4-abs-top-builddir
                                     %m4-package-name
                                     "scheme")
                               file-name-separator-string))
@@ -116,8 +121,12 @@ is not supported for your OS, sorry.\n")
                                     "src"
                                     "libleptonattrib")
                               file-name-separator-string)))
-            ;; Add directory with Scheme modules to %load-path.
-            (add-to-load-path %lepton-scheme-dir)
+            ;; Add directories with Scheme modules to %load-path.
+            ;; Load all modules but (lepton m4) from their source
+            ;; directory.
+            (add-to-load-path %lepton-scheme-srcdir)
+            ;; Load an only built module (lepton m4).
+            (add-to-load-path %lepton-scheme-builddir)
             ;; Set paths to the libs in the build directory.
             (putenv "LIBLEPTON")
             (setenv "LIBLEPTON" %liblepton)


vzh:
@felix-salfelder, you're right :-) It's my oversight, I didn't tested the case of building in a separate directory always starting the build from the source directory.  I've updated the branch, please check if it works for you.

felix-salfelder:
On Sun, Oct 05, 2025 at 07:46:27AM -0700, Vladimir Zhbanov wrote:
> [..] I've updated the branch, please check if it works for you.

I quickly checked e19c16f8d477bc566. Now

    build$ tools/schematic/lepton-schematic

seems to be working fine. Thanks!

